Issue #41, opened 2018-08-07
Separate backend for tidy prediction


alexpghayes:
I've been reworking the `augment()` methods and it's rapidly becoming clear that dealing with idiosyncratic `predict()` methods is going to slow down progress immensely.

In the end `broom` and `parsnip` are both going to want to wrap a bajillion predict methods, and we should report predictions in the same way for consistencies sake. I think we should move this functionality to a separate package. Potentially we could use the [`prediction`](https://github.com/leeper/prediction) package by Thomas Leeper, but we should decide on the behavior we want first.

If we define a new generic / series of generics, we can then test these behaviors in `modeltests` and allow other modelling package developers to guarantee that their `predict()` methods are sane and consistent.

**What I want from a predict method**:

- Returns a tidy tibble
- Never drops missing data (i.e. matches the behavior of `predict.lm(..., na.action = na.pass)`)
- Consistent naming of fitted values
- Uncertainty in predictions

I want all of these to be guaranteed, and for methods that cannot meet these guarantees, I want an informative error rather than a partially correct output.

topepo:
> Returns a tidy tibble

That would be the most tidy thing to do but my first thought is that it should return a vector unless the prediction contains multiple columns (e.g. class probs, multivariate _Y_). Let me ponder this a bit.

> Never drops missing data

I agree on this. It is something that can happen after prediction if people really don't want those data. 

As a corollary, we should always return the same length/nrow as the input even when the prediction produces multiple columns. That may not be automatically 100% tidy, but I think that it is good practice. 

> Consistent naming of fitted values

👍

We should think about the names though. Let's say that we choose `pred` as the name (just for argument). For numeric outputs, what about quantile regression, risk/probability estimates, and others that have a different context than a simple numeric prediction? Do we have different names and document them or should we keep a single label?


> Uncertainty in predictions

Yes, but not be default (if that's what you mean) since 1) most models don't implement them and 2) they might be computationally taxing and, if you don't want them, why would you take the time to generate them. We should have a formal syntax for getting them though; I just disagree with the defaultedness (sp?)

topepo:
Yes, absolutely.

alexpghayes:
Other thoughts:

- Arguments should always default to `NULL` and missing arguments should error
- The default behavior should be to produce a prediction of the same type as the original response data

This should also all end up in `principles` eventually.

alexpghayes:
[...]

PR #62, opened 2018-09-26
Refactor prediction

alexpghayes:
This is an aggressive refactor to move all prediction code to `safepredict`. `safepredict` doesn't have methods for several of the existing models, so definitely don't merge until I've got that done. Plan is to push commits to this PR as an open project.


alexpghayes:
I'm currently pulling most of the prediction tests into `safepredict`. That's where correctness checks should happen.

Once tests are passing, I'm planning to go back through and add simple tests not for correctness, but just that `parsnip` successfully passed the prediction off to `safepredict` and got a reasonable looking result in return. These sanity check tests still need to written, and probably should live in `modeltests`. I'm imagining something along the lines of

```r
model <- fit(model_constructor(...), ...)
output <- predict(model, ...)
modeltests::check_predict_output(model, data, predict_options)
```

(also, this PR should be linked with #41)

alexpghayes:
Current status: close to an MVP.

What I've done so far:

- Stripped out all the prediction code from `parsnip`
- Stripped out all the prediction tests from `parsnip`
- Moved it to `safepredict`

There's a fair amount of translation that needs to happen that is rather time consuming on the `safepredict` side. What I've got right now is minimal functionality -- most of the `safepredict` methods are thiiiiiin wrappers around `predict()`. They most definitely need to get fleshed out and tested on their own. Some (`glm`, `glmnet`, `ranger`) are much more fully featured, but still nothing is well tested.

Things I couldn't implement `safe_predict` methods for because of installation hell with `sparklyr` and `keras`:

- spark boosted trees
- spark linear model
- spark logistic reg
- spark random forest (class and reg)
- keras mlp regression
- keras mlp classification

These models will need `safe_predict` methods (they should have skeletons at the moment) and a sanity check that `parsnip::predict()` is getting something back.

I added some *tiny* tests that should check that at very least `safe_predict` is returning something with the right number of rows. So I *think* this PR won't cause any failures, but I could wrong. There may be some temporary feature loss as certain `type` arguments available in `parsnip` take a moment to become available in `safepredict`.

This PR has taught me an important lesson: namely, that it is sometimes better to make small incremental changes than to burn everything down and start from scratch.

alexpghayes:
Oh, and I also need to add a `kknn` method to `safe_predict()` since Davis added `kknn` models.

alexpghayes:
Ok this is so out of date I'm going to close and once `safe_predict()` is ready will re-open

PR #93, opened 2018-10-18
Quosured argument changes

topepo:
* Tests are updated for the new approach of capturing arguments as quosures as well as the removal of `others()`. 

* Instead of creating a new, clean environment, we go back to creating a child of the current environment. The clean environment had issues with some packages not importing (or calling by namespace) functions from `stats` such as `model.frame`. 

* Fixed `glmnet` prediction issues. The predict method for `model_fit` uses the model specification to get the `lambda` to predictive with. In this branch, all arguments are quosures so this fails. The `glmnet`-specific predict method evaluate the specification arguments before predicting. 

* `varying_args()` now works for complex arguments such as `sampsize = c(12, varying())`. 

* Model wrappers for C5.0, keras, and xgboost are documented.


DavisVaughan:
Should this be merging into the quosure-passthrough branch?

Issue #106, opened 2018-11-18
`parsnip:::multi_predict._multnet()` uses `type=prob` not `type=probs`


DavisVaughan:
So does `parsnip:::multi_predict.lognet()`. Maybe others too? Not sure. Should use `probs`

DavisVaughan:
this might be backwards, and the documentation of `predict.model_fit` argument `type` needs to use `"prob"` not `"probs"`!

From the docs:

```
A single character value or NULL. Possible values are "numeric", "class", "probs", "conf_int", "pred_int", "quantile", or "raw". When NULL, predict will choose an appropriate value based on the model's mode.
```

PR #111, opened 2018-12-05
Update predict.R

grayskripko:
Based on an error message
```{r}
Error: 'type' should be one of: raw, numeric, class, link, prob, conf_int, pred_int and quantile
```


DavisVaughan:
thanks! this will close #106

PR #116, opened 2018-12-14
update vignettes to reflect changes in argument names

EmilHvitfeldt:
Noticed that the bake calls haven't been updates + `roc_auc` and friends have changed argument handling since last time.


DavisVaughan:
Thanks!

PR #121, opened 2019-01-19
add tidy() method for model_fit objects

### LINKED ISSUE #112: add tidy methods for model_fit objects
That just try to run `tidy` on the `fit` element.

malcolmbarrett:
Closes #112

This PR adds a tidy() method for model_fit objects. Right now, it uses a tryCatch to grab the generics message (which seems about the right message to me). I could always remove that and let the generic handle it, since it's a pretty readable message on its own.  Currently, the `tidy.model_fit` object is exported, which is a bit out of norm, so I can change that if needed. 

``` r
library(tidymodels)
library(parsnip)

logistic_reg() %>%
  set_engine("glm") %>%
  fit(Class ~ funded_amnt + int_rate, data = lending_club) %>%
  tidy(conf.int = TRUE, exponentiate = TRUE)
#> # A tibble: 3 x 7
#>   term        estimate  std.error statistic   p.value conf.low conf.high
#>   <chr>          <dbl>      <dbl>     <dbl>     <dbl>    <dbl>     <dbl>
#> 1 (Intercept)  169.    0.157         32.7   3.54e-235  125.      231.   
#> 2 funded_amnt    1.00  0.00000515     0.537 5.91e-  1    1.000     1.00 
#> 3 int_rate       0.853 0.00846      -18.7   1.99e- 78    0.839     0.868

rand_forest(mode = "classification") %>%
  set_engine("randomForest") %>%
  fit(Class ~ funded_amnt + int_rate, data = lending_club) %>%
  tidy()
#> Error: No tidy method for objects of class randomForest
```


R/tidy.R
```diff
@@ -0,0 +1,24 @@
...
+#' @examples
+#' library(broom)
+#'
+#' logistic_reg() %>%
+#'   set_engine("glm") %>%
+#'   fit(Class ~ funded_amnt + int_rate, data = lending_club) %>%
+#'   # tidying model object and passing arguments to broom:::tidy.glm
+#'   tidy(conf.int = TRUE, exponentiate = TRUE)
+tidy.model_fit <- function(x, ...) {
+  tryCatch(generics::tidy(x$fit, ...),
```
alexpghayes: Can I ask about the tryCatch here? This turns an error into a printed message. Is this on purpose? Seems like this could result in confusion when using `map` to `tidy` a list of parsnip models.

My personal preference would be to use 

```
tidy.model_fit <- function(x, ...) generics::tidy(x$fit, ...)
```

to preserve the error.

malcolmbarrett: My original mental model included a tryCatch, but the generic did a pretty good job. Having let it set a bit in my mind, I think everything about the way generics::tidy handles it is clear enough and that it should be as you suggest

topepo:
The example fails with:

```
> logistic_reg() %>%
+   set_engine("glm") %>%
+   fit(Class ~ funded_amnt + int_rate, data = lending_club) %>%
+   # tidying model object and passing arguments to broom:::tidy.glm
+   tidy(conf.int = TRUE, exponentiate = TRUE)
Error in confint.glm(x, level = conf.level) : 
  package 'MASS' must be installed
Calls: %>% ... suppressMessages -> withCallingHandlers -> <Anonymous> -> confint.glm
```
Is there something else that you could demo with so that we wouldn't have to add `MASS` as a dependency?

malcolmbarrett:
Wow, this is a buried one. `stats:::confint.glm` depends on `MASS`, so let me try removing the `conf.int` argument here

malcolmbarrett:
Oh, a follow-up issue: `tidy()`is in `generics`, but `tidy.glm()` is in `broom`. Putting it in suggests will still fail with a documentation example, right? I could add to depends, but that's a big one, too. Any way to have an example and avoid this?

topepo:
Oy vey. Can we do without an example? I'm trying to keep the dependencies low and `broom` would blow that up (even in suggests; `travis` times out).

malcolmbarrett:
Yeah, it's a big one, so I get that! I also feel like the example is nice but not essential given the context, so I removed it

topepo:
Thanks!

Issue #120, opened 2019-01-16
tibble name repair with logistic regression


bcjaeger:
cc @DavisVaughan 

``` r
library(tidymodels)
#> -- Attaching packages ------------------------------------------------------------------------- tidymodels 0.0.2 --
#> v broom     0.5.1     v purrr     0.2.5
#> v dials     0.0.2     v recipes   0.1.4
#> v dplyr     0.7.8     v rsample   0.0.4
#> v ggplot2   3.1.0     v tibble    2.0.1
#> v infer     0.4.0     v yardstick 0.0.2
#> v parsnip   0.0.1
#> -- Conflicts ---------------------------------------------------------------------------- tidymodels_conflicts() --
#> x purrr::discard() masks scales::discard()
#> x dplyr::filter()  masks stats::filter()
#> x dplyr::lag()     masks stats::lag()
#> x recipes::step()  masks stats::step()

analysis=data.frame(target=sample(c(0,1),100,replace=T),
                    x1=rnorm(100),
                    x2=rnorm(100))

head(analysis)
#>   target         x1         x2
#> 1      1 -0.5611995  1.4630014
#> 2      1  1.7830997 -1.6015657
#> 3      1  1.2500967 -1.5411327
#> 4      1 -0.4197018 -0.4608742
#> 5      0 -0.4842685  0.1226822
#> 6      1  0.7493446 -1.2447351

dsplit <- initial_split(analysis, strata = 'target')

trn=training(dsplit)
tst=testing(dsplit)

lrm_fit <- logistic_reg() %>% 
  set_engine('glm') %>%
  fit(target~x1+x2,data=trn) 

lrm_fit %>%
  predict(tst, type='prob')
#> Error: Columns 1, 2 must be named.
#> Use .name_repair to specify repair.
```

<sup>Created on 2019-01-16 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

topepo:
I think that the issue here is the nature of the outcome data. We want to ~coerce~ encourage people to use better data structures (factors here instead of numeric 0/1). 

There is an issue open that will check for factor outcome data in classification models for `parsnip`.

bcjaeger:
[...]

PR #124, opened 2019-01-23
get predicted probabilities from logistic regression

louisahsmith:
Fixes error that occurred when trying to get predicted probabilities from logistic regression when Y was not a factor:

> Error: Columns 1, 2 must be named.
> Use .name_repair to specify repair.

Noted in https://github.com/tidymodels/parsnip/issues/120

When this is the case I replaced the names with "0" and "1" (so the column names of the tibble that results from predict(type = "prob") are .pred_0 and .pred_1) but you may want other names instead -- these seemed to make sense for my use.

### DIFF
--- a/R/logistic_reg_data.R
+++ b/R/logistic_reg_data.R
@@ -50,7 +50,9 @@ logistic_reg_glm_data <-
       pre = NULL,
       post = function(x, object) {
         x <- tibble(v1 = 1 - x, v2 = x)
-        colnames(x) <- object$lvl
+        if (!is.null(object$lvl))
+          colnames(x) <- object$lvl
+        else (colnames(x) <- c("0", "1"))
         x
       },
       func = c(fun = "predict"),


topepo:
> trying to get predicted probabilities from logistic regression when Y was not a factor

I do _not_ want to promote/enable the use of numeric data to represent qualitative data. This is the opposite of what the package is trying to do: having uniform standards for modeling. Sorry that you got to this before I had a chance to make a comment to that effect. 

Having `nnet` as an engine is a good idea and we should probably default `entropy = TRUE`.

topepo:
I'd be happy for a PR as long as it is it is specific to factors as the data type. 

> logistic regression can only be used to model a 0,1 variable

That's technically true of the theoretical model but not true for the common implementations. [Differentiating the computational code from the user-facing code](https://tidymodels.github.io/model-implementation-principles/function-interfaces.html) is our best tool for improving model-related packages. 

We are trying really hard to get people to represent data in the best way. Factors are better because:

* They have levels so that you don't have to manually add in text so that your results have context. You don't care about `Pr[y = 1]`, you care about `Pr[Y = "house sold"]` or whatever the classes represent.
* They have a memory of the levels
* They have a class different then numeric so that S3 classes and other chunks of code can differentiate between the context (i.e. regression vs classification).

Issue #122, opened 2019-01-20
Adding zero rule models for baseline purposes


EmilHvitfeldt:
Would be a small model that would implement the simplest rule(s) possible for the given task.

classification: Always pick the majority class.
regression: Always pick a central tendency (mean/median).
And so on.

alexpghayes:
I am a huge fan of this

topepo:
`caret` has this as a "null model". It is easy to implement.

PR #126, opened 2019-01-23
Adding null model

EmilHvitfeldt:
in reference to #122.

Still WIP. Most of the skeleton is there but the `predict()` outputs length of fitting data not `new_data`.


topepo:
Please use `type = "numeric"` for regression and `null_model` for the user-facing function. 

@DavisVaughan will get a laugh out of this but please lint the R code. Opening brackets should be on the same line as the function definition/`if` etc.

```r
null_model.default <- function(x = NULL, y, ...) {

}
```

Can you also enable multivariate numeric regression?

EmilHvitfeldt:
> Can you also enable multivariate numeric regression?

Consider it done!

``` r
library(parsnip)

iris_basic <- null_model(mode = "regression") %>% 
  set_engine("parsnip") %>%
  fit(cbind(Sepal.Width) ~ ., data = iris)
  
predict_numeric(iris_basic, iris[1:5,])
#> [1] 3.057333 3.057333 3.057333 3.057333 3.057333

iris_basic <- null_model(mode = "regression") %>% 
  set_engine("parsnip") %>%
  fit(cbind(Sepal.Width, Petal.Width) ~ ., data = iris)
  
predict_numeric(iris_basic, iris[1:5,])
#>   Sepal.Width Petal.Width
#> 1    3.057333    1.199333
#> 2    3.057333    1.199333
#> 3    3.057333    1.199333
#> 4    3.057333    1.199333
#> 5    3.057333    1.199333
```

<sup>Created on 2019-01-25 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

While we are at it, should we also add `quantile` and `classprob`? It doesn't make sense to add `conf_int` and `pred_int`.

topepo:
`type = "classprob"` would be helpful. I don't think that we have quantiles implemented anywhere so I'd hold off on that.

topepo:
Thanks!

Issue #110, opened 2018-11-30
`varying_args.step` should include a unique ID column


DavisVaughan:
Eventually, when we are filling in `varying()` parameters we will need to know exactly which step they belong to. The current amount of information is likely not good enough because there is no unique identifier for a recipe step (i.e. you wouldn't be able to tell apart two calls to step_pca with `num_comp` varying in both steps). The `tibble` returned by `varying_args.step` should include a unique id column created from:

``` r
library(recipes)
ex <- recipe(Species ~ ., data = iris) %>% step_log(Species)
ex$steps[[1]]$id
#> [1] "log_YAIYj"
```

PR #137, opened 2019-01-25
varying_args() improvements:

DavisVaughan:
- Update documentation for varying_args()
- For recipes, using the unique step id as the id column
- For model_specs, using the top level class as the id column
- Better return value for recipes with no steps (tibble with correct column names)


DavisVaughan:
Fixes #110

Issue #139, opened 2019-01-29
informative error for using predict on non-fitted model


EmilHvitfeldt:
I ran absentmindedly into this situation where I try to predict a non-fitted model.

``` r
library(parsnip)

model <- linear_reg() %>%
  set_engine("lm")

model %>%
  fit(mpg ~ ., data = mtcars)
#> parsnip model object
#> 
#> 
#> Call:
#> stats::lm(formula = formula, data = data)
#> 
#> Coefficients:
#> (Intercept)          cyl         disp           hp         drat  
#>    12.30337     -0.11144      0.01334     -0.02148      0.78711  
#>          wt         qsec           vs           am         gear  
#>    -3.71530      0.82104      0.31776      2.52023      0.65541  
#>        carb  
#>    -0.19942

predict(model, new_data = mtcars)
#> Error in UseMethod("predict"): no applicable method for 'predict' applied to an object of class "c('linear_reg', 'model_spec')"
```

<sup>Created on 2019-01-28 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

I think a simple `predict.model_spec` with a kind error-message might solve the problem. If it sounds like a good idea I can send a PR.

topepo:
That would be great. Thanks

PR #143, opened 2019-02-02
PR regarding #139

EmilHvitfeldt:
Referencing #139.

This PR add predict method for model_spec to remind users that they have to fit their model before they can use predict.


R/predict.R
```diff
@@ -224,4 +224,7 @@ multi_predict.default <- function(object, ...)
   stop ("No `multi_predict` method exists for objects with classes ",
         paste0("'", class(), "'", collapse = ", "), call. = FALSE)
 
-
+#' @export
+predict.model_spec <- function(object, ...) {
+  stop("You must use `fit` on your model specification before you can use `predict`.", call. = FALSE)
```
DavisVaughan: Can you please change this to `fit()` and `predict()`?

EmilHvitfeldt: Fixed it. however when looking the rest of the code this is not being done.

https://github.com/tidymodels/parsnip/blob/b84fa64ea6f85bddd382dd34b7d34ec6ed9d035d/R/fit.R#L97

https://github.com/tidymodels/parsnip/blob/b84fa64ea6f85bddd382dd34b7d34ec6ed9d035d/R/fit.R#L100

topepo: We haven't caught up to @DavisVaughan's strict standards yet 😃

PR #164, opened 2019-03-29
Addition of liquidSVM engine for svm_rbf and FNN engine for nearest_neighbor

stevenpawley:
Hello,

Thanks for all of your work with parsnip - I think the combination of recipes + parsnip + rsample + dials is fantastic.

As part of working with some large datasets recently, I could only use svm_rbf and nearest_neighbor methods if I considered alternative (and faster) engines, using the liquidSVM and the Fast Nearest Neighbor (FNN) packages respectively.

Would you consider a pull request, with these packages added as engines?


topepo:
We can keep it in `parsnip` but can you email the owners of `FNN` and see if they can add `fnn_train` and `fnn_test` so that we can lose the hard dependency? We'd like to move the wrappers out when possible.

R/svm_rbf.R
```diff
@@ -140,31 +140,46 @@ update.svm_rbf <-
...
       arg_vals$kpar <- kpar
     }
-    
+
+  }
+
+  if (x$engine == "liquidSVM") {
+    # convert parameter arguments
+    if (any(arg_names == "rbf_sigma")) {
+      arg_vals$gamma <- arg_vals$rbf_sigma
```
topepo: Am I right in thinking that `gamma` is usually the same as `1/sigma`?

topepo:
(sorry, I didn't mean to close that)

topepo:
Can you add unit tests and update the NEWS file?

R/svm_rbf.R
stevenpawley: Yes from I what have read, gamma is the same as 1/sigma in other SVM implementations. So, I should use 1/sigma as part of converting the parameter arguments. I also realized that the arg_names passed to the translate function are C and sigma. I'll modify the pull request.

stevenpawley: Sorry, and write unit tests and update the NEWS file

topepo:
Unit tests and we're good!

topepo:
I went to test this out a bit and it looks like [`liquidSVM`](https://cran.r-project.org/web/packages/liquidSVM/index.html) has been orphaned. Do you know anything about the circumstances of that?

stevenpawley:
No, I didn't realize that the package had been orphaned. I'll see if I can get in touch with the author's and will report back. Relating to the use of additional engines, one possible difficulty with the parsnip package is that although a user can add their own model specification+engine, adding an additional engine to an existing model specification appears to require changes to the translate function in the package? Perhaps this part of the package could be further modularized so that the translation of the engine's parameters can be defined outside of the parsnip package by a user-specified function?

topepo:
[liquidSVM](https://cran.r-project.org/web/packages/liquidSVM/index.html) looks to be back on the menu. I'll take another look at this in a few days; we've changed how we register models. 

I'll translate this work to the new system and kick it around.

stevenpawley:
Great, I was in touch with the authors and were unaware that the package had been removed from CRAN. Also if you don't have time, I don't mind making some amendments - I've used the new way of registering models as wanted to try Cubist on a problem.

UnclAlDeveloper:
Has / will this been added to the master repository?  Using show_model_info for svm_rbf only shows kernlab.  It would certainly be useful to have faster SVM and KNN for large datasets.

stevenpawley:
> Has / will this been added to the master repository? Using show_model_info for svm_rbf only shows kernlab. It would certainly be useful to have faster SVM and KNN for large datasets.

Some issues were that liquidSVM went off CRAN for a time, and the structure for adding new models was refactored. I have a liquidSVM engine using the newer approach to adding new engines, I'll submit a new pull request in the next day or so. Another issue that wasn't discussed is that you can't save liquidSVM models using the typical approaches like `saveRDS` so that might need some discussion as to whether another function is needed to save a `model_spec` or a `workflow` containing a liquidSVM model, or should the user just save the liquid part of the model with the `liquidSVM::write.liquid` function.

topepo:
I went to do modifications to the PR a while back and something must have happened to the original source ("merge 14 commits into `tidymodels:master` from `unknown repository`")

stevenpawley:
Sorry, I reforked parsnip at one point. I have a reworked liquidSVM engine that I can resubmit as a pull request. I also have a specification for using `neuralnet` as an engine for mlp. Are you interested in `neuralnet` as a separate pull request?

topepo:
I'd like to try to include these but in a way that does not add dependencies (if possible). Basically, the unit tests would be run locally (for now). 

For `fnn` we can get rid of the wrapper function since `set_ft()` depends on the model mode (so you can specify different functions for classification and regression). 

If you add another PR, I can review and make changes.

stevenpawley:
Thanks for the input, I can look at trying to make those changes. One question before proceeding further - how do you feel about including liquidSVM in `parsnip` based on the issue of saving/loading models? liquidSVM models cannot be saved like regular R objects; the fitted model has to be saved using `write.liquidSVM`. I'm sure that a lot of users won't realize this at first and will run some models in parsnip, `saveRDS` the results and then not be able to load them again.

topepo:
> liquidSVM models cannot be saved like regular R objects; the fitted model has to be saved using `write.liquidSVM`.

We can write notes in the engine details section and provide a simple code example.

topepo:
I'll close this since we added the SVM model. Go ahead and create a new PR if you still want to the KNN engine.

Issue #167, opened 2019-04-04
restructure  model information objects


topepo:
The objects in the [`data` files](https://tidymodels.github.io/parsnip/articles/articles/Scratch.html) could have a better format. Two examples:

 * The `fit$func` slot should be able to have different functions for different model modes. For example, if a package has two separate functions for regression and classification, we end up needing a wrapper to include it. 

 * The parameter key tibble should include a reference to where the recommended function for initializing the object should reside. For example, for a random forest model,  the parameter could be initialized by the corresponding `dials` object. 

Paging @DavisVaughan, @alexpghayes, and @stevenpawley for notes on where the pain points are and recommendations for better organization.

topepo:
For the parameter key, I'd suggest having a list column that has a function reference. For example, 

```r
# linear_reg_arg_key
  tibble::tribble(
   ~parsnip, ~lm,  ~glmnet,              ~spark, ~stan,  ~keras,                             ~func,
  "penalty",  NA, "lambda",         "reg_param",    NA, "decay", c(pkg = "dials", fun = "penalty"),
  "mixture",  NA,  "alpha", "elastic_net_param",    NA,      NA, c(pkg = "dials", fun = "mixture")
  )
```

topepo:
Previous thoughts from @DavisVaughan:

Extensibility:

* Think about how someone could extend parsnip to work with their model without having to make a PR to get that model integrated into the core of parsnip.
* Think about maintaining an extensible but internal list of types of objects that are needed to make parsnip work:
  - **parsnip_modes** - A list of data frames, 1 for each model, with the allowed modes of that model. Each data frame has a single column per engine and the rows are mappings to whether or not the engine allows class/reg/unknown. Accessible for external packages through `register_model_mode()` for new models, and `register_engine_mode()` for new engines. (this could replace both the modes vector and the engines vector since that info is redundant)
  - **parsnip_arg_key** - A list of data frames, one for each model. Each data frame has a single column per engine and the rows are the mapping to the core arguments. Accessible by `register_model_args()` for a new model (new data frame) and `register_engine_args()` for a new column of a data frame.
[...]

PR #178, opened 2019-06-01
Internal refactoring of model information

topepo:
For issue #167 

A set of model registration functions are added so that users can access models via `parsnip`. It also changes how the data about models is represented. 

This needs some review from @DavisVaughan and anyone else who wants to give feedback


DavisVaughan:
i've set aside some time tomorrow morning to review this. could you please give me a few bullet points of the high level changes / ideas? Should I look straight to the reworked vignette?

topepo:
* The model information are now stored in an environment (obtained via `get_mod_env()`)

* `fit()`, `translate()`, and `predict()` internals now use the new data structures (not a huge change)

* A bunch of `set_*()` and `check_*()` functions are used to get models into the environment.

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+#' @examples
+#' # Show the information about a model:
+#' show_model_info("rand_forest")
+#'
+#' # Access the model data:
+#' current_code <- get_model_env()
+#' ls(envir = current_code)
+#'
+#' @export
+check_mod_val <- function(model, new = FALSE, existence = FALSE) {
```
DavisVaughan: Why are both `new` and `existence` needed? It was surprising to me to see both of them

DavisVaughan: Why would `new = FALSE` ever be used? Otherwise it looks like you could have two entries for the same model in `set_new_model()`

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+#' @rdname check_mod_val
+#' @keywords internal
+#' @export
+set_new_model <- function(model) {
+  check_mod_val(model, new = TRUE)
+
+  current <- get_model_env()
+
+  current$models <- c(current$models, model)
+  current[[model]] <- dplyr::tibble(engine = character(0), mode = character(0))
```
DavisVaughan: It might be cleaner to:

1) Extract any time you create empty tibbles into helpers. So `dplyr::tibble(engine = character(0), mode = character(0))` would become `new_model_tbl()`. You'd also have `new_model_pkgs_tbl()` and so on.

2) Add them all to the environment at once using `rlang::env_bind()`. I would probably make a named list that looks like `model_list = list(model = new_empty_model(), model_pkgs = new_empty_model_pkgs(), etc)` and then splice it into `env_bind(current, !!! model_list)`

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+}
+
+# ------------------------------------------------------------------------------
+
+#' @rdname check_mod_val
+#' @keywords internal
+#' @export
+set_model_mode <- function(model, mode) {
+  check_mod_val(model, existence = TRUE)
+  check_mode_val(mode)
```
DavisVaughan: I guess it is up to the developer to spell `"classification"` correctly?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+
+  current <- get_model_env()
+
+  if (!any(current$modes == mode)) {
```
DavisVaughan: Do you really need a global list of all used modes? Is `current$modes` used anywhere else?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+
+# ------------------------------------------------------------------------------
+
+#' @rdname check_mod_val
+#' @keywords internal
+#' @export
+set_model_engine <- function(model, mode, eng) {
+  check_mod_val(model, existence = TRUE)
+  check_mode_val(mode)
+  check_mode_val(eng)
```
DavisVaughan: `check_eng_val`?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+  combined <-
+    dplyr::inner_join(new_arg %>% dplyr::select(engine, parsnip, original),
+                      old_args %>% dplyr::select(engine, parsnip, original),
+                      by = c("engine", "parsnip", "original"))
+  if (nrow(combined) != 0) {
+    stop("A model argument already exists for ", model, " using the ",
+         eng, " engine. You cannot overwrite arguments.", call. = FALSE)
+  }
+
+  # TODO cant currently use `distinct()` on a list column.
```
DavisVaughan: Do you need to do `distinct()` if you already checked that you aren't modifying existing elements above?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+  # Use `vctrs::vctrs_duplicated()` instead
+  updated <- try(dplyr::bind_rows(old_args, new_arg), silent = TRUE)
+  if (inherits(updated, "try-error")) {
+    stop("An error occured when adding the new argument.", call. = FALSE)
+  }
+
+  updated <- dplyr::distinct(updated, engine, parsnip, original, has_submodel)
```
DavisVaughan: So wait, this loses `func`, so it isn't working right now, right?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+  check_mod_val(model, new = TRUE)
+
+  current <- get_model_env()
+
+  current$models <- c(current$models, model)
+  current[[model]] <- dplyr::tibble(engine = character(0), mode = character(0))
+  current[[paste0(model, "_pkgs")]] <- dplyr::tibble(engine = character(0), pkg = list())
+  current[[paste0(model, "_modes")]] <- "unknown"
+  current[[paste0(model, "_args")]] <-
+    dplyr::tibble(
```
DavisVaughan: does this need `has_submodel`?

vignettes/articles/Scratch.Rmd
```diff
@@ -40,70 +40,86 @@ str(mda::mda)
...
-    if (!(mode %in% mixture_da_modes))
-      stop("`mode` should be one of: ",
-           paste0("'", mixture_da_modes, "'", collapse = ", "),
-           call. = FALSE)
+    if (mode  != "classification") {
+      stop("`mode` should be 'classification'", call. = FALSE)
+    }
     
     # Capture the arguments in quosures
     args <- list(sub_classes = rlang::enquo(sub_classes))
```
DavisVaughan: should we expose `new_model_spec` and use it here?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+  }
+  exp_nms <- c("defaults", "func", "interface", "protect")
+  if (!isTRUE(all.equal(sort(names(fit_obj)), exp_nms))) {
+    stop("The `fit` module should have elements: ",
+         paste0("`", exp_nms, "`", collapse = ", "),
+         call. = FALSE)
+  }
+
+  exp_interf <- c("data.frame", "formula", "matrix")
+  if (length(fit_obj$interface) > 1) {
```
DavisVaughan: this should probably be `!= 1`. you check that `fit_obj$interface` exists but someone could do `list(interface = character())` (the next line would catch it, but still)

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+  current <- get_model_env()
+  model_info <- current[[paste0(model)]]
+  old_fits <- current[[paste0(model, "_fit")]]
+
+  has_engine <-
+    model_info %>%
+    dplyr::filter(engine == eng & mode == !!mode) %>%
+    nrow()
+  if (has_engine != 1) {
+    stop("set_fit The combination of engine '", eng, "' and mode '",
```
DavisVaughan: did you mean to say `set_fit` here?

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+  }
+  rlang::env_get(get_model_env(), fit_name)
+}
+
+# ------------------------------------------------------------------------------
+
+#' @rdname check_mod_val
+#' @keywords internal
+#' @export
+set_pred <- function(model, mode, eng, type, value) {
```
DavisVaughan: I wonder if we should have helpers for `set_pred_value()` so people don't have to construct the list by hand. It would autocomplete the names of the `value` list for them, and would set `pre` and `post` to `NULL` by default. And we could do some validation there

vignettes/articles/Scratch.Rmd
```diff
@@ -228,98 +251,82 @@ As a developer, one thing that may come in handy is the `translate` function. Th
...
-```{r glm-alt-show}
+```{r glm-alt-show, eval = FALSE}
 logistic_reg() %>% translate(engine = "glm")
 
 # but you can change it:
 
-logistic_reg() %>% 
-  set_engine("glm", family = stats::binomial(link = "probit")) %>%
-  translate()
+logistic_reg(others = list(family = expr(binomial(link = "probit")))) %>%
```
DavisVaughan: I don't think this is right. `others`?

R/aaa_models.R
topepo: I removed it (so now `value` is saved)

R/aaa_models.R
topepo: Correct. I figured that there was a `vctrs` approach to take here but I removed the `distinct()` about so the `values` are saved now.

R/aaa_models.R
topepo: yep

R/aaa_models.R
topepo: yep

R/aaa_models.R
```diff
@@ -0,0 +1,722 @@
...
+         call. = FALSE)
+  }
+
+  exp_interf <- c("data.frame", "formula", "matrix")
+  if (length(fit_obj$interface) > 1) {
+    stop("The `interface` element should have a single value of : ",
+         paste0("`", exp_interf, "`", collapse = ", "),
+         call. = FALSE)
+  }
+  if (!any(fit_obj$interface == exp_interf)) {
```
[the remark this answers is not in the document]
topepo: sure

R/aaa_models.R
topepo: nope, unintended

R/aaa_models.R
topepo: Would that include the meta-info (e.g. `model`, `eng`, etc) or just something like this:

```r
pred_value_template <-  function(pre = NULL, post = NULL, func, args = list()) {
  if (rlang::is_missing()) {
    stop("Please supply a value to `func`. See `?set_pred`.", call. = FALSE)
  }
  list(pre = pre, post = post, func = func, args = args)
}
```

vignettes/articles/Scratch.Rmd
topepo: That's a vestigial piece of code from the first prototype version. I'll fix that.

R/aaa_models.R
topepo: It is

R/aaa_models.R
topepo: To check them, we need a list of all modes and, if they are scattered across objects, we'd have to spend the time finding and parsing them.

vignettes/articles/Scratch.Rmd
topepo: No, the spec object (in `method`) isn't initialized until `translate()` is used. If we did it here, it would just be over-written.

R/aaa_models.R
topepo: That's more elegant but more likely to break.

PR #192, opened 2019-07-15
added a function to get multi_predict arg names

topepo:


R/aaa_multi_predict.R
```diff
@@ -33,3 +33,102 @@ multi_predict.default <- function(object, ...)
...
+#' @rdname has_multi_predict
+has_multi_predict.model_fit <- function(object, ...) {
+  existing_mthds <- utils::methods("multi_predict")
+  tst <- paste0("multi_predict.", class(object))
+  any(tst %in% existing_mthds)
+}
+
+#' @export
+#' @rdname has_multi_predict
+has_multi_predict.workflow <- function(object, ...) {
```
DavisVaughan: Do you think this method should live in workflows? Can you even test this function in parsnip without suggesting workflows? It's always hard to know where to put what.

topepo: maybe `has_multi_predict.workflow()` should but the other bits should probably stay here. The workflow method is more of a place holder for when that package is more mature

> It's always hard to know where to put what.

Yeah. That is becoming increasingly true.

PR #224, opened 2019-10-29
Compute and print time elapsed

### LINKED ISSUE #175: add a elapsed time slot for fitted models
Basically wrap the fitting parts in `system.time()` and save it in the parsnip object.

patr1ckm:
This computes and prints the elapsed time when fitting the model for both `fit_xy` and `fit`.

Fixes #175 .


topepo:
Can you use `system.time()` instead? Also, use the `prettyunits` package to print the total elapsed time.

patr1ckm:
`prettyunits` is great, thanks! Here's how it looks:

Screen Shot 2019-11-01 at 10 45 17 AM

topepo:
Excellent. Thanks

PR #227, opened 2019-10-30
Fix varying_args with functions

### LINKED ISSUE #219: varying_args() fails on any step that contains a function as an argument
Currently the varying_args.recipe() with throw on error if any of the steps in the recipe contain a function as one of the arguments. This is because the find_varying() helper function recursively checks each argument in a step in:

```
for (i in seq_along(x)) {
    varying_elems[i] <- find_varying(x[[i]])
}
```

Howwever, if that argument is a function it will fail because you cannot subset a function. It fails with steps such as:

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'xts':
#>   method     from
#>   as.zoo.xts zoo
#> ── Attaching packages ─────────────────────────────────────────────────────────────── tidymodels 0.0.2 ──
#> ✔ broom     0.5.2       ✔ purrr     0.3.2  
#> ✔ dials     0.0.3       ✔ recipes   0.1.7  
#> ✔ dplyr     0.8.3       ✔ rsample   0.0.5  
#> ✔ ggplot2   3.2.1       ✔ tibble    2.1.3  
#> ✔ infer     0.5.0       ✔ yardstick 0.0.4  
#> ✔ parsnip   0.0.3.1
#> ── Conflicts ────────────────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> ✖ purrr::discard()  masks scales::discard()
#> ✖ dplyr::filter()   masks stats::filter()
#> ✖ dplyr::lag()      masks stats::lag()
#> ✖ ggplot2::margin() masks dials::margin()
#> ✖ dials::offset()   masks stats::offset()
#> ✖ recipes::step()   masks stats::step()

data(iris)
iris["Species_num"] <- as.numeric(iris$Species)

rec <- iris %>%
  recipe(Species ~ .) %>%
  step_num2factor(Species_num)

varying_args(rec)
#> Error in x[[i]]: object of type 'closure' is not subsettable
```

<sup>Created on 2019-10-03 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This is because step_num2factor contains a `transform` argument with a default of `transform = function(x) x`. Settings transform to NULL also throws an error.

I think that an easy fix is using something like the following, in the find_varying() function in varying_args.R:

```
  # STEP 2 - Recursion

  varying_elems <- vector("logical", length = length(x))

  if (!inherits(x, "function")) {
    for (i in seq_along(x)) {
      varying_elems[i] <- find_varying(x[[i]])
    }
  } else{
    varying_elems <- FALSE
  }
```

patr1ckm:
Occasionally some args to parameters of e.g. a step in a recipe are functions:

```
rec <- iris %>%
  recipe(Species ~ .) %>%
  step_num2factor(Species_num)
```

When this occurs, `find_varying` would fail by attempting to subset on a function. Instead, we return `FALSE` for this base case, as `is_varying` has already been checked and would have returned. The end result is that `find_varying` returns `FALSE` for arbitrary functions in arguments but returns `TRUE` only if the function is `varying()`.

Fixes #219 .


topepo:
The new [`tune`](https://github.com/tidymodels/tune) package has better infrastructure for this and `varying()` and `varying_args()` will be deprecated.

Issue #88, opened 2018-10-10
Move objects frequently used in tests to helper-objects.R to reduce redundancy


alexpghayes:
For example, the control objects are created at the beginning of many tests

```
ctrl <- fit_control(verbosity = 1, catch = FALSE)
caught_ctrl <- fit_control(verbosity = 1, catch = TRUE)
quiet_ctrl <- fit_control(verbosity = 0, catch = TRUE)
```

but could be defined just once in a file `tests/testthat/helper-objects.R`. Similarly, much of the data for testing is the same for different models, and it makes sense to create the data-related objects just once:

```r
lending_club <- head(lending_club, 200)
num_pred <- c("funded_amnt", "annual_inc", "num_il_tl")
```

PR #231, opened 2019-11-01
REF test_control to helper-objects

patr1ckm:
This is a quick refactor to move control objects in tests to a single file, `helper-objects.R` as suggested in #88. This was relatively standard across tests, with exceptions for stan which needed less verbosity.

This is the easiest refactor suggested in the issue to be sure, but I'd like to handle the refactor of data references separately in smaller chunks. There's a bit of variability that will take some work to standardize, but I think it could be nice to have.

### DIFF
--- a/tests/testthat/test_linear_reg_keras.R
+++ b/tests/testthat/test_linear_reg_keras.R
@@ -7,6 +7,7 @@ library(tibble)
 
 context("keras linear regression")
 source("helpers.R")
+source("helper-objects.R")
 
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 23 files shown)


tests/testthat/test_linear_reg_keras.R
```diff
@@ -50,6 +50,7 @@ test_that('model fitting', {
       ),
     regexp = NA
   )
+  fit1$elapsed <- fit2$elapsed
```
patr1ckm: Not sure how this slipped out of the previous commits, but it's necessary for the `keras` tests to the comparison of the objects here.

topepo: Did you update from master? I've merged a bunch of PRs in a short amount of time.

patr1ckm: Maybe I did get out of sync slightly! Apologies

topepo:
Thanks

PR #248, opened 2019-12-16
update vignette "Making a parsnip model from scratch"

hfrick:
fixed the step numbers and a few more typos

also noticed that the pkgdown site still shows the typo from PR #233


topepo:
Thanks!

PR #256, opened 2020-01-24
Update Regression.Rmd

blairj09:
The heading Penalized Logistic Regression has been changed to Penalized Linear Regression


topepo:
Thank you!

PR #300, opened 2020-05-05
Addition of liquidSVM engine to svm_rbf

stevenpawley:
Hi Max,

Here is a pull request with the updated liquidSVM engine. I've removed the nearest neighbours `FNN` package for now because AFAICT it requires a wrapper (FNN contains separate functions for regression and classification but they don't follow the formula or x/y interface). The predict method also needs a wrapper and I can't see a way around that.

Regarding liquidSVM, from the vignette, it appears that `sigma` is the equivalent of liquidSVMs `1/gammas^2` which is in the modified `translate.svm_rbf` method.


topepo:
Thanks! I'll merge and make a few edits. We just started using pre-rendered Rmd files for the engine details.

Issue #315, opened 2020-05-20
non-standard data arguments


topepo:
In some cases, a model function that `parsnip` is calling has non-standard argument names for the data-oriented components. 

For example, a function `foo()` that only has the x/y interface might have a signature like `foo(X, Y)` instead of the standard `foo(x, y)`. 

Some real examples: 

 * `sparklyr` models use `x` as the `data` slot (and put it in the first position)
 * `kernlab::ksvm(x, data)` where `x` is a formula. 
 * `MASS::lda(x, grouping)`

juliasilge:
This change is implemented in #316. 🚀

Issue #274, opened 2020-03-27
coercing rpart model fit with parsnip to partykit fails


juliasilge:
A model fit directly using **rpart** can be coerced to **partykit**, but the same model fit by **parsnip** cannot be coerced, even though they have the same class.

``` r
library(parsnip)
library(rpart)
library(partykit)
#> Loading required package: grid
#> Loading required package: libcoin
#> Loading required package: mvtnorm

tree_spec <- 
  decision_tree(
    cost_complexity = 1e-10,
    tree_depth = 4
  ) %>% 
  set_engine("rpart") %>% 
  set_mode("classification")

rpart_fit <- rpart(Kyphosis ~ Age + Number + Start, data = kyphosis)
parsnip_fit <- tree_spec %>%
  fit(Kyphosis ~ Age + Number + Start, data = kyphosis)

class(rpart_fit)
#> [1] "rpart"
class(parsnip_fit$fit)
#> [1] "rpart"


as.party(rpart_fit)
#> 
#> Model formula:
#> Kyphosis ~ Age + Number + Start
#> 
#> Fitted party:
#> [1] root
#> |   [2] Start >= 8.5
#> |   |   [3] Start >= 14.5: absent (n = 29, err = 0.0%)
#> |   |   [4] Start < 14.5
#> |   |   |   [5] Age < 55: absent (n = 12, err = 0.0%)
#> |   |   |   [6] Age >= 55
#> |   |   |   |   [7] Age >= 111: absent (n = 14, err = 14.3%)
#> |   |   |   |   [8] Age < 111: present (n = 7, err = 42.9%)
#> |   [9] Start < 8.5: present (n = 19, err = 42.1%)
#> 
#> Number of inner nodes:    4
#> Number of terminal nodes: 5
as.party(parsnip_fit$fit)
#> Error in eval(predvars, data, env): invalid 'envir' argument of type 'closure'
```

<sup>Created on 2020-03-26 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

This is related to how **partykit** uses the formula of the original object it is coercing. Can we repair the call formula?

juliasilge:
This is now closed in #316.

``` r
library(parsnip)
library(rpart)
library(partykit)
#> Loading required package: grid
#> Loading required package: libcoin
#> Loading required package: mvtnorm

tree_spec <- 
  decision_tree(
    cost_complexity = 1e-10,
    tree_depth = 4
  ) %>% 
[...]

PR #316, opened 2020-05-20
Handling alternative data argument names and fixing calls

topepo:
As mentioned in #315: 

In some cases, a model function that `parsnip` is calling has non-standard argument names for the data-oriented components. 

For example, a function `foo()` that only has the x/y interface might have a signature like `foo(X, Y)` instead of the standard `foo(x, y)`. 

Some real examples: 

 * `sparklyr` models use `x` as the `data` slot (and put it in the first position)
 * `kernlab::ksvm(x, data)` where `x` is a formula. 
 * `MASS::lda(x, grouping)`  

This PR allows for these types of arguments. 

Model definitions have the option to add a `data` element to the definition that maps the argument name to the objects that are created by `fit()` or `fit_xy()`. 

For example, for formula-based `ksvm()` models, the new definition uses 

```r
set_fit(
  model = "svm_rbf",
  eng = "kernlab",
  mode = "regression",
  value = list(
    interface = "formula",
    data = c(formula = "x", data = "data"),    # <---- the new bit here
    protect = c("x", "data"),
    func = c(pkg = "kernlab", fun = "ksvm"),
    defaults = list(kernel = "rbfdot")
  )
)
```

This is optional so most models are unaffected. The model definitions that were changed in the PR: 

* `svm_rbf()` and `svm_poly()` with `kernlab` engines (for the above reason)

* `mars()` with `earth` engines (for other reasons)

**Bonus Feature 1** 🎉 

Previously, when the arguments were translated to a call, the actual formula was not used. It contained a generic symbol called `formula`. In this PR we substitute the user-provided formula to the call.

```r
> mod <-
+   decision_tree() %>%
+   set_engine("rpart", method = "anova") %>%
+   set_mode("regression")
> 
> mod_fit <- fit(mod, mpg ~ ., data = mtcars)
> 
> # previously
> mod_fit$fit$call
rpart::rpart(formula = formula, data = data, method = ~"anova")
> 
> # now
> mod_fit$fit$call
rpart::rpart(formula = mpg ~ ., data = data, method = ~"anova")
```  

This should help with #274. @StephenMilborrow, @apreshill, and others have mentioned this issue of borked calls in the past. This and the extra special bonus feature below should make this more tolerable. 

**Bonus Feature 2** 🎉 🎉 

The above change doesn't really solve #274 since 1) the generic `data` symbol does not link to the user's data object and 2) extra arguments (like `anova` above) are quosures and not the real object. 

This version contains a function called `repair_call()` that fixes both of these issues if the user provides the original data to the function (just to get its name). 

For the example above: 

```r
> mod <-
+   decision_tree() %>%
+   set_engine("rpart", method = "anova") %>%
+   set_mode("regression")
> 
> mod_fit <- fit(mod, mpg ~ ., data = mtcars)
> 
> # old and busted
> mod_fit$fit$call
rpart::rpart(formula = mpg ~ ., data = data, method = ~"anova")
> 
> # new hotness
> repair_call(mod_fit, mtcars)$fit$call
rpart::rpart(formula = mpg ~ ., data = mtcars, method = "anova")
>
> mod2 <- repair_call(mod_fit, mtcars)
> partykit::as.party(mod2$fit)

Model formula:
mpg ~ cyl + disp + hp + drat + wt + qsec + vs + am + gear + carb

Fitted party:
[1] root
|   [2] cyl >= 5
|   |   [3] hp >= 192.5: 13.414 (n = 7, err = 28.8)
|   |   [4] hp < 192.5: 18.264 (n = 14, err = 59.9)
|   [5] cyl < 5: 26.664 (n = 11, err = 203.4)

Number of inner nodes:    2
Number of terminal nodes: 3
```

You might wonder why we need to manually call this extra function. Can't this function be invoked at the end of `fit()` to happen automatically? 

The problem with automating this is that, once `tune` and `workflows` are used, the data that are passed internally to `fit()` might not available to the user (i.e. the might not match the user's data object). The easiest example is when a recipe is used in a workflow; the processed data are not the same as the original data. 

In this case, the user could run the recipe on the data that was used to build the model and then repair the call with that object.


juliasilge:
Aaaaah, this is great.

StephenMilborrow:
Thank you for this.  This will help compatibility with existing tools.  I love the idea of a uniform interface to models (everyone does), but past efforts I've seen to achieve this (e.g. the mlr package) have done so in a way that borked existing tools.  The caret package is a noteable exception.  I tried to summarize existing old-school practices in [Guidelines for S3 Regression Models](http://www.milbo.org/doc/modguide.pdf) .

PR #317, opened 2020-05-22
Fix a couple of documentation typos.

rorynolan:
I looked for other errors of this type and there were none :-)


juliasilge:
Thank you so much! 🙌

PR #319, opened 2020-05-26
Add engine specification field for predictor encodings

### LINKED ISSUE #290: Adding an engine specification field for predictor encodings
As described in [this post](https://rviews.rstudio.com/2017/03/01/the-r-formula-method-the-bad-parts/), there is an extreme amount of heterogeneity in how models are specified, especially as it relates to the R model formula: 

 1. **traditional R formula** is used in 95% of cases where the formula specifies the variable roles and encodes categorical data with dummy variables (e.g. `lm()`).

 1. **slightly-less-traditional R formula** happens with tree-based models (and others) where roles are declared but no dummy variables are created. `ranger()` is the best example. 

 1. **no formulas** when the underlying model function has no formula method and our code uses a formula to make the appropriate x/y data structures (`glmnet()` is the poster child for this)

 1. **non-standard formulas** don't use the standard R infrastructure and usually specify more specific roles for predictors ([GAMS](https://rdrr.io/cran/mgcv/man/gam.models.html) and [multilevel models](https://rdrr.io/cran/lme4/man/lmer.html) are the best examples). The encoding requirement depends on the model type. Such packages mostly to have their own specialized formula parsers.

Since they try to provide a uniform model interface, there are two major problems for the tidymodels packages because of this: 

* A user expects a formula method given to `parsnip` or `workflows` behaves the same way as it would have with the underlying function. (example: tidymodels/tune#151)

* For nonstandard formulas, there should be a way to easily use them in our infrastructure. (see tidymodels/workflows#34)

## Potential solution

[When engine details are specified](https://www.tidymodels.org/learn/develop/models/) we could add a descriptor for each model/engine that tells us how to handle the encoding issue. For example, the `ranger` engine could have a flag that dummy variables should not be created. 

For `parsnip`, this would affect when the user uses `fit(model, formula, data)` and the underlying function uses the x/y interface. Apart from that, it should not impact the package. 

For `workflows`, this would tell the underlying `hardhat` code what to do for what model. We might be able to give `add_formula()` the ability to use nonstandard formulas. In that case, it could capture the variables used in the formula and treat them appropriately to make the data. 

For example: 

```r
wflow <- 
  workflow() %>% 
  add_formula(Reaction ~ Days + (Days | Subject)) %>% 
  add_model(lmer_mod)
```

would determine with variables to use on each side of the `~` and use `all.vars()` to get the appropriate columns (it's actually not that simple but there are functions in `recipes` already). 

Based on the information in the model specification, the appropriate `hardhat` blueprint could be used. 

(aside: this would not eliminate the need of the extra formula slot of `add_model()` since it would be used with a recipe)

juliasilge:
Closes #290.

This PR adds the field for engine-specific encodings (only dummy/indicator variables for now).

For example, the encoding options for ranger are:

```r
set_encoding(
  model = "rand_forest",
  eng = "ranger",
  mode = "regression",
  options = list(predictor_indicators = FALSE)
)
```
While the encoding options for vanilla logistic regression are:

```r
set_encoding(
  model = "logistic_reg",
  eng = "glm",
  mode = "classification",
  options = list(predictor_indicators = TRUE)
)
```

These changes depend on handling the names for data arguments implemented in #315 and #316.

These encodings can be used in [workflows](https://github.com/tidymodels/workflows) so that the user experiences the same behavior around dummy variable creation in both parsnip and workflows.

I am pretty confident that `predictor_indicators = TRUE / FALSE` is correct for all the model + engine combinations except for liquidSVM. I was having trouble getting output from those models and could use some double-checking.


juliasilge:
This PR also fixes the function used with Spark decision trees 🌳 for regression.

R/mars_data.R
```diff
@@ -47,6 +47,13 @@ set_fit(
   )
 )
 
+set_encoding(
+  model = "mars",
+  eng = "earth",
+  mode = "regression",
+  options = list(predictor_indicators = TRUE)
```
topepo, line 54: Can you change the MARS models to have `predictor_indicators = FALSE`? I believe this will work and will help some other things. The `earth` package uses the `call` object for some functions (like plotting) and having the original variable there would work better.

Issue #235, opened 2019-11-15
Add a function that can query the defaults of an underlying engine


DavisVaughan:
In `get_model_env()` we document the original parameter names of the underlying engine

```r
> parsnip::get_model_env()$decision_tree_args
# A tibble: 6 x 5
  engine parsnip         original               func             has_submodel
  <chr>  <chr>           <chr>                  <list>           <lgl>       
1 rpart  tree_depth      maxdepth               <named list [2]> FALSE       
2 rpart  min_n           minsplit               <named list [2]> FALSE       
3 rpart  cost_complexity cp                     <named list [2]> FALSE       
4 C5.0   min_n           minCases               <named list [2]> FALSE       
5 spark  tree_depth      max_depth              <named list [2]> FALSE       
6 spark  min_n           min_instances_per_node <named list [2]> FALSE  
```

but we don't say anything about the default values of those parameters.

If we record the function that those original parameters go along with, we could have a function that does a dynamic lookup like this:

```r
lookup_default <- function(fn, arg) {
  arg_list <- formals(fn)
  
  has_arg <- arg %in% names(arg_list)
  if (!has_arg) {
    stop("Argument: `", arg, "` does not exist in the function.", call. = FALSE)
  }
  
  default_value <- formals(fn)[[arg]]
  default_value
}

lookup_default(rpart::rpart.control, "cp")
#> [1] 0.01

lookup_default(rpart::rpart.control, "cp_oops")
#> Error: Argument: `cp_oops` does not exist in the function.
```

It would be nice to expose this information to users in the form of a tibble like `parsnip::get_model_env()$decision_tree_args` that also has columns for the function that the original parameters come from, and the default values.

apreshill:
Also, would be great to be able to pass these defaults to `tune::fit_resamples()`. So if this code:

```
library(tune)
dt_by_fold <- fit_resamples(downsampled, 
                            model = dt_learner, 
                            resamples = folds, 
                            metrics = metric_set(roc_auc),
                            control = control_resamples(save_pred = TRUE, 
                                                        save_defaults = TRUE))
```

Produced this kind of output:
```
> collect_metrics(dt_by_fold)
# A tibble: 10 x 7
   cost_complexity min_n .metric .estimator  mean     n std_err
             <dbl> <int> <chr>   <chr>      <dbl> <int>   <dbl>
 1            0.01   20  roc_auc binary     0.650    10  0.0157
```

sckinta:
[...]

PR #321, opened 2020-05-27
Add defaults to engine specific params in model docs

### LINKED ISSUE #211: Create a reference / cheat sheet mapping engine-specific parameters to parsnip ones?
As I'm going through converting algorithm-specific logic to generalized parsnip model specifications, one thing I am thinking that could be useful would be to create one or more reference tables or "cheat sheets" in the parsnip docs describing how to map from implemention-specific parameters (e.g. "eta" in xgboost) to parsnip model parameters (e.g. "learn_rate" in parsnip::boost_tree()).

The simplest form would probably just be a collection of tables (one per model), with columns for each supported engine + a final column corresponding to the parsnip parameter names.

Since this could take a bit of time to add, perhaps if you think this sort of thing would be useful, you could decide on an appropriate location/format, and add placeholders.

Then, myself and other users could help fill it in as we use the models.

juliasilge:
This PR closes #211.

What I have so far here is my initial idea of how to add the default values for each engine specific parameter into the tables in the docs. These values are only needed for `devtools::document()` so I am thinking let's make little internal functions via `paste0()` rather than `get_model_env()` and add all these values via some `set_...` function and keep them around.

What are your thoughts?

Current results for the one model implemented right now:

``` r
library(parsnip)
convert_args("decision_tree")
#> Joining, by = c("model", "engine", "parsnip", "original")
```

<table>
<thead>
<tr class="header">
<th style="text-align: left;"><strong>parsnip</strong></th>
<th style="text-align: left;"><strong>rpart</strong></th>
<th style="text-align: left;"><strong>C5.0</strong></th>
<th style="text-align: left;"><strong>spark</strong></th>
</tr>
</thead>
<tbody>
<tr class="odd">
<td style="text-align: left;">tree_depth</td>
<td style="text-align: left;">maxdepth (30)</td>
<td style="text-align: left;">NA</td>
<td style="text-align: left;">max_depth (5)</td>
</tr>
<tr class="even">
<td style="text-align: left;">min_n</td>
<td style="text-align: left;">minsplit (20)</td>
<td style="text-align: left;">minCases (2)</td>
<td style="text-align: left;">min_instances_per_node (1)</td>
</tr>
<tr class="odd">
<td style="text-align: left;">cost_complexity</td>
<td style="text-align: left;">cp (0.01)</td>
<td style="text-align: left;">NA</td>
<td style="text-align: left;">NA</td>
</tr>
</tbody>
</table>

<sup>Created on 2020-05-27 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


apreshill:
@juliasilge I believe this also would close https://github.com/tidymodels/parsnip/issues/235? :tada:

juliasilge:
For now I'm only adding these to the tables in the docs, with functions that are only run during `devtools::document()`. So no user-facing functions quite yet.

Users could look at the docs and get the values but we haven't moved quite all the way to a dynamic lookup.

juliasilge:
We're going to have to make a call on what to do with the `rand_forest()` table, for `mtry` and `min_n`.

For ranger's `mtry`:
```
Number of variables to possibly split at in each node. Default is the (rounded down) square root
of the number variables. Alternatively, a single argument function returning an integer, given the 
number of independent variables.
```

For ranger's `min.node.size`:
```
Minimal node size. Default 1 for classification, 5 for regression, 3 for survival, and 10 for 
probability.
```
Neither of these are available via `formals()`.

For randomForest, the function argument themselves are defined as:

```r
mtry=if (!is.null(y) && !is.factor(y))
             max(floor(ncol(x)/3), 1) else floor(sqrt(ncol(x)))

nodesize = if (!is.null(y) && !is.factor(y)) 5 else 1
```
I think I'm going to put some of these in manually.

topepo:
Yes. I would add them as expressions to your data structure. Maybe something like "1C|5R|3S" with a legend on the bottom?

juliasilge:
The issues around `rand_forest()` and SVM make me think #235 is going to be **quite** difficult.

juliasilge:
The SVM options are less than ideal. They are hard-coded mostly, and I dug around in the liquidSVM code and came up with... this:

```r
> convert_args("svm_rbf")
```

|**parsnip** |**kernlab**   |**liquidSVM**    |
|:-----------|:-------------|:----------------|
|cost        |C (1)         |lambdas (varies) |
|rbf_sigma   |sigma (1)     |gammas (varies)  |
|margin      |epsilon (0.1) |NA               |

Hmmm

juliasilge:
I updated the `sigma` value for kernlab because although `rbfdot()` has a default of 1, during training the value used depends on the data.

PR #328, opened 2020-06-07
xgboost notes about sample_prop

topepo:
Closes tidymodels/dials#111


topepo:
Also added some improvements to the `parsnip` documentation around `mixture` for `glmnet` models.

PR #333, opened 2020-06-19
Write out iris with modeldata::hpc_data

simonpcouch:
Quite a few of the changes so far were global find-and-replace, with some manual bug fixes. Still quite a few to go—will come back to this over the weekend.


simonpcouch:
Okay, I think this is good to go! Let me know if the `test-coverage` error is an issue—seem like unrelated brew things to me.

Worth mentioning that `tests/testthat/test_varying.R` use `tests/testthat/recipes_examples.RData`, which contain some recipes specified on `iris`—having trouble tracking down the script that created the file, though.

PR #337, opened 2020-06-29
Add MacOS testing back to parsnip

juliasilge:
This PR adds testing on MacOS via GH actions back, and better aligns the R CMD check with current `usethis::use_tidy_github_actions()`.


EmilHvitfeldt:
For the macOS miniconda, you need to set the condo argument in conda_create to `'miniconda/bin/conda'`

PR #373, opened 2020-09-29
Changes to allow sparsity predictor representations

topepo:
Related to discussion in tidymodels/tidymodels#42

* Adds a new encoding fields called `allow_sparse_x` that is TRUE for `glmnet`, `xgboost`, and `ranger` models. 

* `ranger` now uses their x/y interface. Tests indicates that there is no difference. 

* Some refactoring of `xgboost` code. 

Only changes the use of `fit_xy()`. For example:  

``` r
library(tidymodels)
#> ── Attaching packages ─────────────────────────────────────────────────────────── tidymodels 0.1.1 ──
#> ✓ broom     0.7.0          ✓ recipes   0.1.13    
#> ✓ dials     0.0.9          ✓ rsample   0.0.8     
#> ✓ dplyr     1.0.2          ✓ tibble    3.0.3     
#> ✓ ggplot2   3.3.2          ✓ tidyr     1.1.2     
#> ✓ infer     0.5.2          ✓ tune      0.1.1.9000
#> ✓ modeldata 0.0.2          ✓ workflows 0.2.0     
#> ✓ parsnip   0.1.3.9000     ✓ yardstick 0.0.7     
#> ✓ purrr     0.3.4
#> ── Conflicts ────────────────────────────────────────────────────────────── tidymodels_conflicts() ──
#> x purrr::discard() masks scales::discard()
#> x dplyr::filter()  masks stats::filter()
#> x dplyr::lag()     masks stats::lag()
#> x recipes::step()  masks stats::step()
library(Matrix)
#> 
#> Attaching package: 'Matrix'
#> The following objects are masked from 'package:tidyr':
#> 
#>     expand, pack, unpack

xgb_spec <-
  boost_tree(trees = 10) %>% 
  set_engine("xgboost") %>% 
  set_mode("regression")

mtcar_x <- mtcars[, -1]
mtcar_mat <- as.matrix(mtcar_x)
mtcar_smat <- Matrix(mtcar_mat, sparse = TRUE)

set.seed(1)
from_df <- xgb_spec %>% fit_xy(mtcar_x, mtcars$mpg)
set.seed(1)
from_mat <- xgb_spec %>% fit_xy(mtcar_mat, mtcars$mpg)
set.seed(1)
from_sparse <- xgb_spec %>% fit_xy(mtcar_smat, mtcars$mpg)

all.equal(from_df$fit, from_mat$fit)
#> [1] TRUE
all.equal(from_df$fit, from_sparse$fit)
#> [1] TRUE
```

<sup>Created on 2020-09-29 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>


juliasilge:
Here is some benchmarking FWIW:

``` r
library(glmnet)
#> Loading required package: Matrix
#> Loaded glmnet 4.0-2
library(parsnip)

n <- 1e5 
y <- sample(0:1, n, replace = TRUE) 
x1 <- rnorm(n)
x2 <- sample(1:400, n, replace = TRUE)
X1 <- sparse.model.matrix(~ x1 + factor(x2))
X2 <- model.matrix(~ x1 + factor(x2))

lasso_spec <- linear_reg(mixture = 1) %>%
  set_mode("regression") %>%
  set_engine("glmnet")

bench::mark(
  check = FALSE, iterations = 10,
  lasso_spec %>% fit_xy(x = X2, y = y),
  lasso_spec %>% fit_xy(x = X1, y = y),
  glmnet(x = X2, y = y),
  glmnet(x = X1, y = y)
)
#> Warning: Some expressions had a GC in every iteration; so filtering is disabled.
#> # A tibble: 4 x 6
#>   expression                                min   median `itr/sec` mem_alloc
#>   <bch:expr>                           <bch:tm> <bch:tm>     <dbl> <bch:byt>
#> 1 lasso_spec %>% fit_xy(x = X2, y = y)    6.76s    6.81s     0.147   931.9MB
#> 2 lasso_spec %>% fit_xy(x = X1, y = y) 276.37ms  282.1ms     3.49     16.7MB
#> 3 glmnet(x = X2, y = y)                   6.57s    6.68s     0.149     623MB
#> 4 glmnet(x = X1, y = y)                164.27ms 166.48ms     5.75     16.4MB
#> # … with 1 more variable: `gc/sec` <dbl>
```

<sup>Created on 2020-10-02 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

Looking so great! We can see the parsnip overhead here, and the huge improvement to being able to use a sparse data structure.

juliasilge (changes requested):
Super happy about these changes. 😄

In the benchmarking I have been working on, this makes a big difference for glmnet especially in terms of fitting time.

We don't have this change/option documented anywhere in user-facing docs, and I think it should be surfaced at least mildly somewhere (not be entirely undocumented). Some options:

- In `fit.R`, change the `x` parameter to something like

```out
A matrix, sparse matrix, or data frame of predictors. Only some models have support for sparse matrix input. 
See `parsnip::get_encoding()` for details.
```

- Add something to the **Details** of the models that have the support.

I think I like option 1 better.

DavisVaughan:
Quick thought, only version 0.12.0 of ranger supports the x/y interface (the newest one), so you may want to add a version requirement to ranger in Suggests https://cran.r-project.org/web/packages/ranger/NEWS

NAMESPACE
```diff
@@ -130,6 +130,8 @@ export(linear_reg)
 export(logistic_reg)
 export(make_classes)
 export(mars)
+export(maybe_data_frame)
+export(maybe_matrix)
```
DavisVaughan, line 134: Is this for parsnip-adjacent packages to use?

NEWS.md
```diff
@@ -2,6 +2,8 @@
 
 * `show_engines()` will provide information on the current set for a model. 
 
+* For three models, enable sparse matrix use via `fit_xy()` (#373).
```
DavisVaughan, line 5: Do you want to mention what the 3 are?

R/boost_tree.R
```diff
@@ -415,6 +391,37 @@ xgb_pred <- function(object, newdata, ...) {
...
+  if (is.factor(y)) {
+    y <- as.numeric(y) - 1
+  }
+
+  if (!inherits(x, "xgb.DMatrix")) {
+    if (validation > 0) {
+      trn_index <- sample(1:n, size = floor(n * (1 - validation)) + 1)
+      wlist <-
+        list(validation = xgboost::xgb.DMatrix(x[-trn_index, ], label = y[-trn_index], missing = NA))
+      dat <- xgboost::xgb.DMatrix(x[trn_index, ], label = y[trn_index], missing = NA)
```
DavisVaughan, line 412:
```suggestion
        list(validation = xgboost::xgb.DMatrix(x[-trn_index, , drop = FALSE], label = y[-trn_index], missing = NA))
      dat <- xgboost::xgb.DMatrix(x[trn_index, , drop = FALSE], label = y[trn_index], missing = NA)
```

for 1 column `x` matrices?

R/fit.R
```diff
@@ -357,21 +368,33 @@ check_xy_interface <- function(x, y, cl, model) {
         )
       )
 
-  # Determine the `fit()` interface
-  matrix_interface <- !is.null(x) & !is.null(y) && is.matrix(x)
+
+  if (sparse_ok) {
+    matrix_interface <- !is.null(x) & !is.null(y) && (is.matrix(x) | sparse_x)
+  } else {
+    matrix_interface <- !is.null(x) & !is.null(y) && is.matrix(x)
```
DavisVaughan, line 375:
```suggestion
    matrix_interface <- !is.null(x) && !is.null(y) && (is.matrix(x) | sparse_x)
  } else {
    matrix_interface <- !is.null(x) && !is.null(y) && is.matrix(x)
```

I think this was a typo?

R/fit.R
```diff
@@ -357,21 +368,33 @@ check_xy_interface <- function(x, y, cl, model) {
...
 
-  if (inherits(model, "surv_reg") &&
-      (matrix_interface | df_interface))
+  if (inherits(model, "surv_reg") && (matrix_interface | df_interface)) {
     rlang::abort("Survival models must use the formula interface.")
+  }
 
-  if (matrix_interface)
+  if (matrix_interface) {
     return("data.frame")
```
DavisVaughan, line 385: Just checking that this was correct beforehand- `if(matrix_interface)` then `return("data.frame")`? Not `return("matrix")`? Just smelled a little funny

DavisVaughan (approved):

NAMESPACE
topepo: yes.

R/boost_tree.R
topepo: yes

R/fit.R
topepo: yeah. I'll fix that

R/fit.R
topepo: I think so. I've changed it and the tests in `parsnip` and `extratests` pass so it might have been a typo.

juliasilge (approved):

PR #377, opened 2020-10-10
Make arguments robust to values outside of data dimensions

### LINKED ISSUE #184: "harden" parsnip to potential dimensionality bugs
Although the [data descriptors](https://tidymodels.github.io/parsnip/reference/descriptors.html) can solve this, we should look at capping the value of any tuning parameters that are based on either `nrow()` or `ncol()`. 

For example, with `mtry` in random forests, we should, somewhere use `mtry = max(mtry, ncol(x))`.

topepo:
closes #184 

close #270 

Uses the `translate()` function to make some arguments expressions instead of specific values. For example: 

```r
> nearest_neighbor(neighbors = 1000) %>%
+   set_engine("kknn") %>%
+   set_mode("regression") %>% 
+   translate()
K-Nearest Neighbor Model Specification (regression)

Main Arguments:
  neighbors = 1000

Computational engine: kknn 

Model fit template:
kknn::train.kknn(formula = missing_arg(), data = missing_arg(), 
    ks = min(1000, nrow(data) - 5))
```

For C5.0 and xgboost models, the wrapper functions were used to bound the values based on the actual data dimensions.


juliasilge:
I think the reason I was trying to use the data descriptors back in June instead of just `nrow()` or `ncol()` was because I was concerned about the dummy variable column counting. Do you think that is worth worrying about? 

``` r
library(parsnip)

rf_mod <- 
  rand_forest(mtry = 20) %>% 
  set_mode("regression") %>% 
  set_engine("ranger")

translate(rf_mod)
#> Random Forest Model Specification (regression)
#> 
#> Main Arguments:
#>   mtry = 20
#> 
#> Computational engine: ranger 
#> 
#> Model fit template:
#> ranger::ranger(x = missing_arg(), y = missing_arg(), case.weights = missing_arg(), 
#>     mtry = min(~20, ncol(x)), num.threads = 1, verbose = FALSE, 
#>     seed = sample.int(10^5, 1))

fit(rf_mod, circumference ~ ., data = Orange)
#> parsnip model object
#> 
#> Fit time:  17ms 
#> Ranger result
#> 
#> Call:
#>  ranger::ranger(x = maybe_data_frame(x), y = y, mtry = min(~20,      ncol(x)), num.threads = 1, verbose = FALSE, seed = sample.int(10^5,      1)) 
#> 
#> Type:                             Regression 
#> Number of trees:                  500 
#> Sample size:                      35 
#> Number of independent variables:  2 
#> Mtry:                             2 
#> Target node size:                 5 
#> Variable importance mode:         none 
#> Splitrule:                        variance 
#> OOB prediction error (MSE):       125.2415 
#> R squared (OOB):                  0.9621042
```

<sup>Created on 2020-10-13 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0.9001)</sup>

The variable `Tree` is a factor so really it would be OK to go up to `mtry = 5` here. It gets cut down to 2, though.

juliasilge:
ranger was a bad example there, but I am still trying to figure out where the `ncol()` is evaluated.

juliasilge:
OK, I am convinced now! 😄 Moving so many things to `x`/`y` has really had some nice outcomes and made things like this more doable.

This behavior should probably be documented somewhere (not too obtrusive) so it is not a total surprise to people. I'll make a suggestion.

juliasilge:
Actually, I changed my mind on adding anything to the documentation. I think from how these arguments are documented it is clear enough what they are supposed to be. ✅

juliasilge (approved):

PR #391, opened 2020-11-09
Change redirect url

tanho63:
Just proposing that the redirect should go straight to the find/parsnip page which lists all models (saves 1 click)


juliasilge (approved):
Thank you so much! 🙌

PR #400, opened 2020-12-08
Document `allow_sparse_x`

### LINKED ISSUE #397: help for set_encoding
We need to document `allow_sparse_x`

juliasilge:
Closes #397 

This PR adds the fourth option that currently exists for `set_encoding()` to the docs. It's similar to how I described it in tidymodels/tidymodels.org#204.


juliasilge:
The failure on R-devel is because of a recent change to [check the environment](https://github.com/wch/r-source/commit/4329ff7d48dcb745028b5cc1c651e373b4e7258c) when comparing things like the keras objects. I suggest we wait for a while to decide to whether to address this; it's possible that moving the whole package to testthat 3rd edition would avoid the problem.

PR #411, opened 2021-01-14
allow 'objective' for xgboost to be passed as an engine argument

topepo:


R/boost_tree.R
```diff
@@ -312,6 +312,8 @@ xgb_train <- function(
   min_child_weight = 1, gamma = 0, subsample = 1, validation = 0,
   early_stop = NULL, ...) {
 
+  others <- list(...)
```
DavisVaughan, line 315: Should it be a main argument to `xgb_train()` that is defaulted to `NULL`? If `NULL`, then we auto choose the objective, otherwise we let whatever the user passed through be used.

topepo: Yeah. That is probably a better approach since it is more explicit.

DavisVaughan (approved):

jaredlander:
Does this have something to do with the newish warning message in xgboost about the default evaluation metric for binary:logistic being changed from 'error' to 'logloss' as shown below?

> WARNING: amalgamation/../src/learner.cc:1061: Starting in XGBoost 1.3.0, the default evaluation metric used with the objective 'binary:logistic' was changed from 'error' to 'logloss'. Explicitly set eval_metric if you'd like to restore the old behavior.

juliasilge:
No, I'm pretty sure not. We [set the objective](https://github.com/tidymodels/parsnip/blob/master/R/boost_tree.R#L335-L345) but I don't think we do anything with the internal evaluation metric for xgboost.

jaredlander:
Looking at the fitted model after fitting I see this

```
call:
  xgboost::xgb.train(params = list(eta = 0.3, max_depth = 4, gamma = 0, 
    colsample_bytree = 1, min_child_weight = 1, subsample = 0.6, 
    objective = "binary:logistic"), data = x$data, nrounds = 100, 
    watchlist = x$watchlist, verbose = 0, nthread = 1)
```

It looks like you set the `watchlist` argument: `watchlist = x$watchlist`

When the `watchlist` is set I think `{xgboost}` automatically chooses an `eval_metric` which in turn causes that warning.

PR #413, opened 2021-01-17
Control classes

### LINKED ISSUE #406: check control object class
`workflows:::fit.model_spec()` should check to make sure that the right type of control object is passed in. See 
 tidymodels/workflows#89

topepo:
adds pak-based checking

closes #406


topepo:
Merging;  failures are from missing `lincoin` binaries.

PR #415, opened 2021-01-17
augment method for model_fit objects

topepo:
We should add `augment.workflows()` too. I don't know if this will be re-usable.


R/augment.R
```diff
@@ -0,0 +1,73 @@
...
+#' `augment()` will add column(s) for predictions to the given data.
+#'
+#' For regression models, a `.pred` column is added. If `x` was created using
+#' [fit()] and `new_data` contains the outcome column, a `.resid` column is
+#' also added.
+#'
+#' For classification models, the results include a column called `.pred_class`
+#' as well as class probability columns named `.pred_{level}`.
+#' @param x A `model_fit` object produced by [fit()] or [fit_xy()].
+#' @param new_data A data frame or matrix.
```
DavisVaughan, line 12: Is it weird that `broom::augment()` methods always use `data` and this uses `new_data`?

R/augment.R
```diff
@@ -0,0 +1,73 @@
...
+      dplyr::bind_cols(
+        predict(x, new_data = new_data)
+      )
+    if (length(x$preproc$y_var) > 0) {
+      y_nm <- x$preproc$y_var
+      if (any(names(new_data) == y_nm)) {
+        new_data <- dplyr::mutate(new_data, .resid = !!rlang::sym(y_nm) - .pred)
+      }
+    }
+  } else {
```
DavisVaughan, line 64: May want to do `else if (mode == classification)` and then catch any other modes with a final `else { abort("unknown mode") }`. I imagine we might add more modes in the future and this would catch that

R/augment.R
```diff
@@ -0,0 +1,73 @@
+#' Augment data with predictions
+#'
+#' `augment()` will add column(s) for predictions to the given data.
+#'
+#' For regression models, a `.pred` column is added. If `x` was created using
+#' [fit()] and `new_data` contains the outcome column, a `.resid` column is
```
DavisVaughan, line 6: > a `.resid` column is also added

I think this is a _little_ surprising. Maybe I don't like it because it means that `augment()` with regression models can output a different number of columns depending on `new_data`. I guess it is fine to keep it if you think it is useful, but I think it is just as easy to create the `.resid` column afterwards if you really want it. I'm fine with either way.

DavisVaughan (approved):

juliasilge:
Discussions we've had about this feature in the past have really emphasized how we want to discourage the practice of predicting on the training set (including the choice of using `new_data` as the argument name). Right now the examples show doing exactly that, though -- re-predicting the training set. 

Would it be better to make a slightly more complex example with a little test set?

R/augment.R
juliasilge: I think it's pretty useful

PR #427, opened 2021-02-08
Refresh pkgdown

juliasilge:
We are in a weird pkgdown state, because the `DESCRIPTION` file never got bumped to a dev version number.

Can we make a few changes here to clean things up and

- move some functions to `internal` and some to the pkgdown outline, as I have done here?
- bump the version number (TBD, but I will do so)?


juliasilge:
I'm actually going to merge this into master so the "main"/regular pkgdown site gets updated, then change the version number on master.

PR #438, opened 2021-03-01
Add kernlab engine for svm_linear()

### LINKED ISSUE #336: linear svm models
This issues https://github.com/tidymodels/parsnip/issues/35 talks about linear SVM models but we only have `svm_poly()` and `svm_rbf()`.

I can see how `svm_linear()` would be a special case of `svm_poly()` but is there still room for `svm_linear()` for models that doesn't do polynomials?

juliasilge:
Closes #336 

This PR adds a second engine for `svm_linear()`. We already have `"LiblineaR"` and this PR adds `"kernlab"`.

``` r
library(tidymodels)

data(two_class_dat, package = "modeldata")
example_split <- initial_split(two_class_dat, prop = 0.99)
example_train <- training(example_split)
example_test  <-  testing(example_split)

set.seed(123)
mod <- svm_linear() %>%
  set_engine("kernlab") %>%
  fit(Class ~ ., example_train)
#>  Setting default kernel parameters

mod
#> parsnip model object
#> 
#> Fit time:  856ms 
#> Support Vector Machine object of class "ksvm" 
#> 
#> SV type: C-svc  (classification) 
#>  parameter : cost C = 1 
#> 
#> Linear (vanilla) kernel function. 
#> 
#> Number of Support Vectors : 358 
#> 
#> Objective Function Value : -355.0963 
#> Training error : 0.179847 
#> Probability model included.

predict(mod, new_data = example_test)
#> # A tibble: 7 x 1
#>   .pred_class
#>   <fct>      
#> 1 Class2     
#> 2 Class1     
#> 3 Class2     
#> 4 Class1     
#> 5 Class1     
#> 6 Class1     
#> 7 Class2
predict(mod, new_data = example_test, type = "prob")
#> # A tibble: 7 x 2
#>   .pred_Class1 .pred_Class2
#>          <dbl>        <dbl>
#> 1        0.457       0.543 
#> 2        0.850       0.150 
#> 3        0.172       0.828 
#> 4        0.980       0.0203
#> 5        0.698       0.302 
#> 6        0.983       0.0166
#> 7        0.439       0.561
predict(mod, new_data = example_test, type = "raw")
#> [1] Class2 Class1 Class2 Class1 Class1 Class1 Class2
#> Levels: Class1 Class2
```

<sup>Created on 2021-03-01 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

Unlike the `"LiblineaR"` engine, the `"kernlab"` engine does support class probabilities.

### DIFF
--- a/R/svm_linear.R
+++ b/R/svm_linear.R
@@ -29,7 +29,7 @@
 #' The model can be created using the `fit()` function using the
 #'  following _engines_:
 #' \itemize{
-#' \item \pkg{R}:  `"LiblineaR"` (the default)
+#' \item \pkg{R}:  `"LiblineaR"` (the default) or `"kernlab"`
 #' }
 #'
 #'
@@ -173,6 +173,16 @@ translate.svm_linear <- function(x, engine = x$engine, ...) {
     }
   }
 
+  if (x$engine == "kernlab") {
+
+    # unless otherwise specified, classification models predict probabilities
+    if (x$mode == "classification" && !any(arg_names == "prob.model"))
+      arg_vals$prob.model <- TRUE
+    if (x$mode == "classification" && any(arg_names == "epsilon"))
+      arg_vals$epsilon <- NULL
+
+  }
+
   x$method$fit$args <- arg_vals
 
   # worried about people using this to modify the specification
@@ -189,5 +199,9 @@ check_args.svm_linear <- function(object) {
 
 svm_linear_post <- function(results, object) {
   results$predictions
+}
+
+svm_reg_linear_post <- function(results, object) {
+  results[,1]
 }
 
--- a/tests/testthat/test_svm_linear.R
+++ b/tests/testthat/test_svm_linear.R
@@ -14,6 +14,7 @@ hpc <- hpc_data[1:150, c(2:5, 8)]
 test_that('primary arguments', {
   basic <- svm_linear(mode = "regression")
   basic_LiblineaR <- translate(basic %>% set_engine("LiblineaR"))
+  basic_kernlab <- translate(basic %>% set_engine("kernlab"))
 
   expect_equal(
     object = basic_LiblineaR$method$fit$args,
@@ -26,11 +27,21 @@ test_that('primary arguments', {
     )
   )
 
+  expect_equal(
+    object = basic_kernlab$method$fit$args,
+    expected = list(
+      x = expr(missing_arg()),
+      data = expr(missing_arg()),
+      kernel = "vanilladot"
+    )
+  )
+
 })
 
 test_that('engine arguments', {
 
   LiblineaR_type <- svm_linear(mode = "regression") %>% set_engine("LiblineaR", type = 12)
+  kernlab_cv <- svm_linear(mode = "regression") %>% set_engine("kernlab", cross = 10)
 
   expect_equal(
     object = translate(LiblineaR_type, "LiblineaR")$method$fit$args,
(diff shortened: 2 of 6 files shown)


R/svm_linear.R
```diff
@@ -173,6 +173,16 @@ translate.svm_linear <- function(x, engine = x$engine, ...) {
     }
   }
 
+  if (x$engine == "kernlab") {
+
+    # unless otherwise specified, classification models predict probabilities
+    if (x$mode == "classification" && !any(arg_names == "prob.model"))
+      arg_vals$prob.model <- TRUE
+    if (x$mode == "classification" && any(arg_names == "epsilon"))
+      arg_vals$epsilon <- NULL
```
DavisVaughan, line 182: Why do you force a user supplied epsilon to `NULL`?

DavisVaughan: Oh is it because it maps to `margin` and that should be used instead? Should you do the same thing with `"C"` then?

tests/testthat/test_svm_linear.R
```diff
@@ -26,11 +27,21 @@ test_that('primary arguments', {
...
+      kernel = "vanilladot"
+    )
+  )
+
 })
 
 test_that('engine arguments', {
 
   LiblineaR_type <- svm_linear(mode = "regression") %>% set_engine("LiblineaR", type = 12)
+  kernlab_cv <- svm_linear(mode = "regression") %>% set_engine("kernlab", cross = 10)
```
DavisVaughan, line 44: So this `cross` arg seems to do internal cross validation, using accuracy as a metric for classification. When doing binary classification, do you happen to know if it is considering the first or second level as the event level? Does that matter at all here?

DavisVaughan (approved):

R/svm_linear.R
juliasilge: The `epsilon` arg maps to `margin`, right, which is regression only. This is some error checking [that is the same as what's in `svm_rbf()`](https://github.com/tidymodels/parsnip/blob/963ba95d03273820843ac96b63e7f910531e6eb8/R/svm_rbf.R#L143).

DavisVaughan: > which is regression only

oh that is what i missed!

tests/testthat/test_svm_linear.R
juliasilge: It looks like it is doing the same thing as what parsnip is doing, with no problems like what xgboost had:

``` r
library(parsnip)
library(tidyverse)
library(kernlab)
#> 
#> Attaching package: 'kernlab'
#> The following object is masked from 'package:purrr':
#> 
#>     cross
#> The following object is masked from 'package:ggplot2':
#> 
#>     alpha

data("PimaIndiansDiabetes", package = "mlbench")
df <- PimaIndiansDiabetes %>%
  mutate(diabetes = fct_relevel(diabetes, 'pos'))


set.seed(234)
parsnip_fit <- svm_linear(mode = "classification") %>% 
  set_engine("kernlab", cross = 10) %>%
  fit(diabetes ~ ., df)
#>  Setting default kernel parameters

set.seed(234)
kernlab_fit <- ksvm(diabetes ~ ., data = df, kernel = "vanilladot", cross = 10)
#>  Setting default kernel parameters

identical(parsnip_fit$fit@alpha, kernlab_fit@alpha)
#> [1] TRUE
```

<sup>Created on 2021-03-03 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

If there is a problem (I don't think there is), it would apply to all the kernlab engines and we should open a new issue and fix it in a new PR.

DavisVaughan: Sounds good, just wanted to check

Issue #444, opened 2021-03-16
add `survival_reg` model spec 


hfrick:
move `survival_reg` from {censored} to parsnip. See also https://github.com/EmilHvitfeldt/censored/issues/2#issuecomment-788429365

PR #445, opened 2021-03-16
restrict `fit_xy()` from working for censored regression models

### LINKED ISSUE #442: Restrict fit_xy() from working with censored regression models
Censored regression models should be fitted via a formula using `fit()` rather than `fit_xy()`

hfrick:
Closes #442 

Models with the new mode "censored regression" can't use `fit_xy()`, they should use `fit()`.

`surv_reg` models have also been forced to use the formula interface `fit()`. The check for that has been moved into `fit_xy()` from `check_xy_interface()` which checks the details of `x` and `y`. The corresponding tests for `check_xy_interface()` were removed. The tests for an error when using `fit_xy()` with a `surv_reg` model are untouched in `test_sruv_reg_survreg.R` and `test_surv_reg_flexsurv.R`.

The corresponding tests for models for censored regression are to be added when we have the corresponding model specs in parsnip (#444).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,8 @@
 
 * Column names for `x` are now required when `fit_xy()` is used. (#398)
 
+* Censored regression models cannot use `fit_xy()` (use `fit()`). (#442)
+
 
 # parsnip 0.1.4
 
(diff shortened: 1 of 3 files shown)


topepo (approved):
We can add test cases for this when we move the model definitions to `parsnip` from `censored`.

hfrick:
PR #449 adds the model spec for `survival_reg` which uses the new mode and includes a test on `fit_xy()` throwing an error when used with this mode.

PR #455, opened 2021-03-24
Model spec for proportional hazards models

### LINKED ISSUE #451: Add model spec for `proportional_hazards`
use and rename {censored}'s `cox_reg`. make sure that Cox Model is mentioned appropriately in the documentation so it becomes googleable cc @EmilHvitfeldt

hfrick:
Closes #451

This is a joint PR with https://github.com/EmilHvitfeldt/censored/pull/35. This PR on parsnip should be merged first.

On the parsnip side:
* added model specification for `proportional_hazards` models which previously lived as `cox_reg` in {censored}.

On the censored side:
* renamed `cox_reg` to `proportional_hazards` for the engine code and removed the model specification


topepo (changes requested):
Please at a note in the NEWS file.

topepo (approved):

PR #456, opened 2021-03-25
Revised model documentation

topepo:
This restructures the model documentation files so that 

* Each engine has a separate man file (instead of being in the "Engine Details" subsection)
* Engines from `parsnip`-adjacent packages are automatically discovered and linked if they use the same conventions. 
* The engines have more information than before (e.g., preprocessing requirements, references, etc). 

Each engine-specific man file links to a Rmd file in `man/rmd`. The naming convention of the man pages is critical for finding them automatically. 

**Draft PR**: only changes for `linear_reg()` and you'll need to run `pak::pak("tidymodels/multilevelmod@doc-test")` to see how other engines from other packages show up.


R/aaa_models.R
```diff
@@ -911,3 +911,64 @@ get_encoding <- function(model) {
   }
   res
 }
+
+#' Tools for documenting packages
+#' @param mod A character string for the model file
+#' @param pkg The package that contains the model file
```
DavisVaughan, line 917: I don't see an argument for this

R/aaa_models.R
```diff
@@ -911,3 +911,64 @@ get_encoding <- function(model) {
...
+  res
+}
+
+search_for_engine_docs <- function(mod) {
+  all_deps <- get_from_env(paste0(mod, "_pkgs"))
+  all_deps <- unlist(all_deps$pkg)
+  all_deps <- unique(c("parsnip", all_deps))
+  excl <- c("stats", "magrittr")
+  all_deps <- all_deps[!(all_deps %in% excl)]
+  res <- purrr::map(all_deps, parsnip:::find_details_topics, mod = mod)
```
DavisVaughan, line 957:
```suggestion
  res <- purrr::map(all_deps, find_details_topics, mod = mod)
```

R/aaa_models.R
```diff
@@ -911,3 +911,64 @@ get_encoding <- function(model) {
   }
   res
 }
+
+#' Tools for documenting packages
```
DavisVaughan, line 915: Probably useful to add a few lines about how you are supposed to document files that you want to show up dynamically (i.e. with `@name details-`)

R/linear-reg-doc-glmnet.R
```diff
@@ -0,0 +1,9 @@
+#' Linear regression via glmnet
+#'
+#' `glmnet()` uses regularized least squares to fit models with numeric outcomes.
+#'
+#' @includeRmd man/rmd/linear-reg-glmnet.Rmd details
+#'
+#' @name details-linear-reg-glmnet
+#' @keywords internal
+NULL
```
DavisVaughan, line 9: In the long run, it is probably useful to move _all_ of the glmnet linear-reg code into `R/linear-reg-glmnet.R` and then this could live at the top there

DavisVaughan (commented):
I like the idea! Mainly we'd just need some good documentation about how this works (would also be nice to mention that `_` gets subbed for `-` when going from model name -> file name)

R/linear-reg-doc-lm.R
```diff
@@ -1,6 +1,6 @@
 #' Linear regression via lm
 #'
-#' `lm()` uses ordinary least squares to fit models with numeric outcomes.
+#' [stats::lm()] uses ordinary least squares to fit models with numeric outcomes.
```
juliasilge, line 3: It would be nice to have these as clickable links in the documentation. For `lm()` this is easy, because it is always installed. For the majority of models, we do _not_ have them in Depends or Suggests so we can't guarantee that they will be available to folks, in which case they show up as non-clickable links but code. Once they try to use the model, they will get a warning to install and then would be able to click the link.

For `R CMD check` and CRAN, I believe it will be fine to include these as links, as long as the package itself is on CRAN. Hadley recently checked this [with a toy package](https://github.com/hadley/testlink). 

- When [it is checked in an environment that does not have the linked package](https://github.com/hadley/testlink/runs/2230241772?check_suite_focus=true#step:10:65), there is a note:
> Package unavailable to check Rd xrefs: ‘ggplot2’
- When [it is checked on CRAN (where packages are available)](https://win-builder.r-project.org/h7vUaNAZQrJK/00check.log) there is no note.

If one of these packages gets booted off CRAN, we will likely get a quick "resubmit immediately" type email, but we would need to do that because of the model itself anyway.

man/rmd/linear-reg-glmnet.Rmd
```diff
@@ -0,0 +1,70 @@
...
+  set_engine("glmnet") %>% 
+  set_mode("regression") %>% 
+  translate()
+```
+
+## Preprocessing requirements
+
+Factor/categorical predictors need to be converted to numeric values (e.g., dummy or indicators variables). When using the formula method via [fit.model_spec()], `parsnip` will convert these data when required.
+
+Also, predictors should have the same scale. One way to achieve this is to center and scale each so that each predictors have mean zero and a variance of one.
```
juliasilge, line 51: What do you all think is the best way to talk about this for glmnet? The default for the underlying function is `standardize = TRUE` and we don't turn that off so it seems misleading to me to tell people to make sure to center/scale.

man/rmd/linear-reg-glmnet.Rmd
```diff
@@ -0,0 +1,70 @@
...
+
+## Other details
```
juliasilge, line 53: I like this "Other details" section a lot; after chatting with Max, I moved some of the slightly less important details down so that the sections like "Tuning parameters" and such can be more consistent from engine to engine.

Some of the other engines currently talk about case weights, and let's just consider that a placeholder; we should remove those before merging. Eventually we can put them back in but not until we support them.

man/rmd/linear-reg-glmnet.Rmd
juliasilge: Actually, `sparklyr::ml_linear_regression()` says it has `standardization = TRUE` too

hfrick (commented):
Is it possible to get to an engine help page directly using its name? You might want to do that when you have already decided on the engine and don't want to click through from `?linear_reg`. I tried `?details-linear-reg-glmnet` but it didn't like the `-`. Could we use `_` or does that break something?

R/linear-reg-doc-glmnet.R
topepo: Do you mean the parts in the `learn_ref_data.R` file? I like that idea.

man/rmd/glmnet-details.Rmd
```diff
@@ -0,0 +1,117 @@
+```{r, child = "setup.Rmd", include = FALSE}
+```
+
+# About the glmnet model
+
+First, `glmnet` fits two types of penalties: the lasso (or L1) penalty and the ridge penalty (a.k.a L2 penalty a.k.a. weight decay). The `glmnet` argument `alpha` describes the proportion of the lasso penalty in the overall penalty. 
+
+When `glmnet` fits the model, the intercept coefficient is not penalized. 
+
+One aspect of the model is that there is a _coefficient path_ for each data set and a given value of `alpha`. When the `glmnet` model is trained, it estimates a continuum of possible penalty values. The user doesn't need to declare the exact value of the penalty that should be used until sample prediction (i.e., it is not needed when the model is trained). If you specify a single value of `lamda` when calling `glmnet()`, the estimated coefficient path only has that value.
```
hfrick, line 10:
```suggestion
When the `glmnet` model is trained, it estimates a continuum of possible penalty values. This gives a _coefficient path_ for each data set and a given value of `alpha`. The user doesn't need to declare the exact value of the penalty that should be used until sample prediction (i.e., it is not needed when the model is trained). If you specify a single value of `lamda` when calling `glmnet()`, the estimated coefficient path only has that value. 
```

hfrick (approved):

topepo:
This now requires devel `dials`

man/rmd/glmnet-details.Rmd
```diff
@@ -0,0 +1,117 @@
+```{r, child = "setup.Rmd", include = FALSE}
+```
+
+# About the glmnet model
+
+First, `glmnet` fits two types of penalties: the lasso (or L1) penalty and the ridge penalty (a.k.a L2 penalty a.k.a. weight decay). The `glmnet` argument `alpha` describes the proportion of the lasso penalty in the overall penalty. 
+
+When `glmnet` fits the model, the intercept coefficient is not penalized. 
+
+When the `glmnet` model is trained, it estimates a continuum of possible penalty values. This gives a _coefficient path_ for each data set and a given value of `alpha`. The user doesn't need to declare the exact value of the penalty that should be used until sample prediction (i.e., it is not needed when the model is trained). If you specify a single value of `lamda` when calling [glmnet::glmnet()], the estimated coefficient path only has that value.
```
hfrick, line 10:
```suggestion
When the `glmnet` model is trained, it estimates a continuum of possible penalty values. This gives a _coefficient path_ for each data set and a given value of `alpha`. The user doesn't need to declare the exact value of the penalty that should be used until sample prediction (i.e., it is not needed when the model is trained). If you specify a single value of `lambda` when calling [glmnet::glmnet()], the estimated coefficient path only has that value. 
```

man/rmd/glmnet-details.Rmd
```diff
@@ -0,0 +1,117 @@
...
+
+When the `glmnet` model is trained, it estimates a continuum of possible penalty values. This gives a _coefficient path_ for each data set and a given value of `alpha`. The user doesn't need to declare the exact value of the penalty that should be used until sample prediction (i.e., it is not needed when the model is trained). If you specify a single value of `lamda` when calling [glmnet::glmnet()], the estimated coefficient path only has that value. 
+
+# Using glmnet with parsnip
+
+In tidymodels, we want: 
+
+1. The user to declare exactly what penalty value to use.
+
+1. The ability to make predictions at multiple penalty with a single model object.
```
hfrick, line 18:
```suggestion
1. The ability to make predictions at multiple penalty values with a single model object. 
```

R/boost_tree.R
```diff
@@ -1,33 +1,21 @@
...
+#' Boosted trees
+#'
+#' @description
+#'
+#' `boost_tree()` defines a model that creates a series of decision trees that form an
+#' ensemble. Each tree depends on the results of previous trees. All trees in
+#' the ensemble are combined into a final prediction.
+#'
+#' There are different ways to fit this model. Information about the available
+#' _engines_ that can be used for fitting at:
```
hfrick, line 12:
```suggestion
#' There are different ways to fit this model. See the engine-specific pages for more details:
```
Leaving this here as an example and not spamming all pages with the same suggested change

R/boost_tree.R
```diff
@@ -1,33 +1,21 @@
...
+#'
+#' \Sexpr[stage=render,results=rd]{parsnip:::find_engine_files("boost_tree")}
+#'
+#' More information on how `parsnip` is used for model is at
```
hfrick, line 16:
```suggestion
#' More information on how `parsnip` is used for modeling is at
```

hfrick: that's also a change that would need to go in all of the pages

man/rmd/boost_tree_xgboost.Rmd
```diff
@@ -0,0 +1,98 @@
...
+  stop_iter = integer()
+) %>% 
+  set_engine("xgboost") %>% 
+  set_mode("classification") %>% 
+  translate()
+```
+
+[xgb_train()] is a wrapper around [xgboost::xgb.train()] (and other functions) that makes it easier to run this model. 
+
+## Preprocessing requirements
```
hfrick, line 68: This section only covers predictors but we also convert the outcome if it's a factor.

https://github.com/tidymodels/parsnip/blob/2a8da60d605bd63d5d1a7c0dd320921900eafac6/man/rmd/boost-tree.Rmd#L45

R/linear_reg.R
```diff
@@ -1,71 +1,49 @@
...
+#' predictors using a linear function.
+#'
+#' There are different ways to fit this model. Information about the available
+#' _engines_ that can be used for fitting at:
+#'
+#' \Sexpr[stage=render,results=rd]{parsnip:::find_engine_files("linear_reg")}
+#'
+#' Loading model `parsnip`-adjacent packages will show additional engines for
+#' this model. The `multilevelmod` package contains additional engines for
+#' `linear_reg()`.
```
hfrick, line 15:
```suggestion
#' Loading `parsnip`-adjacent packages which contain more engines for 
#' this model will make the additional engines show up here. For example,
#' the [multilevelmod](https://github.com/tidymodels/multilevelmod/) package 
#' contains additional engines for `linear_reg()`.
```
Do we want to point to multilevelmod when it's not on CRAN yet?

man/rmd/glmnet-details.Rmd
```diff
@@ -0,0 +1,117 @@
...
+
+Point 1 is relevant for a single `parsnip` model while Point 2 is more related to using the `tune` package. The second point greatly increases the computation efficiency of model tuning since you only need to fit a `glmnet` model for each value of alpha (see the [tidymodels book](https://www.tmwr.org/grid-search.html#submodel-trick) for a discussion).
+
+To achieve both of these goals, `parsnip` does the following: 
+
+* The model specification (e.g. `linear_reg()`) saves the value of the penalty that the user specifies.
+
+* `parsnip` fits the `glmnet` model without passing the penalty value to the `lambda` argument. This makes the whole coefficient path available for prediction. 
+
+* When `predict()` is used on the `parsnip` model fit, the saved value of the penalty is used.
```
hfrick, line 28:
```suggestion
* When `predict()` is used on the `parsnip` model fit without specifying the `penalty` value, the saved value of the penalty is used. 
```

hfrick: because you can use a different penalty value, as long as it's a single value

man/rmd/glmnet-details.Rmd
```diff
@@ -0,0 +1,117 @@
...
+
+To achieve both of these goals, `parsnip` does the following: 
+
+* The model specification (e.g. `linear_reg()`) saves the value of the penalty that the user specifies.
+
+* `parsnip` fits the `glmnet` model without passing the penalty value to the `lambda` argument. This makes the whole coefficient path available for prediction. 
+
+* When `predict()` is used on the `parsnip` model fit, the saved value of the penalty is used. 
+
+* To make predictions at other values along the path, use the `multi_predict()` function. Multiple predictions can be made simultaneously at different penalty values.
```
hfrick, line 30:
```suggestion
* To make predictions at multiple values along the path, use the `multi_predict()` function. Multiple predictions can be made simultaneously at different penalty values. 
```

man/rmd/glmnet-details.Rmd
```diff
@@ -0,0 +1,117 @@
...
+
+If tune is being used to find optimal values of the penalty, it uses the `parsnip::multi_predict()` function to get predictions over multiple penalty values: 
+
+```{r}
+multi_predict(glmnet_fit, head(Chicago), penalty = c(0, 0.1)) %>% 
+  mutate(row = row_number()) %>% 
+  unnest(cols = c(.pred))
+```
+
+You can modify the penalties contained in the path by using the [glmnet::glmnet()] arguments `nlambda` and `lambda.min.ratio`. The latter is discussed more below. These arguments would be passed when `set_engine()` is called.
```
hfrick, line 82:
```suggestion
You can modify the penalties contained in the path by using the [glmnet::glmnet()] arguments `nlambda` and `lambda.min.ratio`. The latter is discussed more below. These arguments can be specified in `set_engine()`. 
```

R/linear_reg.R
topepo: I'll take that sentence out until it is ready.

man/rmd/glmnet-details.Rmd
topepo: I'll edit this and use the new path argument instead.

R/linear-reg-doc-glmnet.R
topepo: Ok. I agree with this but we have to figure a few things out. Those `_data` files have an engine-independent bit at the top. For example: 

```r
set_new_model("boost_tree")
set_model_mode("boost_tree", "classification")
set_model_mode("boost_tree", "regression")
```

These would have to be executed before the engine-specific files. Maybe put all of them in `aaa_models.R` or a new file?

topepo:
a0a683b added an overview of how this all works. Maybe @hfrick can test this later when we do it for `censored`.

NEWS.md
```diff
@@ -24,7 +24,11 @@ r
 
 * `set_mode()` now checks if `mode` is compatible with the model class, similar to `new_model_spec()` (@jtlandis, #467). Both `set_mode()` and `set_engine()` now error for `NULL` or missing arguments (#503).
 
-* Re-organized model documentation for `update` methods (#479).
+* Re-organized model documentation:
+
+   * `update` methods were moved out of the model help files (#479).
+   * Each model/engine combination has it's own help page.
```
hfrick, line 30:
```suggestion
   * Each model/engine combination has its own help page. 
```

man/rmd/template-reg-chicago.Rmd
```diff
@@ -0,0 +1,18 @@
...
+
+```{r, results = "hide"}
+library(tidymodels)
+tidymodels_prefer()
+data(Chicago)
+
+n <- nrow(Chicago)
+Chicago <- Chicago %>% select(ridership, Clark_Lake, Quincy_Wells)
+
+Chicago_train <- Chicago[1:(n - 6), ]
```
hfrick, line 15: 1:(n-7) so that (n-6) isn't in train and test

topepo: Yes 🙄

hfrick:
```suggestion
Chicago_train <- Chicago[1:(n - 7), ]
```

man/rmd/linear_reg_stan.Rmd
```diff
@@ -0,0 +1,44 @@
...
+
+## Important engine-specific options
+
+Some relevant arguments that can be passed to `set_engine()`: 
+
+ * `chains`: A positive integer specifying the number of Markov chains. The default is 4.
+ * `iter`: A positive integer specifying the number of iterations for each chain (including warmup). The default is 2000.
+ * `seed`: The seed for random number generation. 
+ * `cores`: Number of cores to use when executing the chains in parallel.
+ * `prior`: The prior distribution for the (non-hierarchical) regression coefficients. The `stan` engine does not fit any hierarchical terms. See the `stan_glmer` engine from the `multilevelmod` package for that type of model.
```
juliasilge, line 18: Do we want to mention this package that is not on CRAN yet?

man/rmd/multinom_reg_glmnet.Rmd
```diff
@@ -0,0 +1,93 @@
...
+mr_cls_fit
+```
+
+The holdout data can be predicted for both hard class predictions and probabilities. We'll bind these together into one tibble:
+
+```{r}
+predict(mr_cls_fit, penguins_test, type = "prob") %>% 
+  bind_cols(
+    predict(mr_cls_fit, penguins_test)
+  )
```
juliasilge, line 85: Style-wise, I'd put these both inside, like: `bind_cols(predict(...), predict(...))`

juliasilge: Same for other examples

man/rmd/multinom_reg_glmnet.Rmd
```diff
@@ -0,0 +1,93 @@
...
+## Preprocessing requirements
+
+```{r child = "template-makes-dummies.Rmd"}
+```
+
+```{r child = "template-same-scale.Rmd"}
+```
+By default, [glmnet::glmnet()] uses the argument `standardize = TRUE` to center and scale the data. 
+
+## Working example
```
juliasilge, line 57:
```suggestion
## Example
```

This seems like a bit of a funny heading to me, like maybe we will have some non-working examples too??? 😆 Same change throughout

man/rmd/rand_forest_ranger.Rmd
```diff
@@ -0,0 +1,145 @@
...
+Note that a `ranger` probability forest is always fit (unless the `probability` argument is changed by the user via [set_engine()]).
+
+## Preprocessing requirements
+
+```{r child = "template-split-factors.Rmd"}
+```
+
+## Other notes
+
+By default, parallel processing is turned off. To run each model in parallel, change the `num.threads` argument via [set_engine()]. When tuning, it is more efficient to parallelize over the resamples and tuning parameters.
```
juliasilge, line 77:
```suggestion
By default, parallel processing is turned off. To run each model in parallel, change the `num.threads` argument via [set_engine()]. When tuning, it is more efficient to parallelize over the resamples and tuning parameters.
```
This doesn't seem quite right to me. Should it be: "To run each tree in parallel"?

man/rmd/template-split-factors.Rmd
```diff
@@ -0,0 +1 @@
+This engine does not require any special encoding of the predictors. Categorical predictors can be split into groups without creating dummy variables for each category.
```
juliasilge, line 1: I find "split into groups" kind of confusing, maybe because I haven't heard about these (mostly tree) models talked about this way. I think it would take more explanation to really get into the groups we mean (groups of factor levels within one factor splitting at a node? groups across features?) so I think it might be better to be more general.

juliasilge:
What a set of changes 😅  but this is looking really great! 🙌 

Other than the questions I have open above, I think the only other thing I want to check on in that we don't have these changes implemented for the survival models. That is as designed, correct?

man/rmd/multinom_reg_glmnet.Rmd
hfrick: I've also switched the order around, first predicting the class (the default)  then the class probabilities. that also matches up better with the sentence right before it.

juliasilge: That looks nice and makes sense!

man/rmd/linear_reg_stan.Rmd
juliasilge: Let's remove this

man/rmd/template-split-factors.Rmd
topepo: How about:

> This engine does not require any special encoding of the predictors. Categorical predictors can be partitioned into groups of factor levels (e.g. `{a, c, d}` vs `{b}`) when splitting occurs. Dummy variables are not required for this model.

Issue #467, opened 2021-04-08
Additional checks in `set_mode` prevent improper use of model specs


jtlandis:
## Feature

I want to preface this by saying I'm unsure if this is technically a bug or a feature request. 

I think `parsnip` could benefit from adding an additional check in `set_mode` to ensure that the user does not add an improper mode to a model Specification. For example, when initializing a `linear_reg` specification, it will check if `mode` is either `c("unknown", "regression")`:

``` r
library(tidymodels)

# correct specification
mod1 <- linear_reg() %>%
  set_engine('lm') 

mod1
#> Linear Regression Model Specification (regression)
#> 
#> Computational engine: lm

#returns an error - lienar_reg cannot be used for classification
mod_error <- linear_reg('classification') %>%
  set_engine('lm')
#> Error: `mode` should be one of: 'unknown', 'regression'
```

However, the user can still do `set_mode("classification")` without any warnings or errors.

```r
mod2 <- mod1 %>%
  set_mode('classification') #should this be allowed?

mod2
#> Linear Regression Model Specification (classification)
#> 
#> Computational engine: lm

```

As a result, the user gets a really confusing downstream error when trying to use that "classification" model specification:

```r

mod2 %>%
  fit(Species ~ ., data = iris)
#> Error: formula_ is unknown.

```
To someone who has just discovered `tidymodels` and has followed many online tutorials in which setting your mode with `set_mode()` seems to be a common API practice, they would be under the impression that `mod2` is a perfectly acceptable model spec. 

I would imagine it isn't intended for users to add an incompatible mode to a model specification. So, adding something like the following function to validate proper spec modes might be desirable:

```r
check_valid_spec_mode <- function(model, mode) {
    spec_modes <- rlang::env_get(get_model_env(), paste0(model, "_modes"))
        if (!(mode %in% spec_modes)) 
            rlang::abort(glue::glue("`mode` should be one of: ", 
                glue::glue_collapse(glue::glue("'{spec_modes}'"), 
                    sep = ", ")))
    invisible(TRUE)
}
```
Incase the reader was curious, I found the source of the downstream error. The"`formula_`" comes from a `switch` statement in `fit.model_spec`, which is attributed to no `fit$interface` existing on `object$method`. For example, `add_method` is called, which then calls `get_model_spec`, which is then unable to find a record in `linear_reg_fit` of a mode equivalent to "classification" and an engine of "lm".

[...]

PR #470, opened 2021-04-14
Check valid spec mode

jtlandis:
Hello, 

This pull request is to fix the small bug found in #467 . This pull request will prevent users from adding an incompatible `mode` to a `model_spec` object via `set_mode`.

The implementation was simple:

1. refactor section of `new_model_spec` that checks class and mode compatibility into a separate function `check_spec_mode_val`
2. replace code in `new_model_spec` with new function
3. add new function to `set_mode`

I also added about 47 additional tests in the file `tests/testthat/test_spec_and_modes.R`. Some of these may be redundant as they simply check that compatible `mode`s can be added to a given `model_spec`. The novel part of the test that script includes `expect_error` for when adding an improper `mode` type.


NEWS.md
```diff
@@ -12,6 +12,8 @@
 
 * Re-licensed package from GPL-2 to MIT. See [consent from copyright holders here](https://github.com/tidymodels/parsnip/issues/462).
 
+* `set_mode` now checks if mode is compatible with model class, similar to `new_model_spec` (jtlandis #467)
```
DavisVaughan, line 15:
```suggestion
* `set_mode()` now checks if `mode` is compatible with the model class, similar to `new_model_spec()` (@jtlandis, #467).
```

R/arguments.R
```diff
@@ -87,6 +87,10 @@ set_mode <- function(object, mode) {
       )
     )
   }
+  #only check if object is a model_spec
+  if(inherits(object, "model_spec")) {
+    check_spec_mode_val(class(object)[1], mode)
+  }
```
DavisVaughan, line 93:
```suggestion
  check_spec_mode_val(class(object)[1], mode)
```

I think you should be able to do this check unconditionally. `object` should always be a model-spec

tests/testthat/test_spec_and_modes.R
```diff
@@ -0,0 +1,165 @@
+library(testthat)
```
DavisVaughan, line 1: Rather than having a new file, I would put the test in `test_args_and_modes.R`

I also think you really only need 1 new test here. What we really want to test is the fact that `set_mode()` is now checking for this behavior. The specific model spec doesn't matter quite as much, testing 1 is sufficient since all of them go through `set_mode()`.

So I would add a test to the end of `test_args_and_modes.R` that looks like:

```r
test_that("can't set a mode that isn't allowed by the model spec", {
  expect_error(
    set_mode(linear_reg(), "classification"),
    "`mode` should be one of"
  )
})
```

and then delete this test file. I appreciate the diligence though!

jtlandis:
I will make these changes and redo the PR 👍

juliasilge (approved):
Thank you so much @jtlandis! 🙌

PR #479, opened 2021-04-26
Move all models' `update` docs

### LINKED ISSUE #423: Removing `update.*()` from Github Pages
Right now the `update.*()` functions are listed alongside their respective model specification function on the [reference page](https://parsnip.tidymodels.org/reference/index.html) on Github Pages. I feel that they add more visual noise than the utility. Don't know if this change is desirable or doable.

# Current version
Screen Shot 2021-02-01 at 1 39 43 PM

# Suggested version
Screen Shot 2021-02-01 at 1 40 30 PM

juliasilge:
Closes #423 

This PR moves the documentation for `update` methods from the individual model pages to a centralized `update` page. We want to do this _before_ we make any more headway on the big documentation change for models outlined in #456.


R/update.R
```diff
@@ -0,0 +1,51 @@
...
+#' @inheritParams proportional_hazards
+#' @inheritParams rand_forest
+#' @inheritParams surv_reg
+#' @inheritParams svm_linear
+#' @inheritParams svm_poly
+#' @inheritParams svm_rbf
+#' @param object A model specification.
+#' @param parameters A 1-row tibble or named list with _main_
+#'  parameters to update. If the individual arguments are used,
+#'  these will supersede the values in `parameters`. Also, using
```
DavisVaughan, line 24: > If the individual arguments are used, these will supersede the values in `parameters`.

I'm not entirely sure what this means. Could you expand on what "individual arguments" refers to?

R/update.R
```diff
@@ -0,0 +1,51 @@
...
+#'  engine arguments in this object will result in an error.
+#' @param ... Not used for `update()`.
+#' @param fresh A logical for whether the arguments should be
+#'  modified in-place of or replaced wholesale.
```
DavisVaughan, line 28:
```suggestion
#' @param fresh A logical for whether the arguments should be
#'  modified in-place or replaced wholesale.
```

+ Redocument

DavisVaughan (approved):

R/update.R
juliasilge: Do you think this is more clear?

> A 1-row tibble or named list with _main_  parameters to update. Use **either** `parameters` **or** the main arguments directly when updating. If the main arguments are used, these will supersede the values in `parameters`. Also, using engine arguments in this object will result in an error.

DavisVaughan: Ah that is very clear yes, thank you!

PR #488, opened 2021-05-12
Improve errors for `set_mode()` and `set_engine()`

### LINKED ISSUE #426: More informative error when `engine` argument is missing in set_engine()
Currently, the error message is telling you that `engine` should be specified, but not what values it can take.

``` r
library(parsnip)

linear_reg() %>%
  set_engine()
#> Error in set_engine(.): argument "engine" is missing, with no default
```

<sup>Created on 2021-02-06 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

It wouldn't take too much (famous last words) to have the error changed to something like

``` r
#> Error in set_engine(.): argument "engine" is missing, Please use one of: 'lm', 'glmnet', 'stan', 'spark', 'keras'
```

If you like the idea I can whip up a PR

juliasilge:
Closes #426

This PR makes our handling of a _missing_ argument and a _bad_ argument a bit more consistent for `set_mode()` and `set_engine()`. There was some handling coded up in both cases but we weren't really getting those benefits for missing (or even `NULL`) inputs before. This rearranging should make help things out.

The already existing warning/replacement for engines now works:
``` r
library(parsnip)
linear_reg() %>% set_engine()
#> Warning: `engine` was NULL and updated to be `lm`
#> Linear Regression Model Specification (regression)
#> 
#> Computational engine: lm

## same as before
linear_reg() %>% set_engine("ranger")
#> Error: Engine 'ranger' is not available. Please use one of: 'lm', 'glmnet', 'stan', 'spark', 'keras'
```

<sup>Created on 2021-05-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

For mode, I just set this up to tell the user their options:

``` r
library(parsnip)
rand_forest() %>% set_mode()
#> Error: `mode` should be one of 'classification', 'regression', 'censored regression'
rand_forest() %>% set_mode(NULL)
#> Error: `mode` should be one of 'classification', 'regression', 'censored regression'
rand_forest() %>% set_mode("ranger")
#> Error: `mode` should be one of 'classification', 'regression', 'censored regression'
```

<sup>Created on 2021-05-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

I did need to adjust some tests, switching between errors and warnings, but I don't think any of that is substantive.

### DIFF
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -135,13 +135,15 @@ check_mode_val <- function(mode) {
 # check if class and mode are compatible
 check_spec_mode_val <- function(cls, mode) {
   spec_modes <- rlang::env_get(get_model_env(), paste0(cls, "_modes"))
-  if (!(mode %in% spec_modes))
-    rlang::abort(
-      glue::glue(
-        "`mode` should be one of: ",
-        glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
-      )
+  compatible_modes <-
+    glue::glue(
+      "`mode` should be one of: ",
+      glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
     )
+
+  if (is.null(mode)) rlang::abort(compatible_modes)
+  else if (!(mode %in% spec_modes)) rlang::abort(compatible_modes)
+
   invisible(NULL)
 }
 
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -127,8 +127,9 @@ check_model_doesnt_exist <- function(model) {
 }
 
 check_mode_val <- function(mode) {
-  if (rlang::is_missing(mode) || length(mode) != 1 || !is.character(mode))
+  if (rlang::is_missing(mode) || length(mode) != 1 || !is.character(mode)) {
     rlang::abort("Please supply a character string for a mode (e.g. `'regression'`).")
+  }
   invisible(NULL)
 }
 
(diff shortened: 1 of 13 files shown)


juliasilge:
For both `set_mode()` and `set_engine()`, we could set a default value of `NULL` for the mode/engine and remove those lines like:

```r
if (rlang::is_missing(engine)) engine <- NULL
```

Do we want to have a default value show up in the docs? It seems a little weird when a value of `NULL` just triggers these warnings / errors.

topepo:
There might be a number of places in the code and docs to remove traces of "unknown" mode. TBH I'm fine with leaving that possible mode but not referencing it in warnings or errors that happen during checks about model fitting.

juliasilge:
I looked at this again today and changed my mind about the `set_mode()` approach. Now it all goes through `check_spec_mode_val()` and there is no more returning the generic `parsnip:::all_modes` in some situations:

``` r
library(parsnip)
logistic_reg() %>% set_mode()
#> Error: `mode` should be one of: 'unknown', 'classification'
logistic_reg() %>% set_mode("regression")
#> Error: `mode` should be one of: 'unknown', 'classification'

rand_forest() %>% set_mode()
#> Error: `mode` should be one of: 'unknown', 'classification', 'regression'
rand_forest() %>% set_mode(NULL)
#> Error: `mode` should be one of: 'unknown', 'classification', 'regression'
rand_forest() %>% set_mode("ranger")
#> Error: `mode` should be one of: 'unknown', 'classification', 'regression'
```

<sup>Created on 2021-05-12 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

I tend to think it is OK to leave "unknown" for now and we can make a call about that in the future.

topepo (approved):

PR #497, opened 2021-05-18
temporarily mark survival models as internal

topepo:
Until censored is fully operational

### DIFF
--- a/R/proportional_hazards.R
+++ b/R/proportional_hazards.R
@@ -27,7 +27,7 @@
 #' @seealso [fit()], [set_engine()], [update()]
 #' @examples
 #' show_engines("proportional_hazards")
-#'
+#' @keywords internal
 #' @export
 proportional_hazards <- function(mode = "censored regression",
                     penalty = NULL,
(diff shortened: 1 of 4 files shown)


hfrick:
We soft-deprecated `surv_reg()`, i.e. it now gives a warning and tells you to use `survival_reg()`, the help file including examples points to the new function, and the help file itself is now also internal. So I'd say we either undo all of that or we don't make `survival_reg()` internal.

Are we also adapting the news file?

juliasilge:
I do think we should take this out of the NEWS file and put it back in after it goes to CRAN, for the next release.

topepo:
I've update the news file in the RC branch. 

I think the message is ok. It's not removed, just not listed (easily) in the help.

hfrick:
okay sure, it just means that in this branch neither `surv_reg()` nor `survival_reg()` are listed in the documentation index. if you think that's fine then I have no more comments on this one :+1:

hfrick (approved):

PR #499, opened 2021-05-19
xgboost mtry parameter swap for #495

### LINKED ISSUE #461: when `mtry =1` all predictors are used in xgboost engine
Whenever `mtry = 1` in the model spec the corresponding value of `colsample_bytree` is not converted to proportion correctly. 
example: when `mtry = 1` , `colsample_bytree = 1` when it should be 0.125.

``` r
library(tidyverse)
library(tidymodels)
library(mlbench)
library(xgboost)
#> 
#> Attaching package: 'xgboost'
#> The following object is masked from 'package:dplyr':
#> 
#>     slice
library(reprex)


data("PimaIndiansDiabetes")
df <- PimaIndiansDiabetes %>%
  mutate(diabetes = fct_relevel(diabetes, 'pos'))

#Model with mtry = 1, col_sample_by_tree should be 1/8 = 0.125

tidy_model <- 
  boost_tree(trees = 10,
             mtry = 1,
             tree_depth = 3) %>%
  set_engine('xgboost', 
             eval_metric = 'auc',
             event_level = "first",
             verbose = 1) %>%
  set_mode('classification')

set.seed(24)

tidy_model_fitted <- tidy_model %>% 
  fit(diabetes ~ . , data = df)
#> [1]  training-auc:0.834787 
#> [2]  training-auc:0.853881 
#> [3]  training-auc:0.873048 
#> [4]  training-auc:0.875000 
#> [5]  training-auc:0.882873 
#> [6]  training-auc:0.891716 
#> [7]  training-auc:0.896067 
#> [8]  training-auc:0.900414 
#> [9]  training-auc:0.904313 
#> [10] training-auc:0.906937

#Model with mtry  = 2, col_sample_by_tree should by 2/8 = 0.25
tidy_model_2 <- 
  boost_tree(trees = 10,
                         mtry = 2,
                         tree_depth = 3) %>%
  set_engine('xgboost', 
             eval_metric = 'auc',
             event_level = "first",
             verbose = 1) %>%
  set_mode('classification')

set.seed(24)
tidy_model_fitted_2 <- tidy_model_2 %>% 
  fit(diabetes ~ . , data = df)
#> [1]  training-auc:0.684049 
#> [2]  training-auc:0.776993 
#> [3]  training-auc:0.802187 
#> [4]  training-auc:0.808041 
#> [5]  training-auc:0.830478 
#> [6]  training-auc:0.837851 
#> [7]  training-auc:0.847817 
#> [8]  training-auc:0.847254 
#> [9]  training-auc:0.850034 
#> [10] training-auc:0.856455

#col_sample by tree should be 1/8 = 0.125, instead model is using all predictors
tidy_model_fitted$fit$call$params$colsample_bytree
#> [1] 1

#colsample_bytree is correct 2/8 = 0.25
tidy_model_fitted_2$fit$call$params$colsample_bytree
#> [1] 0.25
```

<sup>Created on 2021-04-06 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

### LINKED ISSUE #495: mtry maps to wrong parameter for XGBoost
From `?boost_tree`, we have 

>`mtry`: The number of predictors that will be randomly sampled at each split when creating the tree models.

In the [model argument translation](https://github.com/tidymodels/parsnip/blob/228a6dc6975fc91562b63d191e43d2164cc78e3d/R/boost_tree_data.R#L36) for XGBoost, we have `mtry` maps to  `colsample_bytree`.

However, the [XGBoost documentation](https://xgboost.readthedocs.io/en/latest/parameter.html) says
>colsample_bytree is the subsample ratio of columns when constructing each tree. Subsampling occurs once for every tree constructed.

I don't read this and `mtry` as meaning the same thing. In fact, `mtry` sounds more like this XGBoost documentation excerpt:
>colsample_bynode is the subsample ratio of columns for each node (split). Subsampling occurs once every time a new split is evaluated. Columns are subsampled from the set of columns chosen for the current level.

topepo:
closes #495
closes #461

`colsample_bytree` remains an argument to `xgb_train()`. The `mtry` parameter in `boost_tree()` parameter now points to `colsample_bynode`. 

We might want to add more engine-specific tunables to `tune` for this engine. 

@mdancho84 I think that `modeltime` would need the same switch. 

(edit for clarity)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,9 +11,7 @@
   
 * The `liquidSVM` engine for `svm_rbf()` was deprecated due to that package's removal from CRAN. (#425)
 
-* New model specification `survival_reg()` for the new mode `"censored regression"` (#444). `surv_reg()` is now soft-deprecated (#448).
-
-* New model specification `proportional_hazards()` for the `"censored regression"` mode (#451).
+* The xgboost engine for boosted trees was translating `mtry` to xgboost's `colsample_bytree`. We now map `mtry` to `colsample_bynode` since that is more consistent with how random forest works. `colsample_bytree` can still be optimized by passing it in as an engine argument. (#495)
 
 ## Other Changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,7 +11,9 @@
   
 * The `liquidSVM` engine for `svm_rbf()` was deprecated due to that package's removal from CRAN. (#425)
 
-* The xgboost engine for boosted trees was translating `mtry` to xgboost's `colsample_bytree`. We now map `mtry` to `colsample_bynode` since that is more consistent with how random forest works. `colsample_bytree` can still be optimized by passing it in as an engine argument. (#495)
+* The xgboost engine for boosted trees was translating `mtry` to xgboost's `colsample_bytree`. We now map `mtry` to `colsample_bynode` since that is more consistent with how random forest works. `colsample_bytree` can still be optimized by passing it in as an engine argument. `colsample_bynode` was added to xgboost after the `parsnip` package code was written. (#495)
+
+* For xgboost boosting, `mtry` and `colsample_bytree` can be passed as integer counts or proportions while `subsample` and `validation` should be proportions. `xgb_train()` now has a new option `counts` for state what scale `mtry` and `colsample_bytree` are being used. (#461)  
 
 ## Other Changes
 
@@ -21,11 +23,7 @@
 
 * Re-organized model documentation for `update` methods (#479).
 
- 
-  
 * `generics::required_pkgs()` was extended for `parsnip` objects. 
-
-
 
 # parsnip 0.1.5
 
--- a/R/boost_tree.R
+++ b/R/boost_tree.R
@@ -264,7 +264,9 @@ check_args.boost_tree <- function(object) {
 #' @param max_depth An integer for the maximum depth of the tree.
 #' @param nrounds An integer for the number of boosting iterations.
 #' @param eta A numeric value between zero and one to control the learning rate.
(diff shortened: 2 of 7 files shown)


juliasilge:
So now both of these parameters are going to have the problem outlined in #461

juliasilge:
I reran some analyses with xgboost, including [this one](https://juliasilge.com/blog/xgboost-tune-volleyball/), and the results are different in about the ways I think we'd expect:

new_mtry

No big changes in overall performance after tuning, but it switched some things in the variable importance and such. People who come back to train with the same data after this change are going to notice.

juliasilge (approved):

mdancho84:
I'm here now - just seeing this. Yep, Modeltime will need the switch (I believe).

mdancho84:
Hey, [I've reviewed](https://github.com/business-science/modeltime.ensemble/issues/10#issuecomment-848002142) and I see one potential issue with backwards compatibility. What happens is that models that may have been specified with a value of 1 thinking this means 100%, gets converted to one column, which makes model performance very bad. 

I recommend handling 1 as 100% of columns, and not 1 as 1 column.

topepo:
The previous behavior of "1.00 mean s 100% but otherwise it is a count" was a big mistake on my part. 

It is 100% is consistent with `xgboost`, but completely inconsistent with `randomForest`, `ranger`, `gbm`, `C50` and so on. We want to avoid people having to think "but for xgboost I have to do _this_ instead of _that_". There are probably 4-5 things in the overall `xgboost` api that are irregular (and we fix them).

mdancho84:
We are good to go. Let me know when `parsnip` is accepted by CRAN, and I will update `modeltime`.

PR #503, opened 2021-05-21
Harmonize errors for `set_mode()` and `set_engine()`

### LINKED ISSUE #494: `set_engine(object, engine)` should be `set_engine(object, engine = NULL)`
Since if `engine = NULL` we choose a default for the user with a message
https://github.com/tidymodels/parsnip/pull/488/files

It is better if this is explicit in the function signature, and documented in the help file.

Possibly also `set_mode(object, mode = NULL)`?

juliasilge:
Closes #494 

This PR harmonizes the errors (previously a mix of warnings + errors) for `set_mode()` and `set_engine()`. I made little functions for the error messages and removed the functionality to choose a default engine, as we discussed. Folks will now get results like this for the engine:

``` r
library(parsnip)
linear_reg() %>% set_engine()
#> Error: Available engines are: 'lm', 'glmnet', 'stan', 'spark', 'keras'
linear_reg() %>% set_engine(NULL)
#> Error: Available engines are: 'lm', 'glmnet', 'stan', 'spark', 'keras'
linear_reg() %>% set_engine(3)
#> Error: Available engines are: 'lm', 'glmnet', 'stan', 'spark', 'keras'
linear_reg() %>% set_engine("ranger")
#> Error: Available engines are: 'lm', 'glmnet', 'stan', 'spark', 'keras'
```

<sup>Created on 2021-05-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

And this for mode:

``` r
library(parsnip)
logistic_reg() %>% set_mode()
#> Error: Available modes are: 'unknown', 'classification'
logistic_reg() %>% set_mode(NULL)
#> Error: Available modes are: 'unknown', 'classification'
logistic_reg() %>% set_mode(42)
#> Error: Available modes are: 'unknown', 'classification'
logistic_reg() %>% set_mode("regression")
#> Error: Available modes are: 'unknown', 'classification'

rand_forest() %>% set_mode()
#> Error: Available modes are: 'unknown', 'classification', 'regression'
rand_forest() %>% set_mode(NULL)
#> Error: Available modes are: 'unknown', 'classification', 'regression'
rand_forest() %>% set_mode("ranger")
#> Error: Available modes are: 'unknown', 'classification', 'regression'
```

<sup>Created on 2021-05-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -134,20 +134,21 @@ check_mode_val <- function(mode) {
 }
 
 
-glue_compatible_modes <- function(cls, spec_modes) {
-  glue::glue(
+stop_incompatible_mode <- function(spec_modes) {
+  msg <- glue::glue(
     "Available modes are: ",
     glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
   )
+  rlang::abort(msg)
 }
 
 # check if class and mode are compatible
 check_spec_mode_val <- function(cls, mode) {
   spec_modes <- rlang::env_get(get_model_env(), paste0(cls, "_modes"))
   if (is.null(mode) || length(mode) > 1) {
-    rlang::abort(glue_compatible_modes(cls, spec_modes))
+    stop_incompatible_mode(spec_modes)
   } else if (!(mode %in% spec_modes)) {
-    rlang::abort(glue_compatible_modes(cls, spec_modes))
+    stop_incompatible_mode(spec_modes)
   }
   invisible(NULL)
 }
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -133,21 +133,22 @@ check_mode_val <- function(mode) {
   invisible(NULL)
 }
 
+
+glue_compatible_modes <- function(cls, spec_modes) {
+  glue::glue(
+    "Available modes are: ",
+    glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
+  )
+}
+
 # check if class and mode are compatible
 check_spec_mode_val <- function(cls, mode) {
   spec_modes <- rlang::env_get(get_model_env(), paste0(cls, "_modes"))
-  compatible_modes <-
-    glue::glue(
-      "`mode` should be one of: ",
-      glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
-    )
-
-  if (is.null(mode)) {
-    rlang::abort(compatible_modes)
+  if (is.null(mode) | length(mode) > 1) {
+    rlang::abort(glue_compatible_modes(cls, spec_modes))
   } else if (!(mode %in% spec_modes)) {
-    rlang::abort(compatible_modes)
+    rlang::abort(glue_compatible_modes(cls, spec_modes))
   }
-
   invisible(NULL)
 }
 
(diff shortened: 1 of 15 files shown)


R/aaa_models.R
```diff
@@ -133,21 +133,22 @@ check_mode_val <- function(mode) {
...
   spec_modes <- rlang::env_get(get_model_env(), paste0(cls, "_modes"))
-  compatible_modes <-
-    glue::glue(
-      "`mode` should be one of: ",
-      glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
-    )
-
-  if (is.null(mode)) {
-    rlang::abort(compatible_modes)
+  if (is.null(mode) | length(mode) > 1) {
```
DavisVaughan, line 147:
```suggestion
  if (is.null(mode) || length(mode) > 1) {
```

R/aaa_models.R
```diff
@@ -133,21 +133,22 @@ check_mode_val <- function(mode) {
   invisible(NULL)
 }
 
+
+glue_compatible_modes <- function(cls, spec_modes) {
+  glue::glue(
+    "Available modes are: ",
+    glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
+  )
+}
```
DavisVaughan, line 142: In vctrs when we have something like this, we normally push the `abort()` call in here too, and call it like:

```r
stop_incompatible_mode(cols, spec_modes)
```

Then you don't have to do `rlang::abort()` everywhere

R/arguments.R
```diff
@@ -76,11 +76,12 @@ set_args <- function(object, ...) {
 #' @rdname set_args
 #' @export
 set_mode <- function(object, mode) {
+  cls <- class(object)[1]
   if (rlang::is_missing(mode)) {
-    mode <- NULL
+    spec_modes <- rlang::env_get(get_model_env(), paste0(cls, "_modes"))
+    rlang::abort(glue_compatible_modes(cls, spec_modes))
```
DavisVaughan, line 82: It looks like you actually only need to pass `cls` through, then `stop_incompatible_modes()` could lookup `spec_modes`

R/engines.R
```diff
@@ -10,23 +10,20 @@ possible_engines <- function(object, ...) {
...
-  }
-  if (!(object$engine %in% avail_eng)) {
-    rlang::abort(
-      glue::glue(
-        "Engine '{object$engine}' is not available. Please use one of: ",
-        glue::glue_collapse(glue::glue("'{avail_eng}'"), sep = ", ")
-      )
-    )
+  eng <- object$engine
+  if (is.null(eng) | length(eng) > 1) {
```
DavisVaughan, line 23:
```suggestion
  if (is.null(eng) || length(eng) > 1) {
```

R/engines.R
```diff
@@ -10,23 +10,20 @@ possible_engines <- function(object, ...) {
   unique(engs$engine)
 }
 
+glue_compatible_engines <- function(engine, avail_eng) {
```
DavisVaughan, line 13: Same comments as above

DavisVaughan, line 13: I think you can just pass the `object` through, then look up the available engines to make the error message

juliasilge: Actually we don't need the object at all since we already tested it and already have the available engines/modes; we can just pass the available engines/modes through.

(I believe we do need to do the look-up outside these little functions because we need them to see if the engine/mode is one of the available options.)

DavisVaughan: Oh I see, you're right!

PR #508, opened 2021-06-01
Export `convert_*()` functions

### LINKED ISSUE #507: Export `convert_form_to_xy_*()` functions
For {censored} it would be handy to be able to use `convert_form_to_xy_fit()` and `convert_form_to_xy_new()`

hfrick:
Closes #507

This PR exports `convert_form_to_xy_fit()` and `convert_form_to_xy_new()` so they can be used in parsnip-adjacant packages like censored.

Should this also export `convert_xy_to_form_fit()` and `convert_xy_to_form_new()`? The former has a `# TODO slots for other roles` comment which makes me think maybe not yet?

Should this go into the NEWS file?


DavisVaughan:
I think I would prefix these functions with a `.` like `.convert_*()` to emphasize that they are not for general usage. Also mark them as `@keyword internal`. If we add a NEWS bullet (which I feel can be useful for our own tracking purposes), I would make sure it mentions that these are for developer usage.

R/convert_data.R
```diff
@@ -1,29 +1,48 @@
...
+#' of `lm` (and also see the notes at
+#' https://developer.r-project.org/model-fitting-functions.html).
+#'
+#' `.convert_form_to_xy_fit()` and `.convert_xy_to_form_fit()` are for when the
+#' data are created for modeling.
+#' `.convert_form_to_xy_fit()` saves both the data objects as well as the objects
+#' needed when new data are predicted (e.g. `terms`, etc.).
+#'
+#' `.convert_form_to_xy_new()` and `.convert_xy_to_form_new()` are used when new
+#' samples are being predicted and only require the predictors to be available.
```
DavisVaughan, line 17: The first paragraph will show up as the Description, while the following are in Details, is this intended?

R/convert_data.R
```diff
@@ -1,29 +1,48 @@
...
-
-# `convert_form_to_xy_new` is used when new samples are being predicted
-# and only requires the predictors to be available.
-
+#' Helper functions to convert between formula and matrix interface
+#'
+#' Functions to take a formula interface and get the resulting
+#' objects (y, x, weights, etc) back or the other way around. The functions are
+#' intended for developer use. For the most part, this emulates the internals
+#' of `lm` (and also see the notes at
```
DavisVaughan, line 8:
```suggestion
#' of `lm()` (and also see the notes at
```

R/convert_data.R
```diff
@@ -1,29 +1,48 @@
...
+#'   column after `model.matrix()` is finished.
+#' @inheritParams fit.model_spec
+#' @rdname convert_helpers
+#' @keywords internal
+#' @export
+#'
 #' @importFrom stats .checkMFClasses .getXlevels delete.response
 #' @importFrom stats model.offset model.weights na.omit na.pass
+.convert_form_to_xy_fit <- function(formula,
+                                   data,
```
DavisVaughan, line 39: The indentation got slightly off from the `.`. Ditto for the other exported functions

R/convert_data.R
```diff
@@ -1,29 +1,48 @@
...
+                                   ...,
+                                   na.action = na.omit,
```
DavisVaughan, line 41: So we omit for modeling, but pass for prediction? I guess this is probably right. VS passing for both

R/convert_data.R
```diff
@@ -1,29 +1,48 @@
...
+#' samples are being predicted and only require the predictors to be available.
+#'
+#' @param data A data frame containing all relevant variables (e.g. outcome(s),
+#'   predictors, case weights, etc).
+#' @param ... Additional arguments passed to [stats::model.frame()] and
+#'   specification of `offset` and `contrasts`.
+#' @param na.action A function which indicates what should happen when the data
+#'   contain NAs.
+#' @param indicators A string describing whether and how to create
+#'   indicator/dummy variables from factor predictors.
```
DavisVaughan, line 26: It might be worth mentioning what the alternatives are, besides `"traditional"`

DavisVaughan (approved):

R/convert_data.R
hfrick: Looking at the help page again, I think it's too far down if it's in Details so I've moved it up into the Description section, thanks!

R/convert_data.R
hfrick: good spot! I've fixed it and ran styler of the functions in general

R/convert_data.R
hfrick: I've added `"none"`, `"traditional"`, and `"one_hot"` :+1:

PR #512, opened 2021-06-10
add gen_additive_mod

mdancho84:
Hey @topepo et al, 
Here is the work-in-progress version of the `gen_additive_mod()` function. I'm OK with changing anything, I'd just like to see GAM's get into `parsnip`.


R/gen_additive_mod_data.R
```diff
@@ -0,0 +1,209 @@
...
+
+      if (ncol(tbl) == 1) {
+        res <- prob_to_class_2(tbl, object) %>%
+          tibble::as_tibble() %>%
+          stats::setNames("values") %>%
+          dplyr::mutate(values = as.factor(values))
+      } else{
+        res <- tbl %>%
+          apply(., 1, function(x)
+            which(max(x) == x)[1]) - 1 %>% #modify in the future for something more elegant when gets the formula ok
```
topepo, line 160: Can you talk about this for a min? What is going on there?

topepo:
Thanks for send this in. 

Do we want to have `gam` as an engine? They frequently update the package to the most recent version of R and make breaking changes (without much developer sympathy/consideration). At the minimum, I'd like to make the default engine `mgcv`. 

We're almost done with a new documentation system for `parsnip` and I've adjusted the doc file (incompletely) to prepare for this. (I'll do a `modeltime` PR to show how it is done; it is opt-in).

R/gen_additive_mod_data.R
mdancho84: Hey, Alberto wrote this part. I need to ask.

mdancho84:
> Thanks for send this in.
> 
> Do we want to have `gam` as an engine? They frequently update the package to the most recent version of R and make breaking changes (without much developer sympathy/consideration). At the minimum, I'd like to make the default engine `mgcv`.
> 
> We're almost done with a new documentation system for `parsnip` and I've adjusted the doc file (incompletely) to prepare for this. (I'll do a `modeltime` PR to show how it is done; it is opt-in).

Hey, I'll leave this one up to you. `mgcv::gam()` is the function that is being used, so maybe `mgcv` is fine.

R/gen_additive_mod_data.R
AlbertoAlmuinha: Hi @topepo ,

The intent of that part of the code is to obtain the class in a multi-class classification problem. If I remember correctly, that part of the code is not tested precisely because we didn't know how to implement the multiclass classification models, you can see the issue we opened in the following link (the problem was specifically with the type of "formula" used with these models):

https://github.com/business-science/gammodels/issues/11

R/gen_additive_mod.R
```diff
@@ -0,0 +1,166 @@
...
+#'
+#' @details
+#'
+#' This function only defines what _type_ of model is being fit. Once an engine
+#'  is specified, the _method_ to fit the model is also defined.
+#'
+#' The model is not trained or fit until the [fit.model_spec()] function is used
+#' with the data.
+#'
+#' __gam__
```
hfrick, line 40: The documentation style doesn't match the current one, neither does it match the new work-in-progress one. Do we want to use either of these @topepo ?

R/gen_additive_mod_data.R
```diff
@@ -0,0 +1,257 @@
+
+set_new_model("gen_additive_mod")
```
hfrick, line 2: don't we also need a `set_model_mode()`?

R/gen_additive_mod_data.R
```diff
@@ -0,0 +1,257 @@
...
+        nms <- names(res)
+        res <- res[, 2:1]
+        names(res) <- nms
+      }
+
+      if (object$spec$method$pred$conf_int$extras$std_error) {
+        res$.std_error <- results$se.fit
+      }
+      res
+    },
```
hfrick, line 99: Could this all go into its own function instead of being defined inline? it's a bit long for that.

R/gen_additive_mod_data.R
```diff
@@ -0,0 +1,257 @@
...
+      hi_nms <- paste0(".pred_upper_", object$lvl)
+      colnames(res_1) <- c(lo_nms[1], hi_nms[1])
+      colnames(res_2) <- c(lo_nms[2], hi_nms[2])
+      res <- bind_cols(res_1, res_2)
+
+      if (object$spec$method$pred$conf_int$extras$std_error) {
+        res$.std_error <- results$se.fit
+      }
+      res
+    },
```
hfrick, line 246: same comment as on the corresponding part for the regression mode

tests/testthat/test_gen_additive_model.R
```diff
@@ -0,0 +1,99 @@
...
+
+data(two_class_dat, package = "modeldata")
+
+# ------------------------------------------------------------------------------
+
+context("generalized additive models")
+
+# ------------------------------------------------------------------------------
+
+reg_mod <- gen_additive_mod(select_features = TRUE) %>% set_engine("mgcv") %>% set_mode("regression")
```
hfrick, line 15: could we move this into the `test_that()` so that it is self-contained?

tests/testthat/test_gen_additive_model.R
```diff
@@ -0,0 +1,99 @@
...
+    mgcv_ci$fit - qt(0.025, df = mgcv_mod$df.residual, lower.tail = FALSE) * mgcv_ci$se.fit
+  expect_equivalent(f_ci[[".pred_lower"]], unname(lower))
+
+})
+
+
+
+# ------------------------------------------------------------------------------
+
+cls_mod <- gen_additive_mod(adjust_deg_free = 1.5) %>% set_engine("mgcv") %>% set_mode("classification")
```
hfrick, line 58: this one could also easily go into the call to `test_that()` to make that self-contained

hfrick (changes requested):

topepo (approved):

hfrick (approved):

Issue #513, opened 2021-06-10
Adding an engine argument to model_spec functions (with default)


dgrtwo:
Could you consider having an `engine = ` argument to each model_spec function, rather than requiring `set_engine()` afterward? For instance:

```
# Instead of requiring:
logistic_reg() %>%
  set_engine("glm")

# Allow
logistic_reg(engine = "glm")
```

This is more compact, particularly since it becomes more straightforward to pass a model spec as an argument (e.g. to `set_model()`) without giving it its own two lines of code for assignment. (Besides being shorter, not having to create an object means not having to give it a name like `lin_mod` that might need to change in two places if you change the model spec).

I'd also find it a more natural workflow if these engines had defaults, like `"lm"` for `linear_reg()`. parsnip is already opinionated about the model if `set_engine()` is not called, like choosing `"ranger"` for `rand_forest()`, though it gives a warning about that. (Some like `nearest_neighbor()` or `prophet()` have only one engine implemented, but still give the warning, which seems especially harsh).

---

None of this precludes `set_engine()` still keeping its current functionality for the sake of reverse compatibility, ability to modify an existing recipe. and setting extra arguments (Similarly, in ggplot2 you can do either `ggplot(data, aes(...))` or `ggplot(data) + aes(...)`).

topepo:
I'll start a branch for this to test out. I like it but there are some additional things that we need to think about. 

For example, there are going to be model definitions in `parsnip` where all of the engines are in separate packages. We are about to add generalized additive models and that is probably how it will happen. 

In that case, we could default `engine = "mgvc"` without error in the model function but almost everything else would error (like `translate()`) without the other package loaded. So, we'll need to see if that type of model function should not have a default or error trap the supporting functions well (again, if the other package is not loaded).

I very much like the idea of a one-line path to an `lm()` model spec (and so on). I'd love to be able to let the data assign the model mode, but we tried really hard to do that and it can't happen (we need the code temp[late before model fit time and that is mode-specific).

dgrtwo:
> So, we'll need to see if that type of model function should not have a default or error trap the supporting functions well (again, if the other package is not loaded).

Is there a reason not to throw an error immediately?

I'm not sure this presents a different problem than the `set_engine()` syntax (they could still do `set_engine("mgvc")`, and then run into errors later). Yes it's explicit rather than setting a default, but just because the user types `set_engine("mgvc")` doesn't mean they know that this relies on an external dependency and confirmed that they had the package installed.

[...]

PR #515, opened 2021-06-15
default engine changes for #513

topepo:
Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. 

This also adds functionality to change to declare an engine in the model specification function. 

`set_engine()` is still required if engine-specific arguments need to be added

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # parsnip (development version)
 
-* Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to change to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
+* Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
 
 * The default engine for `multinom_reg()` was changed to `nnet`. 
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 # parsnip (development version)
 
+* Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to change to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
+
+* The default engine for `multinom_reg()` was changed to `nnet`. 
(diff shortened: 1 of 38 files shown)


R/boost_tree.R
```diff
@@ -92,6 +94,7 @@
 
 boost_tree <-
   function(mode = "unknown",
+           engine = "xgboost",
```
DavisVaughan, line 97: Just want to explicitly bring up that placing this argument here is technically a breaking change. If users had been doing `boost_tree("classification", 2)`, then that will no longer work correctly. This is _probably_ pretty rare? Just wanted to bring it up.

DavisVaughan (approved):

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # parsnip (development version)
 
+* Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to change to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
```
hfrick, line 3:
```suggestion
* Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
```

hfrick (approved):

R/boost_tree.R
topepo: I agree but think that it is a minimal issue. I can't imagine very many people using a series of unnamed arguments for these functions.

juliasilge:
This is a really nice change; I am super happy about it.

I just made a few edits to how we describe `mode` and `engine` so they are more consistent and more in line with how we talk about them elsewhere, like [here](https://www.tmwr.org/models.html#create-a-model) and [here](https://parsnip.tidymodels.org/articles/parsnip_Intro.html). For example, I don't think we should use the word "type" for either of those since we use it for the algorithm choice.

juliasilge (approved):

juliasilge:
I can't imagine there will be any problems but have any of us run [extratests](https://github.com/tidymodels/extratests) with this PR built?

topepo:
> I can't imagine there will be any problems but have any of us run extratests with this PR built?

`[ FAIL 0 | WARN 0 | SKIP 6 | PASS 252 ]`

PR #524, opened 2021-06-29
updated GAM documentation

topepo:
Also sets the default engine (to the only engine, `mgcv`)


R/gen_additive_mod.R
```diff
@@ -1,79 +1,43 @@
...
-#'  is specified, the _method_ to fit the model is also defined.
-#'
-#' The model is not trained or fit until the [fit.model_spec()] function is used
-#' with the data.
-#'
-#' __gam__
+#'  smoothing parameter is estimated as zero then the extra penalty has no
+#'  effect. Use `adjust_deg_free` to increase level of penalization.
+#' @param adjust_deg_free If `select_features = TRUE`, then acts as a multiplier
+#'  for smoothness. Increase this beyond 1 to produce smoother models.
```
hfrick, line 23: This description reads a bit "unsmooth". When I read it, I fee like it's occasionally missing the noun of a sentence, e.g., in "If this is TRUE then can add"?

hfrick: `TRUE` and `FALSE` should be in code styling :nerd_face:

man/rmd/gen_additive_mod_mgcv.Rmd
```diff
@@ -0,0 +1,83 @@
...
+```{r glm-cls}
+gen_additive_mod(adjust_deg_free = numeric(1), select_features = logical(1)) %>% 
+  set_engine("mgcv") %>% 
+  set_mode("classification") %>% 
+  translate()
+```
+
+## Model fitting
+
+This model should use used with a model formula so that smooth terms can be specified. For example:
```
hfrick, line 58:
```suggestion
This model should be used with a model formula so that smooth terms can be specified. For example:
```

PR #527, opened 2021-07-08
Add an article with minimal examples

### LINKED ISSUE #519: add fitting and prediction examples to pkgdown
In the new doc system, we _had_ simple examples. We can move them here and have good naming conventions (for linking), collapsable sections etc.

hfrick:
Closes #519 

This PR adds an article with minimal examples for fitting and predicting.

Note that this currently sets 
```
development:
  mode: release
```
so that the site gets build and deployed to https://parsnip.tidymodels.org - this means that the site that will be online does include any changes which are part of the devel version of the pkg rather than just the CRAN version. The header of the site should indicate that it's the devel version via the version number. 

We'll change the mode back to 
```
development:
  mode: auto
```
after this PR but the site will stay like this until the next CRAN release. 

To add more examples, add a 

```
  <details id="model-spec-engine"> 
  </details>
```
section to the `vignettes/articles/Example.Rmd`. The indentation is necessary for the section to be properly collapsed in the pkgdown site. For the id, the convention is model-spec-engine where any capitalization in the engine name is kept, e.g., `id="svm-linear-LiblineaR"`. The id is used on the `<model_engine>.Rmd` help pages to link to the relevant section in the article.

### DIFF
--- a/vignettes/articles/template-cls-multi-class.R
+++ b/vignettes/articles/template-cls-multi-class.R
@@ -1,6 +1,6 @@
 #' <h3>Classification Example</h3>
 
-#' We'll model the island of the penguins with two predictors in the same unit (mm): bill length and bill depth.
+#' We'll predict the island where the penguins were observed with two variables in the same unit (mm): bill length and bill depth.
 
 #+ results = "hide", messages = FALSE
 library(tidymodels)


juliasilge:
Addresses #519

topepo:
I suggest trying to avoid links for function names in the section headings. It is easy to click on those when trying to expand a section and be taken to the function's man page. 

Maybe use

```r
<summary><tt>multinom_reg()</tt> with the `"glmnet"` engine</summary>
```

instead of 
```r
<summary>`multinom_reg()` with the `"glmnet"` engine</summary>
```

(or some way to avoid `pkgdown` doing the links)

hfrick:
There are now examples for all model/engine combinations with the exception of survival models and `spark` engines.

vignettes/articles/Examples.Rmd
```diff
@@ -0,0 +1,1371 @@
...
+
+```{r startup, include = FALSE}
+library(utils)
+library(ggplot2)
+theme_set(theme_bw())
+```
+
+This is a collection of examples for fitting and predicting with various models and engines. 
+
+Model/engine combinations which can be used for different modes have an example each. For regression, we use the Chicago ridership data. For classification, we use an artificial dataset for a binary example and the Palmer penguins data for a multi-class example.
```
juliasilge, line 18:
```suggestion
These examples show how to *fit* and *predict* with different combinations of model, mode, and engine. As a reminder, in parsnip, 

- the **model type** differentiates basic modeling approaches, such as random forests, logistic regression, linear support vector machines, etc.,

- the **mode** denotes in what kind of modeling context it will be used (most commonly, classification or regression), and

- the computational **engine** indicates how the model is fit, such as with a specific R package implementation or even methods outside of R like Keras or Stan.

The following examples use consistent data sets throughout. For regression, we use the Chicago ridership data. For classification, we use an artificial data set for a binary example and the Palmer penguins data for a multiclass example.
```

vignettes/articles/template-cls-multi-class.R
```diff
@@ -0,0 +1,12 @@
+#' <h3>Classification Example</h3>
+
+#' We'll model the island of the penguins with two predictors in the same unit (mm): bill length and bill depth.
```
juliasilge, line 3:
```suggestion
#' We'll predict the island where the penguins were observed with two variables in the same unit (mm): bill length and bill depth.
```

juliasilge (approved):
I think it would be good to add this info yet again here in this bit of documentation, just briefly.

topepo (approved):

Issue #500, opened 2021-05-20
do we need the internal parsnip objects?


topepo:
[These lines](https://github.com/tidymodels/parsnip/blob/adf0f323d28d0aaa45348dddf9dd41992d2adc58/R/aaa_models.R#L29) may not be needed (but perhaps for docs).

PR #530, opened 2021-07-16
better engine and mode checking code

### LINKED ISSUE #529: Error messages for incorrect mode+engine combination
Not all engines are available for all modes of a model. It would be nice if an unavailable combination of mode+engine would throw an informative error.

Currently, you can specify such an incorrect combination (the order of `set_engine()` and `set_mode()` doesn't matter) and then `fit()` errors with a message that doesn't refer to the mode+engine combination.

``` r
library(parsnip)
data(mtcars)

# engine C5.0 is only available for classification

# neither of the two specifications error
dt_reg_spec <- decision_tree() %>% 
  set_mode("regression") %>% 
  set_engine("C5.0")

dt_reg_spec <- decision_tree() %>% 
  set_engine("C5.0") %>% 
  set_mode("regression")

# fit() errors but the message doesn't point to the actual problem
dt_reg_spec %>% fit(mpg ~ ., data = mtcars)
#> Error: formula_ is unknown.
```

<sup>Created on 2021-07-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

topepo:
closes #529 

``` r
library(parsnip)
data(mtcars)

# engine C5.0 is only available for classification

# neither of the two specifications error
dt_reg_spec <- decision_tree() %>% 
  set_mode("regression") %>% 
  set_engine("C5.0")
#> Error: Available modes for engine C5.0 are: 'unknown', 'classification'

dt_reg_spec <- decision_tree() %>% 
  set_engine("C5.0") %>% 
  set_mode("regression")
#> Error: Available modes for engine C5.0 are: 'unknown', 'classification'
```

<sup>Created on 2021-07-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,7 +6,7 @@
 
 * Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
 
-* parsnip now checks for a valid engine for a given mode (#529)
+* parsnip now checks for a valid combination of engine and mode (#529)
 
 * The default engine for `multinom_reg()` was changed to `nnet`. 
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,14 +1,21 @@
 # parsnip (development version)
 
+## Model Specification Changes
+
+* A model function (`gen_additive_mod()`) was added for generalized additive models. 
+
 * Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
 
+* parsnip now checks for a valid engine for a given mode (#529)
+
 * The default engine for `multinom_reg()` was changed to `nnet`. 
+
+## Other Changes
 
 * The helper functions `.convert_form_to_xy_fit()`, `.convert_form_to_xy_new()`, `.convert_xy_to_form_fit()`, and  `.convert_xy_to_form_new()` for converting between formula and matrix interface are now exported for developer use (#508).
 
 * Fix bug in `augment()` when non-predictor, non-outcome variables are included in data (#510).
 
-* A model function (`gen_additive_mod()`) was added for generalized additive models. 
 
 # parsnip 0.1.6
 
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -135,19 +135,24 @@ check_mode_val <- function(mode) {
 }
 
 
-stop_incompatible_mode <- function(spec_modes, eng) {
-  if (is.null(eng)) {
-    msg <- glue::glue(
-      "Available modes are: ",
-      glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
-    )
-  } else {
-    msg <- glue::glue(
-      "Available modes for engine {eng} are: ",
-      glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
-    )
+stop_incompatible_mode <- function(spec_modes, eng = NULL, cls = NULL) {
+  if (is.null(eng) & is.null(cls)) {
+    msg <- "Available modes are: "
+  }
+  if (!is.null(eng) & is.null(cls)) {
+    msg <- glue::glue("Available modes for engine {eng} are: ")
+  }
+  if (is.null(eng) & !is.null(cls)) {
+    msg <- glue::glue("Available modes for model type {cls} are: ")
+  }
+  if (!is.null(eng) & !is.null(cls)) {
+    msg <- glue::glue("Available modes for model type {cls} with engine {eng} are: ")
   }
 
+  msg <- glue::glue(
+    msg,
+    glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
+  )
   rlang::abort(msg)
 }
 
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -1,5 +1,7 @@
 # Initialize model environments
 
+all_modes <- c("classification", "regression", "censored regression")
+
 # ------------------------------------------------------------------------------
 
 ## Rules about model-related information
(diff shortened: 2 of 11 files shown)


NEWS.md
```diff
@@ -1,14 +1,21 @@
 # parsnip (development version)
 
+## Model Specification Changes
+
+* A model function (`gen_additive_mod()`) was added for generalized additive models. 
+
 * Each model now has a default engine that is used when the model is defined. The default for each model is listed in the help documents. This also adds functionality to declare an engine in the model specification function. `set_engine()` is still required if engine-specific arguments need to be added. (#513)
 
+* parsnip now checks for a valid engine for a given mode (#529)
```
hfrick, line 9: and a valid mode for a given engine
```suggestion
* parsnip now checks for a valid combination of engine and mode (#529)
```

R/aaa_models.R
```diff
@@ -1,5 +1,7 @@
 # Initialize model environments
 
+all_modes <- c("classification", "regression", "censored regression")
```
hfrick, line 3: Just noting here that [a few lines down](https://github.com/tidymodels/parsnip/blob/a8f1cd611a3b3a8a82d9eefaaf393050bb16b8d0/R/aaa_models.R#L29) there is also 
`parsnip$modes <- c("regression", "classification", "unknown")` 
That's missing the `"censored regression"` mode and is possibly unused (#500). Given that this PR just moves the already existing `all_modes` from one file to another, should we deal with the possible duplication in #500 or here?

R/aaa_models.R
```diff
@@ -134,25 +136,90 @@ check_mode_val <- function(mode) {
...
+  model_info <- rlang::env_get(get_model_env(), cls)
+
+  # Cases where the model definition is in parsnip but all of the engines
+  # are contained in a different package
+  if (nrow(model_info) == 0) {
+    return(invisible(NULL))
+  }
+
+  # ------------------------------------------------------------------------------
+  # First check engine against any mode
```
hfrick, line 179:
```suggestion
  # First check engine against any mode for the given model class
```

hfrick:
I've tried out a few combinations as additional test cases:
* for model classes without engines we do know the possible mode(s) so it would be nice to check for that 
* when first setting the mode, then the engine, we could show available engine given the mode rather than the other way around
* when using `set_mode()` without arguments, the error message is that the engine is missing (and vice versa) - this could be turned into a separate issue though?

``` r
library(parsnip)

# check mode for given model class ----------------------------------------

# can we have this error?
proportional_hazards() %>% set_mode("regression")
#> Proportional Hazards Model Specification (regression)
#> 
#> Computational engine: survival
# similar to how this errors
linear_reg() %>% set_mode("censored regression")
#> Error: Available modes for engine lm are: 'unknown', 'regression'


# check engine for given model class and mode  ----------------------------

# do we want it to send a message about available engines for the mode (rather than modes for the engine)?
dt_reg_spec <- decision_tree() %>% 
  set_mode("regression") %>% 
  set_engine("C5.0")
#> Error: Available modes for engine C5.0 are: 'unknown', 'classification'


# check for missing mode arg ----------------------------------------------

# shouldn't this error and complain about missing mode (instead of engine)?
linear_reg() %>% set_mode()
#> Error in stop_incompatible_mode(spec_modes): argument "eng" is missing, with no default

# check for missing engine arg --------------------------------------------

# shouldn't this error and complain about missing engine (instead of mode)?
linear_reg() %>% set_engine()
#> Error in eval(parse(text = text, keep.source = FALSE), envir): argument "mode" is missing, with no default
```

<sup>Created on 2021-07-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

R/aaa_models.R
topepo: I reduced it down to one source. I think that we do need these as data in the package and will close #500

topepo:
Fixed these issues ☝️ except:

> when first setting the mode, then the engine, we could show available engine given the mode rather than the other way around

That's not trivial to figure out since they are separate function calls. I'm good with the current message.

R/aaa_models.R
```diff
@@ -134,25 +135,114 @@ check_mode_val <- function(mode) {
...
+    )
+  } else {
+    msg <- glue::glue(
+      "Available modes for engine {eng} are: ",
+      glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
+    )
+  }
+
+  rlang::abort(msg)
+}
```
hfrick, line 152: When we use it with the model class passed as the `eng` argument, it prints messages like "Available modes for *engine linear_reg* are:"  so here's a suggestion on how to change that:
```suggestion
stop_incompatible_mode <- function(spec_modes, eng = NULL, cls = NULL) {
  if (is.null(eng) & is.null(cls)) {
    msg <- "Available modes are: "
  }
  if (!is.null(eng) & is.null(cls)) {
    msg <- glue::glue("Available modes for engine {eng} are: ")
  }
  if (is.null(eng) & !is.null(cls)) {
    msg <- glue::glue("Available modes for model type {cls} are: ")
  }
  if (!is.null(eng) & !is.null(cls)) {
    msg <- glue::glue("Available modes for model type {cls} with engine {eng} are: ")
  }

  msg <- glue::glue(
    msg,
    glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
  )
  rlang::abort(msg)
}
```

R/aaa_models.R
```diff
@@ -134,25 +135,114 @@ check_mode_val <- function(mode) {
...
+    stop_incompatible_engine(spec_engs, mode)
+  }
+
   invisible(NULL)
 }
 
+check_mode_with_no_engine <- function(cls, mode) {
+  spec_modes <- get_from_env(paste0(cls, "_modes"))
+  if (!(mode %in% spec_modes)) {
+    stop_incompatible_mode(spec_modes, cls)
```
hfrick, line 242:
```suggestion
    stop_incompatible_mode(spec_modes, cls = cls)
```

R/arguments.R
```diff
@@ -79,9 +79,9 @@ set_mode <- function(object, mode) {
   cls <- class(object)[1]
   if (rlang::is_missing(mode)) {
     spec_modes <- rlang::env_get(get_model_env(), paste0(cls, "_modes"))
-    stop_incompatible_mode(spec_modes)
+    stop_incompatible_mode(spec_modes, cls)
```
hfrick, line 82:
```suggestion
    stop_incompatible_mode(spec_modes, cls = cls)
```

tests/testthat/test_args_and_modes.R
```diff
@@ -49,6 +49,60 @@ test_that('pipe engine', {
...
+  expect_error(
+    decision_tree(engine = NULL)%>%
+      set_mode("regression") %>%
+      set_engine("C5.0"),
+    "Available modes for engine C5"
+  )
+
+  expect_error(
+    proportional_hazards() %>% set_mode("regression"),
+    "Available modes for engine proportional_hazards"
```
hfrick, line 88:
```suggestion
    "Available modes for model type proportional_hazards"
```

tests/testthat/test_args_and_modes.R
```diff
@@ -49,6 +49,60 @@ test_that('pipe engine', {
...
+  )
+
+  expect_error(
+    linear_reg() %>% set_mode(),
+    "Available modes for engine linear_reg"
```
hfrick, line 93:
```suggestion
    "Available modes for model type linear_reg"
```

hfrick (commented):
Sounds good to me. I found one more thing and added a suggestion to change it

``` r
library(parsnip)

# these messages call a model type an engine
decision_tree() %>% set_mode()
#> Error: Available modes for engine decision_tree are: 'unknown', 'classification', 'regression'
proportional_hazards() %>% set_mode()
#> Error: Available modes for engine proportional_hazards are: 'unknown', 'censored regression'
proportional_hazards() %>% set_mode("regression")
#> Error: Available modes for engine proportional_hazards are: 'unknown', 'censored regression'
```

<sup>Created on 2021-07-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0.9000)</sup>

PR #536, opened 2021-07-21
cleanup after examples article

hfrick:
as mentioned in #527, the dev mode for pkgdown should go back to `auto` - and I've added the missing bullet in the NEWS file


juliasilge (approved):
🚀

hfrick:
thanks :raised_hands:

PR #556, opened 2021-09-22
Fix TOC of Examples article

hfrick:
The article now has engine-specific section headings to reduce direct repetition in the table of contents. So now you get 
image
instead of 
image


juliasilge (approved):
Looks great! 🙌

PR #560, opened 2021-09-23
Dynamically generate list of engines in `set_engine()` docs

DavisVaughan:
https://user-images.githubusercontent.com/19150088/134585206-c873db83-e9a0-4d03-b9ed-9f00150eaf71.mov

### DIFF
--- a/R/engines.R
+++ b/R/engines.R
@@ -55,6 +55,12 @@ load_libs <- function(x, quiet, attach = FALSE) {
 #' `set_engine()` is used to specify which package or system will be used
 #'  to fit the model, along with any arguments specific to that software.
 #'
+#' @section Engines:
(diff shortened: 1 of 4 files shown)


R/engines.R
```diff
@@ -55,6 +55,12 @@ load_libs <- function(x, quiet, attach = FALSE) {
 #' `set_engine()` is used to specify which package or system will be used
 #'  to fit the model, along with any arguments specific to that software.
 #'
+#' @section Engines:
+#' The following lists the set of engines available to each model
```
topepo, line 59:
```suggestion
#' Based on the currently loaded packages, the following lists the set of engines available to each model
```

topepo (commented):
I agree with @EmilHvitfeldt that we need something after the model functions that have no engines available.

Issue #492, opened 2021-05-14
Disable `...` for `fit()`


hfrick:
Currently, the documentation for `fit()` says for `...`

> Not currently used; values passed here will be ignored. Other options required to fit the model should be passed using `set_engine()`.

However, some arguments get passed through in some cases, see examples below. Any additional arguments should be specified via `set_engine()` and `fit()` should error informatively if the dots are used.

Current status:
* formula to formula: fails silently
* formula to xy: allowed args are `c("subset", "weights", "contrasts", "offset")`
* xy to xy: fails in `eval_mod()`
* xy to formula: fails in `eval_mod()`

``` r
library(parsnip)

nrow(mtcars) # 32 obs total
#> [1] 32

# formula to formula: fails silently
fit_parsnip <- linear_reg() %>% 
  set_engine("lm") %>% 
  fit(mpg ~ disp + hp, data = mtcars, subset = mtcars$cyl == 6)
fit_parsnip$fit %>% fitted() %>% length()
#> [1] 32

# formula to xy: works (when it shouldn't)
fit_parsnip <- linear_reg(penalty = 0) %>% 
  set_engine("glmnet") %>% 
  fit(mpg ~ disp + hp, data = mtcars, subset = mtcars$cyl == 6)
fit_parsnip$fit$nobs 
#> [1] 7

# xy to xy: fails but could have better handling
fit_parsnip <- linear_reg(penalty = 0) %>% 
  set_engine("glmnet") %>% 
  fit_xy(x = mtcars[, c("disp", "hp")], y = mtcars$mpg, subset = mtcars$cyl == 6)
#> Error in eval_tidy(e, ...): unused argument (subset = c(TRUE, TRUE, FALSE, TRUE, FALSE, TRUE, FALSE, FALSE, FALSE, TRUE, TRUE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, FALSE, TRUE, FALSE, FALSE))
#> Timing stopped at: 0.001 0 0.001

# xy to formula: fails but could have better handling
fit_parsnip <- linear_reg(penalty = 0) %>% 
  set_engine("lm") %>% 
  fit_xy(x = mtcars[, c("disp", "hp")], y = mtcars$mpg, subset = mtcars$cyl == 6)
[...]

PR #569, opened 2021-10-07
Remove all mentions of `contrasts` as an argument

DavisVaughan:
Part of https://github.com/tidymodels/parsnip/issues/492. I will do multiple PRs related to this.

This PR removes all mentioning and usage of `contrasts` as an argument.

You can still set the global `contrasts` option and the `.convert_form_*()` functions will respect that. I've added a test case to show that.


topepo:
Yes.

(both please)

PR #571, opened 2021-10-08
move method for `tune_args()` from tune to here

hfrick:
This PR is part of moving methods for `tune_args()` from tune to the package which holds the corresponding object class:

* the methods for `recipe`, `step`, and `check` are moved to recipes https://github.com/tidymodels/recipes/pull/830
* the method for `model_spec` is moved to parsnip (aka this PR here)
* the method for `workflow` is moved to workflows https://github.com/tidymodels/workflows/pull/123

In all cases, the methods are registered conditionally on the version of tune from which the methods have been removed. The tune PR is https://github.com/tidymodels/tune/pull/420

The tests require multiple packages so they are all in extratests: https://github.com/tidymodels/extratests/pull/30

The PR on extratests should be merged first, then the ones on recipes and parsnip, then workflows, then tune.

This is part of the wider aim of introducing `extract_parameter_dials()` and `extract_parameter_set_dials()` and the next steps are:
* moving the methods for `tunable()` similar to this PR set
* moving the methods for `parameters()` similar to this PR set
* introducing `extract_parameter_set_dials()`
* introducing `extract_parameter_dials()`
* updating documentation and learning materials

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -77,5 +77,3 @@ Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.2
-Remotes: 
-    r-lib/generics


DESCRIPTION
```diff
@@ -77,3 +77,5 @@ Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.2
+Remotes: 
+    r-lib/generics
```
DavisVaughan, line 81: We are going to wait until generics hits cran to merge this, and we will update to use CRAN generics once that happens

PR #591, opened 2021-11-04
move doc for `update.poisson_reg()` 

hfrick:
to be with the other `update()` methods


hfrick:
Is it okay to ignore this one failing action? :thinking:

PR #592, opened 2021-11-04
Update Examples article

hfrick:
add examples for `bart()` and `gen_additive_mod()` which have engines in parsnip


juliasilge:
If you merge in #591 and then rerun the actions, this should be OK

PR #594, opened 2021-11-08
check for non-numeric cols when converting to matrix

topepo:
`maybe_matrix()` not checks and gives an informative error when non-numeric columns are present.

### DIFF
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -356,6 +356,14 @@ check_dup_names <- function(x, y) {
 maybe_matrix <- function(x) {
   inher(x, c("data.frame", "matrix", "dgCMatrix"), cl = match.call())
   if (is.data.frame(x)) {
+    cat_cols <- vapply(x, function(x) !is.numeric(x), logical(1))
+    if (any(cat_cols)) {
+      cat_cols <- names(cat_cols)[cat_cols]
+      cat_cols <- paste0("'", cat_cols, "'", collapse = ", ")
+      msg <- glue::glue("Some columns are non-numeric. The data cannot be",
+                        "converted to numeric matrix: {cat_cols}")
+      rlang::abort(msg)
+    }
     x <- as.matrix(x)
   }
   # leave alone if matrix or sparse matrix
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -359,9 +359,9 @@ maybe_matrix <- function(x) {
     cat_cols <- vapply(x, function(x) !is.numeric(x), logical(1))
     if (any(cat_cols)) {
       cat_cols <- names(cat_cols)[cat_cols]
-      cat_cols <- paste0("'", cat_cols, "'", collapse = ", ")
-      msg <- glue::glue("Some columns are non-numeric. The data cannot be",
-                        "converted to numeric matrix: {cat_cols}")
+      cat_cols <- glue::glue_collapse(glue::single_quote(cat_cols), sep = ", ")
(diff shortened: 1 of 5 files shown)


R/convert_data.R
```diff
@@ -356,6 +356,14 @@ check_dup_names <- function(x, y) {
 maybe_matrix <- function(x) {
   inher(x, c("data.frame", "matrix", "dgCMatrix"), cl = match.call())
   if (is.data.frame(x)) {
+    cat_cols <- vapply(x, function(x) !is.numeric(x), logical(1))
```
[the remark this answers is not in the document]
DavisVaughan: They aren't necessarily categorical, so I'd just call these `non_num_cols` rather than `cat_cols`

R/convert_data.R
```diff
@@ -356,6 +356,14 @@ check_dup_names <- function(x, y) {
...
+    if (any(cat_cols)) {
+      cat_cols <- names(cat_cols)[cat_cols]
+      cat_cols <- paste0("'", cat_cols, "'", collapse = ", ")
```
DavisVaughan, line 362: Since we use glue 

```suggestion
      cat_cols <- glue::glue_collapse(glue::single_quote(cat_cols), sep = ", ")
```

R/convert_data.R
```diff
@@ -356,6 +356,14 @@ check_dup_names <- function(x, y) {
...
+      msg <- glue::glue("Some columns are non-numeric. The data cannot be",
+                        "converted to numeric matrix: {cat_cols}")
```
DavisVaughan, line 364:
```suggestion
                        "converted to numeric matrix: {cat_cols}.")
```

R/convert_data.R
```diff
@@ -356,6 +356,14 @@ check_dup_names <- function(x, y) {
 maybe_matrix <- function(x) {
   inher(x, c("data.frame", "matrix", "dgCMatrix"), cl = match.call())
   if (is.data.frame(x)) {
+    cat_cols <- vapply(x, function(x) !is.numeric(x), logical(1))
+    if (any(cat_cols)) {
+      cat_cols <- names(cat_cols)[cat_cols]
+      cat_cols <- paste0("'", cat_cols, "'", collapse = ", ")
+      msg <- glue::glue("Some columns are non-numeric. The data cannot be",
```
DavisVaughan, line 363:
```suggestion
      msg <- glue::glue("Some columns are non-numeric. The data cannot be ",
```

tests/testthat/test_convert_data.R
```diff
@@ -617,4 +617,16 @@ test_that("convert to matrix", {
     inherits(parsnip::maybe_matrix(Matrix::Matrix(as.matrix(mtcars), sparse = TRUE)),
              "dgCMatrix")
   )
+
+  data(ames, package = "modeldata")
+  expect_error(
```
DavisVaughan, line 622: We should really update parsnip to testthat 3e so we can do `expect_snapshot(error = TRUE, {code})`

DavisVaughan (changes requested):

R/convert_data.R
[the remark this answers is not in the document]
topepo: There could be _a lot_ of columns and I assume that `vapply()` is a smidgeon fast.

topepo:
Merging despite failures on windows. These are due to issue compiling the new xgboost version and, hopefully, binaries will be available soon.

PR #605, opened 2021-11-17
Set mode "censored regression"

### LINKED ISSUE #603: set censored regression mode in parsnip
For models, even those with engines in other packages, have their modes [defined in parsnip](https://github.com/tidymodels/parsnip/blob/main/R/bag_tree.R#L122:L124). 

Censored sets currently them and those lines should be move to parsnip.

hfrick:
closes #603

Sets the mode "censored regression" for bag_tree, boost_tree, and decision_tree here instead of in censored.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: parsnip
 Title: A Common API to Modeling and Analysis Functions
-Version: 0.1.7.9001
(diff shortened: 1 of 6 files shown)


topepo (approved):
It looks great. I upped the version number so you should use that version censored.

Issue #578, opened 2021-10-20
option to turn off `fit time` printing of fitted parsnip objects


EmilHvitfeldt:
When rerunning uncached code, it can be a little annoying to get git diffs because of a different fit time. Would it be possible to have a global option to turn it off? 

https://github.com/tidymodels/parsnip/blob/a97e938d4a967e3f777639efbd6398b62153ac09/R/fit.R#L421-L423

juliasilge:
Closed in #614

PR #614, opened 2021-11-30
Updates to execution time collection

### LINKED ISSUE #611: Add an option to set gcFirst to FALSE in the internal calls to system.time in form_form and xy_xy
I am not sure if this is a bug or if this should be handled as a feature. In [fit_helpers](https://github.com/tidymodels/parsnip/blob/main/R/fit_helpers.R), both `form_form` and `xy_xy` call `system.time` with the default value of `TRUE` for `gcFirst`. This triggers unnecessary garbage collection which slows down significantly e.g. `tune_grid` for simple models such as decision trees. For instance the following code:
```R
library(tidymodels)
library(mlbench)
data(PimaIndiansDiabetes)

my_grid <- expand.grid(min_n=2:50)
cv_folds <- vfold_cv(PimaIndiansDiabetes, v = 5, strata="diabetes")
my_model <- decision_tree(cost_complexity=0, min_n=tune()) %>% 
    set_engine("rpart",xval=0) %>% set_mode("classification")

tune_results <-  my_model %>% tune_grid(diabetes~.,
                                        resamples=cv_folds,
                                        grid=my_grid,
                                        metrics=metric_set(accuracy))
```
runs in roughly 50 seconds on my computer with `gcFirst=TRUE` but takes only 20 seconds with  `gcFirst=FALSE`. 

I'm not sure whether precise timing reporting is needed. If this is not the case, maybe `gcFirst=FALSE` should always be used. If some flexibility is needed, it could be a user visible option. In any case, the default to `gcFirst=TRUE` is adding a significant overhead to simple model fitting.

topepo:
Closes #611

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -15,6 +15,8 @@
 ## Other Changes
 
  * When the xy interface is used and the underlying model expects to use a matrix, a better warning is issued when predictors contain non-numeric columns (including dates). 
+ 
+ * The fit time is only calculated when the `verbosity` argument of `control_parsnip()` is 2L or greater. Also, the call to `system.time()` now uses `gcFirst = FALSE`. (#611)
 
 # parsnip 0.1.7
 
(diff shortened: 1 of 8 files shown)


juliasilge (approved):
This will be a really nice change. 🙌 

It will effect quite a lot of output in the long term (tidymodels.org, our book, various READMEs in various packages) but I think we should just change those as they come up.

EmilHvitfeldt:
Would these changes be enough to handle https://github.com/tidymodels/parsnip/issues/578 as well?

juliasilge:
Yep, also closes #578

PR #615, opened 2021-12-01
add argument `interval` for survival/quantile predictions

hfrick:
This is prep for https://github.com/tidymodels/censored/pull/125 and introduces the `interval` argument for predictions which allows to add confidence or prediction interval estimats to be added, if available.  

It makes use of the new arg for predictions of type `"survival"` and `"quantile"`, thus covering part of #616.

### DIFF
--- a/man/bart-internal.Rd
+++ b/man/bart-internal.Rd
@@ -4,7 +4,7 @@
 \alias{bart-internal}
 \alias{bartMachine_interval_calc}
 \alias{dbart_predict_calc}
-\title{Model predictions}
+\title{Developer functions for predictions via BART models}
 \usage{
 bartMachine_interval_calc(new_data, obj, ci = TRUE, level = 0.95)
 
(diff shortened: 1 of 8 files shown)


man/bart-internal.Rd
```diff
@@ -54,14 +54,12 @@ the unadulterated results of the prediction function.
 For censored regression:
 \itemize{
 \item \code{type = "time"} produces a column \code{.pred_time}.
-\item \code{type = "hazard"} results in a column \code{.pred_hazard}.
-\item \code{type = "survival"} results in a list column containing tibbles with a
-\code{.pred_survival} column.
+\item \code{type = "hazard"} results in a list column \code{.pred} containing tibbles
```
topepo, line 57: Any idea why this is affected?

hfrick: I assume it's because `bartMachine_interval_calc()` inherits parameter documentation from `predict.model_fit()` here: https://github.com/tidymodels/parsnip/blob/main/R/bart.R#L168

hfrick: should this only be `@inheritParams` instead of `@inherit` or did you purposefully want more than parameter documentation to be passed through?

topepo: It should use `@inheritParams`. Maybe this is true of other models too

hfrick: I'll change it then!

PR #618, opened 2021-12-07
deprecate fit_control in favor of control_parsnip

### LINKED ISSUE #609: deprecate 'fit_control()' in favor of 'control_parsnip()'
https://github.com/tidymodels/parsnip/blob/dd36a954d58c73b67a118e0b79f79f770e9ce703/R/control_parsnip.R#L17-L28

should be

```
#' @return An S3 object with class "control_parsnip" that is a named list with the 
```

EmilHvitfeldt:
This PR aims to close #609.

the typo was fixed in https://github.com/tidymodels/parsnip/pull/614, so the only change here is the official deprecation of `fit_control()`.


R/control_parsnip.R
```diff
@@ -32,6 +32,7 @@ control_parsnip <- function(verbosity = 1L, catch = FALSE) {
 #' @export
 #' @rdname control_parsnip
 fit_control <- function(verbosity = 1L, catch = FALSE) {
```
topepo, line 34: Can you: 

 * split these out into different R files
 * use the deprecation badge in the man file for `fit_control()`? See the svg's in `man/figures`

EmilHvitfeldt: Done!

EmilHvitfeldt:
I fixed the pkgdown issues by doing the same things we did when we deprecated `surv_reg()`
https://github.com/tidymodels/parsnip/pull/454

topepo (approved):

PR #629, opened 2021-12-21
Mode specific dependencies

### LINKED ISSUE #604: make model package dependencies mode-specific
parsnip registers required packages for a model/engine combination. The package list serves as a requirement for model fitting, prediction, and now deployment. 

The problem: since the package dependency is not mode dependent, we will have extra packages that are installed and loaded for modes that do not need them. For example, if the censored package is loaded, the censored and pec packages are required for classification and regression models. 

```r
library(tidymodels)
library(censored)

decision_tree() %>% 
  set_mode("regression") %>% 
  req_pkgs()
#> [1] "censored" "pec"      "rpart"
```

<sup>Created on 2021-11-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

This is also a problem when we are listing engine in the parsnip docs since there is no way to note that censored is required to use a particular mode.  

Note that this has not been an issue to date since engines that work for multiple modes have had the same package dependencies. censored is changing that with the new "censored regression" mode. 

An option can be added to `set_dependency()` for the mode the structure of the parsnip database for required packages would have a mode column. `get_model_spec()` should also be updated to make sure the correct packages are saved in the `method` slot when the model is fit. This would affect new models created with parsnip. 

What about existing fitted models? These should be fine since there is not really an issue yet (pre-censored). Model specifications should also be fine since the `methods` slot is not yet populated. 

`req_pkg()` (more specifically `get_pkgs()`) will also change to take into account the mode.

topepo:
Closes #604 

Non-breaking change; fixes issue with vetiver getting package dependencies across _all model modes_. 

Some changes will be made to parsnip extension packages at our leisure (but most importantly to censored).

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,8 @@
 
 * Argument `interval` was added for prediction: For types "survival" and "quantile", estimates for the confidence or prediction interval can be added if available (#615).
 
+* `set_dependency()` now allows developers to create package requirements that are specific to the model's mode (604). 
+
 
 # parsnip 0.1.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,7 +26,7 @@
 
 * Argument `interval` was added for prediction: For types "survival" and "quantile", estimates for the confidence or prediction interval can be added if available (#615).
 
-* `set_dependency()` now allows developers to create package requirements that are specific to the model's mode (604). 
+* `set_dependency()` now allows developers to create package requirements that are specific to the model's mode (#604). 
 
 
 # parsnip 0.1.7
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -676,7 +676,9 @@ set_dependency <- function(model, eng, pkg = "parsnip", mode = NULL) {
     dplyr::filter(engine == eng) %>%
     nrow()
   if (has_engine != 1) {
-    rlang::abort("The engine '{eng}' has not been registered for model '{model}'.")
+    rlang::abort(
+      glue::glue("The engine '{eng}' has not been registered for model '{model}'.")
+    )
   }
 
   # ----------------------------------------------------------------------------
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -535,7 +535,7 @@ set_new_model <- function(model) {
   set_env_val(model, dplyr::tibble(engine = character(0), mode = character(0)))
   set_env_val(
     paste0(model, "_pkgs"),
-    dplyr::tibble(engine = character(0), pkg = list())
+    dplyr::tibble(engine = character(0), pkg = list(), mode = character(0))
   )
   set_env_val(paste0(model, "_modes"), "unknown")
   set_env_val(
@@ -659,7 +659,7 @@ set_model_arg <- function(model, eng, parsnip, original, func, has_submodel) {
 #' @rdname set_new_model
 #' @keywords internal
 #' @export
-set_dependency <- function(model, eng, pkg = "parsnip") {
+set_dependency <- function(model, eng, pkg = "parsnip", mode = NULL) {
   check_model_exists(model)
   check_eng_val(eng)
   check_pkg_val(pkg)
@@ -668,6 +668,8 @@ set_dependency <- function(model, eng, pkg = "parsnip") {
   model_info <- get_from_env(model)
   pkg_info <- get_from_env(paste0(model, "_pkgs"))
 
+  # ----------------------------------------------------------------------------
+  # Check engine
   has_engine <-
     model_info %>%
     dplyr::distinct(engine) %>%
@@ -677,6 +679,23 @@ set_dependency <- function(model, eng, pkg = "parsnip") {
     rlang::abort("The engine '{eng}' has not been registered for model '{model}'.")
   }
 
+  # ----------------------------------------------------------------------------
+  # check mode; if missing assign all modes
+  all_modes <- unique(model_info$mode)
+  if (is.null(mode)) {
+    # For backward compatibility
+    mode <- all_modes
+  } else {
+    if (length(mode) > 1) {
+      rlang::abort("'mode' should be a single character value or NULL.")
+    }
+    if (!any(mode == all_modes)) {
+      rlang::abort(glue::glue("mode '{mode}' is not a valid mode for '{model}'"))
+    }
+  }
+
+  # ----------------------------------------------------------------------------
+
   existing_pkgs <-
     pkg_info %>%
     dplyr::filter(engine == eng)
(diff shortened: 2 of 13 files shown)


NEWS.md
```diff
@@ -26,6 +26,8 @@
 
 * Argument `interval` was added for prediction: For types "survival" and "quantile", estimates for the confidence or prediction interval can be added if available (#615).
 
+* `set_dependency()` now allows developers to create package requirements that are specific to the model's mode (604).
```
hfrick, line 29:
```suggestion
* `set_dependency()` now allows developers to create package requirements that are specific to the model's mode (#604). 
```

R/aaa_models.R
```diff
@@ -677,24 +679,45 @@ set_dependency <- function(model, eng, pkg = "parsnip") {
     rlang::abort("The engine '{eng}' has not been registered for model '{model}'.")
```
juliasilge, line 679: Is this supposed to have `glue::glue()` so we see what the engine and model are?

R/aaa_models.R
```diff
@@ -677,24 +679,45 @@ set_dependency <- function(model, eng, pkg = "parsnip") {
...
-    pkg_info <-
-      pkg_info %>%
-      dplyr::filter(engine != eng) %>%
-      dplyr::bind_rows(existing_pkgs)
+    if (length(mode) > 1) {
+      rlang::abort("'mode' should be a single character value or NULL.")
+    }
+    if (!any(mode == all_modes)) {
+      rlang::abort(glue::glue("mode '{mode}' is not a valid mode for '{model}'"))
+    }
```
juliasilge, line 694: I feel a little confused about this part; I can't get to these errors no matter what I try.

juliasilge (commented):
So what is the situation with `set_model_mode()` after this change? It looks to me like we might not need it at all, if the modes are being set per dependency? (I can get a new model to behave normally without it.)

Is this a situation where you can do it multiple ways? That sounds a bit confusing to me and we should at least only document one way to do it.

We should update [`/learn/develop/models`](https://www.tidymodels.org/learn/develop/models/) with the final changes here.

topepo:
We basically have a poor-man's relational database via environments.

`set_model_mode()` sets the equivalent of the mode table and declares what the possible modes are. 

`set_dependency()` updates the model-specific tables (and not the overall mode table). It also, via `check_model_exists()`, makes sure that the mode has been registered. 

Overall, the `set_*()` functions update a single table. I don't think that `set_dependency()` should update the mode table.

R/aaa_models.R
topepo: Yes, it should be "glued". I'll update the PR.

juliasilge:
Yes, but if we for example walk through this "create a new model" process but _skip_ using `set_model_mode(model = "discrim_mixture", mode = "classification")` and instead set the mode via `set_dependency()`, the whole rest of the tutorial works fine:
https://www.tidymodels.org/learn/develop/models/
Do we need to check that the mode has been set in the model table?

``` r
library(tidymodels)
set_new_model("discrim_mixture")

## skip this:
## set_model_mode(model = "discrim_mixture", mode = "classification")
set_model_engine(
  "discrim_mixture", 
  mode = "classification", 
  eng = "mda"
)
set_dependency("discrim_mixture", eng = "mda", pkg = "mda", mode = "classification")

set_model_arg(
  model = "discrim_mixture",
  eng = "mda",
  parsnip = "sub_classes",
  original = "subclasses",
  func = list(pkg = "foo", fun = "bar"),
  has_submodel = FALSE
)

discrim_mixture <-
  function(mode = "classification",  sub_classes = NULL) {
    # Check for correct mode
    if (mode  != "classification") {
      rlang::abort("`mode` should be 'classification'")
    }
    
    # Capture the arguments in quosures
    args <- list(sub_classes = rlang::enquo(sub_classes))
    
    # Save some empty slots for future parts of the specification
    new_model_spec(
      "discrim_mixture",
      args = args,
      eng_args = NULL,
      mode = mode,
      method = NULL,
      engine = NULL
    )
  }

set_fit(
  model = "discrim_mixture",
  eng = "mda",
  mode = "classification",
  value = list(
    interface = "formula",
    protect = c("formula", "data"),
    func = c(pkg = "mda", fun = "mda"),
    defaults = list()
  )
)

set_encoding(
  model = "discrim_mixture",
  eng = "mda",
  mode = "classification",
  options = list(
    predictor_indicators = "traditional",
    compute_intercept = TRUE,
    remove_intercept = TRUE,
    allow_sparse_x = FALSE
  )
)

data("two_class_dat", package = "modeldata")
set.seed(4622)
example_split <- initial_split(two_class_dat, prop = 0.99)
example_train <- training(example_split)
example_test  <-  testing(example_split)

mda_spec <- discrim_mixture(sub_classes = 2) %>% 
  set_engine("mda")

mda_fit <- mda_spec %>%
  fit(Class ~ ., data = example_train, engine = "mda")
mda_fit
#> parsnip model object
#> 
#> Call:
#> mda::mda(formula = Class ~ ., data = data, subclasses = ~2)
#> 
#> Dimension: 2 
#> 
#> Percent Between-Group Variance Explained:
#>     v1     v2 
#>  82.63 100.00 
#> 
#> Degrees of Freedom (per dimension): 3 
#> 
#> Training Misclassification Error: 0.17241 ( N = 783 )
#> 
#> Deviance: 671.391
```

<sup>Created on 2022-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/aaa_models.R
topepo: Here's a case that I'll add to the existing registration tests: 

``` r
library(parsnip)

set_new_model("sponge")
set_model_mode("sponge", "classification")
set_model_mode("sponge", "regression")
set_model_mode("sponge", "censored regression")

set_model_engine("sponge", mode = "classification", eng = "gum")
set_dependency("sponge",  "gum", "trident", mode = c("regression", "classification"))
#> Error: 'mode' should be a single character value or NULL.
```

<sup>Created on 2022-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

juliasilge: But this does _not_ error:

``` r
library(parsnip)

set_new_model("sponge")
set_model_mode("sponge", "regression")
set_model_mode("sponge", "censored regression")

set_model_engine("sponge", mode = "classification", eng = "gum")
```

<sup>Created on 2022-01-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

topepo: Add an issue and we'll fix that.

topepo:
That is a bug in `check_mode_val()`. I'll file an issue for that.

topepo:
Give it try but the mode should be properly checked when setting the engine and dependency.

juliasilge (approved):
This all looks good to go from my perspective now 🚀

PR #632, opened 2022-01-05
Extract parameter sets and single parameters

hfrick:
This PR adds a `extract_parameter_set_dials()` method and a `extract_parameter_dials()` method for model specs. It needs https://github.com/tidymodels/dials/pull/199 to be merged first.

`extract_parameter_set_dials.model_spec()` is to replace `parameters.model_spec()` because the name makes it clearer how the function is used. Deprecation of `parameters.model_spec()` is to happen in tune along with updates to the long-form documentation on tidymodels.org and in TMwR.

Left to do:
- deprecate `pull_dials_object()` generic and methods in dials + update TMwR
- deprecate the `parameters()` methods for recipes, model specs, and workflows in tune in favour of `extract_parameter_set_dials()` + update tidymodels.org and TMwR


hfrick:
I'm not sure why the test-coverage workflow fails. Should this be moved up to `v2` as well?

juliasilge (approved):
Ready to go in terms of the new code ⭐ 

I also am not sure why the test coverage is failing but probably the right thing to try is [updating it](https://github.com/r-lib/actions/blob/v2-branch/examples/test-coverage.yaml).

hfrick:
@juliasilge I've updated the test coverage job but it still failed. I don't know why the new tests cause problems (only) there but I have now set them to get skipped in that job (but not in the regular R CMD check job). Should we merge?

Issue #559, opened 2021-09-23
deprecate `varying_args()`


hfrick:
deprecate `varying()` and `varying_args()` which have been superseded by `tune()` and friends

EmilHvitfeldt:
`varying()` is currently being used in `update` tests throughout the package. Should I add a copy of `varying()` to the test helpers so we can keep the tests and avoid pulling anything from {tune}?

https://github.com/tidymodels/parsnip/blob/a8454b499eb9044bb388fc7d4e3cd4c2ec6da45c/tests/testthat/test_boost_tree.R#L108-L113

EmilHvitfeldt:
wait on https://github.com/tidymodels/hardhat/pull/181

topepo:
I would remove the use of `varying()`/`tune()` in parsnip examples

PR #636, opened 2022-01-19
Deprecate `varying()` and `varying_args()`

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/parsnip/issues/559.

The changes are

- reexport tune from hardhat. This change doesn't give conflicts with dev version of tune, which also reexports from hardhat.

``` r
library(parsnip)
library(tune)
```

<sup>Created on 2022-01-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

- lifecycle changes to start deprecation process
- remove usage of `varying()` and `varying_args()` internally in package code and tests
- update "getting started" vignette


hfrick:
should we make explicit that `tune()` is in hardhat and `tune_args()` in generics (and neither is in parsnip)?

EmilHvitfeldt:
If you are referring to the "Getting Started" page. Then I would vote to not mention it. {hardhat} at least is a very developer focused package that most users will never touch, and bringing it up in the introduction might send the wrong message

hfrick:
Totally agree about not mentioning it in the Getting Started page for the reasons you mentioned. Sorry I wasn't clear - I meant in the calls to `lifecycle::deprecate_soft()` like in the [lifecycle vignette](https://lifecycle.r-lib.org/articles/communicate.html#deprecate-a-function-1) and perhaps the NEWS. I'll leave a code suggestion for illustration.

R/varying.R
```diff
@@ -1,9 +1,13 @@
 #' A placeholder function for argument values
 #'
+#' @description
+#' `r lifecycle::badge("deprecated")`
+#'
 #' [varying()] is used when a parameter will be specified at a later date.
 #' @export
 #' @keywords internal
 varying <- function() {
+  lifecycle::deprecate_soft("0.1.8", "varying()", "tune()")
```
hfrick, line 10:
```suggestion
  lifecycle::deprecate_soft("0.1.8", "varying()", "hardhat::tune()")
```

topepo (changes requested):
Wherever you can link to tune, that would be good.

Issue #613, opened 2021-11-16
AttributeError: 'Sequential' object has no attribute 'predict_classes'


psads-git:
I am trying to reproduce the example 

Classification models using a neural network

offered at

https://www.tidymodels.org/learn/models/parsnip-nnet/

but getting the errors with the example below.

Thanks in advance!

```
library(keras)
library(tidymodels)
library(tensorflow)

data(bivariate)

ggplot(bivariate_train, aes(x = A, y = B, col = Class)) + 
  geom_point(alpha = .2)

biv_rec <- 
  recipe(Class ~ ., data = bivariate_train) %>%
  step_BoxCox(all_predictors())%>%
  step_normalize(all_predictors()) %>%
  prep(training = bivariate_train, retain = TRUE)

# We will bake(new_data = NULL) to get the processed training set back

# For validation:
val_normalized <- bake(biv_rec, new_data = bivariate_val, all_predictors())
# For testing when we arrive at a final model: 
test_normalized <- bake(biv_rec, new_data = bivariate_test, all_predictors())

set.seed(57974)
nnet_fit <-
  mlp(epochs = 100, hidden_units = 5, dropout = 0.1) %>%
  set_mode("classification") %>% 
  # Also set engine-specific `verbose` argument to prevent logging the results: 
  set_engine("keras", verbose = 0) %>%
  fit(Class ~ ., data = bake(biv_rec, new_data = NULL))

val_results <- 
  bivariate_val %>%
  bind_cols(
    predict(nnet_fit, new_data = val_normalized),
    predict(nnet_fit, new_data = val_normalized, type = "prob")
  )
val_results %>% slice(1:5)

Error in py_get_attr_impl(x, name, silent) : 
  AttributeError: 'Sequential' object has no attribute 'predict_classes'
In addition: Warning message:
In keras::predict_classes(object = object$fit, x = as.matrix(new_data)) :
  `predict_classes()` is deprecated and and was removed from tensorflow in version 2.6.
Please update your code:
  * If your model does multi-class classification:
    (e.g. if it uses a `softmax` last-layer activation).

      model %>% predict(x) %>% k_argmax()

  * if your model does binary classification
    (e.g. if it uses a `sigmoid` last-layer activation).

      model %>% predict(x) %>% `>`(0.5) %>% k_cast("int32")
```

Issue #596, opened 2021-11-10
Update keras code


topepo:
We are getting these warnings: 

```
── Warning (test_logistic_reg_keras.R:186:3): classification probabilities ─────
`predict_proba()` is deprecated and was removed from tensorflow in version 2.6, please use `predict()` instead
Backtrace:
 1. stats::predict(plrfit, te_dat[, -1], type = "prob") test_logistic_reg_keras.R:186:2
 6. keras::predict_proba(object = object$fit, x = as.matrix(new_data))

-- Warning (test_multinom_reg_keras.R:124:3): classification prediction --------
`predict_classes()` is deprecated and and was removed from tensorflow in version 2.6.
Please update your code:
  * If your model does multi-class classification:
    (e.g. if it uses a `softmax` last-layer activation).

      model %>% predict(x) %>% k_argmax()

  * if your model does binary classification
    (e.g. if it uses a `sigmoid` last-layer activation).

      model %>% predict(x) %>% `>`(0.5) %>% k_cast("int32")
```

juliasilge:
Looks like these are removed in TensorFlow 2.6, as reported in tidymodels/tidymodels#71

nickryd:
Hello, Im wondering if there was any resolution of this issue. Currently Im getting an error when running models with Keras and not just a warning. Looks like this issue is tagged to be resolved with the next release, just curious if you could provide the solution so I could get this running in the meantime. Thanks!

juliasilge:
In the meantime @nickryd you would need to install back to an earlier version of Tensorflow (the Python software, not the R package), pre-2.6.

topepo:
@nickryd I'm working on it this week.

siegfried:
This also breaks `tune_grid` and `tune_bayes`. Here is what I grab from `.notes`.

```
 [1] "preprocessor 1/4, model 1/2 (predictions): `predict_proba()` is deprecated and was removed from tensorflow in version 2.6, please use `predict()` instead"  
 [2] "preprocessor 1/4, model 1/2 (predictions): Error in py_get_attr_impl(x, name, silent): AttributeError: 'Sequential' object has no attribute 'predict_proba'"
```

PR #639, opened 2022-01-26
Update internal functions to reflect deprecated keras functions

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/parsnip/issues/613 and close https://github.com/tidymodels/parsnip/issues/596.

With the changes to tensorflow, internal functions have been updated to reflect deprecated functions.


R/mlp.R
```diff
@@ -438,5 +438,29 @@ reformat_torch_num <- function(results, object) {
   results
 }
 
+#' Wrapper for keras class predictions
+#' @param object A keras model fit
+#' @param x A data set.
+#' @export
+#' @keywords internal
+keras_predict_classes <- function(object, x)  {
```
topepo, line 446: Is the reason that we are returning integers related to availability of the factor levels?

tests/testthat/test_linear_reg_keras.R
```diff
@@ -29,6 +29,11 @@ test_that('model fitting', {
   skip_if_not_installed("keras")
 
   set.seed(257)
+  if (tensorflow::tf_version() >= package_version("2.0")) {
```
topepo, line 32: Should we write a wrapper for this (`set_tf_seed()`) and export it?

topepo (changes requested):
Also, please make an entry in the news file about this.

R/mlp.R
EmilHvitfeldt: I assume that this was initially done because that was what {keras} did. We are converting those integers back to factors, in the `post` part of prediction
https://github.com/tidymodels/parsnip/blob/0c7f44080b57c7b79e15e35c932eb1a1d2745508/R/mlp_data.R#L149-L151

tests/testthat/test_linear_reg_keras.R
EmilHvitfeldt: Yes, let us do that

R/mlp.R
topepo: Can we do that in this function? It would be more clean.

tests/testthat/test_mlp_keras.R
```diff
@@ -74,7 +74,7 @@ test_that('keras classification prediction', {
     control = ctrl
   )
 
-  xy_pred <- keras::predict_classes(xy_fit$fit, x = as.matrix(hpc[1:8, num_pred]))
+  xy_pred <- predict(xy_fit$fit, x = as.matrix(hpc[1:8, num_pred])) %>% keras::k_argmax() %>% as.integer()
```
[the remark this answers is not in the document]
EmilHvitfeldt: I'm gonna swap between the prediction styles conditionally on the tensorflow version such that the "old tensorflow" GHA can check those tests. Otherwise we didn't need that GHA

R/mlp.R
EmilHvitfeldt: Done!

tests/testthat/test_linear_reg_keras.R
EmilHvitfeldt: Done!

tests/testthat/test_mlp_keras.R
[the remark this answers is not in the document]
EmilHvitfeldt: Done!

DESCRIPTION
```diff
@@ -84,6 +85,6 @@ Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
 RoxygenNote: 7.1.2
-Remotes:  
+Remotes:
```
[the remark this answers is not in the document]
EmilHvitfeldt: Removed!

R/mlp.R
```diff
@@ -436,5 +436,53 @@ reformat_torch_num <- function(results, object) {
...
+#' @export
+#' @keywords internal
+keras_predict_classes <- function(object, x) {
+  if (utils::packageVersion("keras") >= package_version("2.6")) {
+    preds <- predict(object$fit, x)
+    if (tensorflow::tf_version() <= package_version("2.0.0")) {
+      # -1 to assign with keras' zero indexing
+      index <- apply(preds, 1, which.max) - 1
+    } else {
+      index <- preds %>% keras::k_argmax() %>% as.integer()
```
t-kalinowski, line 451: If performance becomes an issue, you may want to avoid using k_argmax() here because it introduces another somewhat expensive full conversation from R -> python -> R

R/mlp.R
```diff
@@ -436,5 +436,53 @@ reformat_torch_num <- function(results, object) {
...
+  object$lvl[index + 1]
+}
 
+#' Wrapper for keras class probability predictions
+#' @param object A keras model fit
+#' @param x A data set.
+#' @export
+#' @keywords internal
+keras_predict_proba <- function(object, x) {
+  if (utils::packageVersion("keras") >= package_version("2.6")) {
```
t-kalinowski, line 465: You probably want to test the tensorflow version here `tf_version()`, not the R package version

R/mlp.R
EmilHvitfeldt: Okay, we could write our own `k_argmax()` function if we run into complaints

R/mlp.R
EmilHvitfeldt: Are you saying to test the tensorflow version instead of the keras version or in addition?

t-kalinowski: Can you use `predict` directly in all versions?

EmilHvitfeldt: doesn't work for all versions :(

I get the following error if I try to switch to `predict()` only for `utils::packageVersion("keras") == "2.4.0"`

```
<Rcpp::exception/C++Error/error/condition>
Error in `py_call_impl(callable, dots$args, dots$keywords)`: AttributeError: 'Sequential' object has no attribute 'shape'
```

EmilHvitfeldt:
This PR tries to handle a couple of different problems.

The first problem is the error emitted by Keras as mentioned in #613 and #596, where `predict_classes` and `predict_proba` were changed. There is now a conditional (version(keras) >= 2.6) on the package version of `keras` to switch between using `predict()` and `predict_proba()`

Secondly the use of `keras::k_argmax()` is conditioned to be used when the tensorflow version is v2.

Lastly, `set_tf_seed()` has been added, exported and used in tests. This function sets seed in R and tensorflow at the same time, conditionally switching between `tensorflow::tf$random$set_seed` and `tensorflow::tf$random$set_random_seed` for v1 and v2 respectively.

I have tested these changes locally using all combinations that could be installed together of

- Keras: 2.7.0, 2.6.1, 2.6.0, 2.4.0
- Tensorflow: 2.7.1, 2.6.3, 1.15.0

R/mlp.R
t-kalinowski: Do you maybe have a reprex you can share?

R/mlp.R
topepo: Would something like `apply(x, 1, which.max)` work?

R/mlp.R
EmilHvitfeldt: My mistake. I updated in https://github.com/tidymodels/parsnip/pull/639/commits/12ce5b85d4a10e3bdda810a1d96fdcb5eb0d11f0

R/mlp.R
t-kalinowski: also, for matrices (`length(dim(x)) == 2`) you can use `max.col()`

PR #643, opened 2022-02-02
glmnet autoplot method for #642

### LINKED ISSUE #642: proper glmnet autoplot method
Our training materials have had a function in them for a long time. We should move it here and take a ggrepl dependency.

topepo:
Closes #642 

Tests will be in extratests.

Example:

``` r
library(tidymodels)
#> Warning: package 'broom' was built under R version 4.1.2
tidymodels_prefer()
theme_set(theme_bw())

data(meats)
meats <- meats %>% select(-protein, -fat)

glmn_spec <- 
 linear_reg(penalty = 0.1) %>% 
  set_engine("glmnet")
rec <- 
  recipe(water ~ ., data = meats) %>% 
  step_normalize(all_predictors())

glmn_fit <- 
  workflow(rec, glmn_spec) %>% 
  fit(meats)
  
autoplot(glmn_fit)
```

``` r
autoplot(glmn_fit, min_penalty = 0.01, top_n = 2)
```

<sup>Created on 2022-02-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


EmilHvitfeldt:
I know it would require a bit more work. But what about showing the labels on the dotted line when `best_penalty` is specified? Something like this?

Screen Shot 2022-02-01 at 8 28 12 PM

DESCRIPTION
```diff
@@ -20,6 +20,8 @@ Depends:
 Imports: 
     dplyr (>= 0.8.0.1),
     generics (>= 0.1.0.9000),
+    ggplot2,
+    ggrepel,
```
DavisVaughan, line 24: I think these could be in Suggests, then use `rlang::check_installed()` to ensure they are there

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+#' @param top_n A non-negative integer for how many model predictors to label.
+#' The top predictors are ranked by their absolute coefficient value. For
+#' multinomial or multivariate models, the `top_n` terms are selected within
+#' class or response, respectively.
+#' @param ... For [autoplot.glmnet()], options to pass to
+#' [ggrepel::geom_label_repel()]. Otherwise, this argument is ignored.
+#' @return A ggplot object with penalty on the x-axis and coefficients on the
+#' y-axis. For multinomial or multivariate models, the plot is faceted.
+#'
+#' @export
```
DavisVaughan, line 21: I think rather than doing `@export`, you should add a comment that this is lazily registered in the .onLoad

You already do this with `s3_register()` in the `zzz.R` file. You do not need to both lazily register like that and `@export` here

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+autoplot.model_fit <- function(object, ...) {
+  autoplot(object$fit, ...)
+}
+
+#' @export
+#' @rdname autoplot.model_fit
+autoplot.workflow <- function(object, ...) {
+  object %>% extract_fit_engine %>% autoplot(...)
+}
```
DavisVaughan, line 30: I think that workflows should be the one who exports an `autoplot()` method for a workflow. It seems like a reasonable thing to add to workflows after this version of parsnip makes it to CRAN. Could you please open an issue in workflows if you agree?

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+
+# glmnet is not a formal dpendency here.
+# unit tests are located at https://github.com/tidymodels/extratests
+
+# nocov start
+
+#' @export
+#' @rdname autoplot.model_fit
```
DavisVaughan, line 38: Again, mention that you are doing lazy registration and remove the `@export`

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+autoplot.glmnet <- function(object, min_penalty = 0, best_penalty = NULL,
+                            top_n = 3L, ...) {
```
DavisVaughan, line 40: I would prefer the optional arguments to come after the `...` to force them to be named.

If someone calls `autoplot(obj, 2)` then I have no idea that `2` maps to `min_penalty`.

This also allows us to freely add more options in any order in the future

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+    ggplot2::geom_line(alpha = .4, show.legend = FALSE) +
+    ggplot2::scale_x_log10()
+
+  if(top_n > 0) {
+    p <- p +
+      ggrepel::geom_label_repel(
+        data = label_coefs,
+        ggplot2::aes(y = estimate, label = label),
+        show.legend = FALSE,
+        ...
```
DavisVaughan, line 146: I am slightly uncomfortable with the idea that dots are passed all the way through to here. I would be more comfortable with an explicit argument for `repel_opts = named list()` that gets passed through to here instead.

Then the dots of this `autoplot()` method would either be ignored or you would call `rlang::check_dots_empty()` to ensure the user didn't have any typos

_pkgdown.yml
```diff
@@ -48,6 +48,7 @@ reference:
       - svm_rbf
   - title: Infrastructure
     contents:
+      - autoplot.model_fit
```
DavisVaughan, line 51: If you lazily register the method (which I think you should), then this will not be required

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+}
+
+
+map_glmnet_coefs <- function(x) {
+  coefs <- coef(x)
+  p <- x$dim[1]
+  if (is.list(coefs)) {
+    classes <- names(coefs)
+    coefs <- purrr::map(coefs, reformat_coefs, p = p, penalty = x$lambda)
+    coefs <- purrr::map2_dfr(coefs, classes, ~ .x %>% dplyr::mutate(class = .y))
```
DavisVaughan, line 51:
```suggestion
    coefs <- purrr::map2_dfr(coefs, classes, ~ dplyr::mutate(.x, class = .y))
```

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+  num_estimates <- nrow(x)
+  if (num_estimates > p) {
+    # The intercept is first
+    x <- x[-(num_estimates - p),, drop = FALSE]
+  }
+  term_lab <- rownames(x)
+  colnames(x) <- paste(seq_along(penalty))
+  x <- tibble::as_tibble(x)
+  x$term <- term_lab
+  x <- tidyr::pivot_longer(x, cols = c(-term), names_to = "index", values_to = "estimate")
```
DavisVaughan, line 69:
```suggestion
  x <- tidyr::pivot_longer(x, cols = -term, names_to = "index", values_to = "estimate")
```

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+    dplyr::slice(1) %>%
+    dplyr::ungroup() %>%
+    dplyr::arrange(dplyr::desc(abs(estimate))) %>%
+    dplyr::slice(1:top_n)
+}
+
+autoplot_glmnet <- function(x, min_penalty = 0, best_penalty = NULL, top_n = 3L, ...) {
+  if (!is.null(min_penalty)) {
+    check_penalty_value(min_penalty)
+  }
```
DavisVaughan, line 88: What does a `NULL` `min_penalty` mean? The default is `0`?

R/autoplot.R
```diff
@@ -0,0 +1,162 @@
...
+    dplyr::filter(penalty >= min_penalty)
+
+  actual_min_penalty <- min(tidy_coefs$penalty)
+  num_terms <- length(unique(tidy_coefs$term))
+  top_n <- min(top_n[1], num_terms)
+  if (top_n < 0) {
+    top_n <- 0
+  }
+
+  has_groups <- any(names(tidy_coefs) %in% c("class"))
```
DavisVaughan, line 101:
```suggestion
  has_groups <- any(names(tidy_coefs) == "class")
```

EmilHvitfeldt:
Added a commit to have labels follow `best_penalty` line

``` r
library(tidymodels)
tidymodels_prefer()
theme_set(theme_bw())

data(meats)
meats <- meats %>% select(-protein, -fat)

glmn_spec <- 
  linear_reg(penalty = 0.1) %>% 
  set_engine("glmnet")
rec <- 
  recipe(water ~ ., data = meats) %>% 
  step_normalize(all_predictors())

glmn_fit <- 
  workflow(rec, glmn_spec) %>% 
  fit(meats)

autoplot(glmn_fit)
```



``` r
autoplot(glmn_fit, best_penalty = 0.033)
```



``` r
autoplot(glmn_fit, best_penalty = 0.6)
```



<sup>Created on 2022-02-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/autoplot.R
topepo: We'll also need to import and re-export `autoplot()`. Right now it only works when one of the other packages reexports it.

R/autoplot.R
topepo: FYI I thought we would need to check for multiple columns to work with multivariate models.

R/autoplot.R
topepo: Why is it uncomfortable? It seems like the most standard application of `...`

juliasilge:
I fixed the broken GH actions in a previous PR if you want to merge from main.

Issue #655, opened 2022-02-10
In xgboost, `ntreelimit` has been deprecated in favour of `iteration_range`


tiagomaie:
https://github.com/tidymodels/parsnip/blob/0af9c8617b7961dbf74b5a29578374e97ffa143a/R/boost_tree.R#L484-L486

L485 should instead be:
`pred <- xgb_pred(object$fit, newdata = new_data,  iterationrange = c(1, tree + 1), ntreelimit = NULL)`

`ntreelimit` has been deprecated.
It is `tree + 1` as in the xgboost [docs](https://cran.r-project.org/web/packages/xgboost/xgboost.pdf#Rfn.predict.xgb.Booster.1) they mention that `iteration_range` should be a half-open set.
This should fix the warnings given when running `tune::tune_grid` for example, see below.

``` r
library(parsnip)
library(dplyr)

# using iris dataset

# defining train/test split
train_ids <- 1:(nrow(iris) - 50)
test_ids <- (nrow(iris) - 49):nrow(iris)

# defining recipe
d_recipe <- recipes::recipe(
  Species ~ .,
  data = iris[train_ids, ] %>% dplyr::mutate(Species = factor(Species))
)

# defining model
g_model <-
  parsnip::boost_tree(trees = tune::tune()) %>%
  parsnip::set_engine("xgboost", objective = "binary:logistic") %>%
  parsnip::set_mode("classification")


# defining workflow
wf <- workflows::workflow() %>%
  workflows::add_recipe(d_recipe) %>%
  workflows::add_model(g_model)

# tuning params
res <- wf %>% tune::tune_grid(
    resamples=rsample::vfold_cv(
      recipes::bake(d_recipe %>% recipes::prep(), new_data = NULL)
    )
  )
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip
#> [18:38:53] WARNING: amalgamation/../src/c_api/c_api.cc:718: `ntree_limit` is deprecated, use `iteration_range` instead.
#> [18:38:53] WARNING: amalgamation/../src/c_api/c_api.cc:718: `ntree_limit` is deprecated, use `iteration_range` instead.
#> [18:38:53] WARNING: amalgamation/../src/c_api/c_api.cc:718: `ntree_limit` is deprecated, use `iteration_range` instead.
...
[...]

PR #656, opened 2022-02-10
`ntreelimit` has been `deprecated` in favour of iteration_range

tiagomaie:
This is a fix for issue #655 

As `ntreelimit` as been deprecated, adding `iterationrange` to the
`xgb_pred` call with the correct modifications to fit the current use case.

Further work is needed at a later stage to extend the new functionality of `iterationrange`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,7 +42,7 @@
 
 * parsnip is now more robust working with keras and tensorflow for a larger range of versions (#596).
 
-* xgboost engines now use the new `iteration_range` parameter instead of the deprecated `ntreelimit` (#656).  
+* xgboost engines now use the new `iterationrange` parameter instead of the deprecated `ntreelimit` (#656).  
 
 # parsnip 0.1.7
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -42,6 +42,8 @@
 
 * parsnip is now more robust working with keras and tensorflow for a larger range of versions (#596).
 
+* xgboost engines now use the new `iteration_range` parameter instead of the deprecated `ntreelimit` (#656).  
+
 # parsnip 0.1.7
 
 ## Model Specification Changes
--- a/R/boost_tree.R
+++ b/R/boost_tree.R
@@ -482,7 +482,7 @@ multi_predict._xgb.Booster <-
   }
 
 xgb_by_tree <- function(tree, object, new_data, type, ...) {
-  pred <- xgb_pred(object$fit, newdata = new_data, ntreelimit = tree)
+  pred <- xgb_pred(object$fit, newdata = new_data, iterationrange = c(1,tree+1), ntreelimit = NULL)
 
   # switch based on prediction type
   if (object$spec$mode == "regression") {
--- a/R/boost_tree.R
+++ b/R/boost_tree.R
@@ -482,7 +482,12 @@ multi_predict._xgb.Booster <-
   }
 
 xgb_by_tree <- function(tree, object, new_data, type, ...) {
-  pred <- xgb_pred(object$fit, newdata = new_data, iterationrange = c(1,tree+1), ntreelimit = NULL)
+  pred <- xgb_pred(
+    object$fit,
+    newdata = new_data,
+    iterationrange = c(1, tree + 1),
+    ntreelimit = NULL
+  )
 
   # switch based on prediction type
   if (object$spec$mode == "regression") {
--- a/R/boost_tree.R
+++ b/R/boost_tree.R
@@ -485,7 +485,7 @@ xgb_by_tree <- function(tree, object, new_data, type, ...) {
   pred <- xgb_pred(
     object$fit,
     newdata = new_data,
-    iterationrange = c(1, tree + 1),
+    iteration_range = c(1, tree + 1),
     ntreelimit = NULL
   )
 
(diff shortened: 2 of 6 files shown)


juliasilge:
Do we want to make this change now and set `xgboost (>= 1.5.0.1)` in DESCRIPTION? Or try to handle both old/new versions somehow?

As a reminder, xgboost is only in Suggests for parsnip but how would this impact our ability to build on older versions of R? We are building back to `oldrel-2` right now.

topepo:
xgboost uses R (≥ 3.3.0) so it probably shouldn't been an issue for CI. 

I"m inclined to add a version restriction in parsnip. The xgboost api does change a lot and programming around this (and other changes) would be messy.

juliasilge:
@tiagomaie I just want to confirm with you that I changed the parameter to `iteration_range` based on the [NEWS](https://github.com/dmlc/xgboost/blob/master/NEWS.md#v150-2021-oct-11) and [docs](https://xgboost.readthedocs.io/en/stable/prediction.html#prediction-options) (from what you submitted in your PR here, `iterationrange`.

tiagomaie:
Thanks @juliasilge ! I should mention that although in some warnings and in the docs it says that `iteration_range` should be used, I think these stem from implementations in other languages, like python.
On my end (and also looking through xgboost R code base and documentation, see [predict.xgb.Booster](https://github.com/dmlc/xgboost/blob/7f738e7f6fd0fa5b11db38fd0717b15426af53a9/R-package/R/xgb.Booster.R#L329) for example) it seems that the correct parameter and the one that produces the correct behavior is `iterationrange`.

Running parsnips test for xgboost (ie. `devtools::test(filter='boost_tree_xgboost')`), fail to produce the correct result if `iteration_range` is used instead of `iterationrange` in `xgb_pred`. From my understanding, this is because [predict.xgb.Booster](https://github.com/dmlc/xgboost/blob/7f738e7f6fd0fa5b11db38fd0717b15426af53a9/R-package/R/xgb.Booster.R#L327) won't take it as a parameter and instead will just assign the `ntreelimit` with the value of 0.

NEWS.md
```diff
@@ -42,6 +42,8 @@
 
 * parsnip is now more robust working with keras and tensorflow for a larger range of versions (#596).
 
+* xgboost engines now use the new `iteration_range` parameter instead of the deprecated `ntreelimit` (#656).
```
tiagomaie, line 45:
```suggestion
* xgboost engines now use the new `iterationrange` parameter instead of the deprecated `ntreelimit` (#656).  
```

R/boost_tree.R
```diff
@@ -482,7 +482,12 @@ multi_predict._xgb.Booster <-
   }
 
 xgb_by_tree <- function(tree, object, new_data, type, ...) {
-  pred <- xgb_pred(object$fit, newdata = new_data, ntreelimit = tree)
+  pred <- xgb_pred(
+    object$fit,
+    newdata = new_data,
+    iteration_range = c(1, tree + 1),
```
tiagomaie, line 488:
```suggestion
    iterationrange = c(1, tree + 1),
```

tests/testthat/test_boost_tree_xgboost.R
```diff
@@ -191,7 +191,7 @@ test_that('submodel prediction', {
 
   x <-  xgboost::xgb.DMatrix(as.matrix(mtcars[1:4, -1]))
 
-  pruned_pred <- predict(reg_fit$fit, x, ntreelimit = 5)
+  pruned_pred <- predict(reg_fit$fit, x, iteration_range = 5)
```
tiagomaie, line 194:
```suggestion
  pruned_pred <- predict(reg_fit$fit, x, iterationrange = c(1, 6))
```

tests/testthat/test_boost_tree_xgboost.R
```diff
@@ -206,7 +206,7 @@ test_that('submodel prediction', {
 
   x <-  xgboost::xgb.DMatrix(as.matrix(wa_churn[1:4, vars]))
 
-  pred_class <- predict(class_fit$fit, x, ntreelimit = 5)
+  pred_class <- predict(class_fit$fit, x, iteration_range = 5)
```
tiagomaie, line 209:
```suggestion
  pred_class <- predict(class_fit$fit, x, iterationrange = c(1, 6))
```

tests/testthat/test_boost_tree_xgboost.R
tiagomaie: `iterationrange` should be a numeric vector of size 2 with the iterations of the trees to use.
To have the same behavior as `ntreelimit = 5` (use all iterations until and including the fifth) we need to do `iterationrange = c(1, 6)` (use iterations 1 through 6, but do not include 6, as it is a half open set).

See [iterationrange in predict.xgb.Booster](https://cran.r-project.org/web/packages/xgboost/xgboost.pdf#Rfn.predict.xgb.Booster.1).

tests/testthat/test_boost_tree_xgboost.R
tiagomaie: `iterationrange` should be a numeric vector of size 2 with the iterations of the trees to use.
To have the same behavior as `ntreelimit = 5` (use all iterations until and including the fifth) we need to do `iterationrange = c(1, 6)` (use iterations 1 through 6, but do not include 6, as it is a half open set).

See [iterationrange in predict.xgb.Booster](https://cran.r-project.org/web/packages/xgboost/xgboost.pdf#Rfn.predict.xgb.Booster.1).

juliasilge:
Ah, thank you very much for clarifying @tiagomaie! Very helpful.

juliasilge (approved):

PR #659, opened 2022-02-18
Add release bullet about `knit_engine_docs()`

### LINKED ISSUE #645: Add release question about `knit_engine_docs()`
We can add [customized questions to be added to a release issue](https://github.com/r-lib/devtools/issues/1754#issuecomment-379896636) created via usethis. Let's add one to remind us to `knit_engine_docs()` before each release.

juliasilge:
Closes #645 

This will add a bullet [when we do `usethis::use_release_issue()`](https://usethis.r-lib.org/reference/use_release_issue.html).

### DIFF
--- a/R/aaa.R
+++ b/R/aaa.R
@@ -96,4 +96,8 @@ utils::globalVariables(
    )
 )
 
+release_bullets <- function() {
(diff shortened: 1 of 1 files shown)


topepo:
If you need to update something, please add `update_model_info_file()` too.

PR #660, opened 2022-02-18
Add links between `logistic_reg()` and `multinom_reg()`

### LINKED ISSUE #579: Linking between logistic_reg() and multinom_reg()
Would it be worth it to link between `logistic_reg()` and `multinom_reg()` in their documentation?
I have a feeling that some users will end up on `logistic_reg()` but actually need `multinom_reg()`.

juliasilge:
Closes #579

### DIFF
--- a/R/logistic_reg.R
+++ b/R/logistic_reg.R
@@ -28,6 +28,9 @@
 #'
 #' @template spec-details
 #'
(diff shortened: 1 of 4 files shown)


EmilHvitfeldt (approved):
LGTM 🙌

PR #663, opened 2022-02-19
Refactor `make_engine_list()` to refine docs

### LINKED ISSUE #649: Replace "may require" wording
This code:

https://github.com/tidymodels/parsnip/blob/1439173f264345da28ace5044a91586210c01b77/R/engine_docs.R#L231

generates a sentence saying someone "may require" an extension package, which is a bit perplexing. We need to know the _mode_ to really say one way or the other (for example, `decision_tree()` requires an extension package for the `"rpart"` engine only for censored regression.

It looks to me like we have that info and can avoid saying "may require":

```
library(tidyverse)
utils::read.delim(system.file("models.tsv", package = "parsnip")) %>% 
  filter(model == "decision_tree")
#>           model                mode engine      pkg
#> 1 decision_tree censored regression  party censored
#> 2 decision_tree censored regression  rpart censored
#> 3 decision_tree      classification   C5.0     <NA>
#> 4 decision_tree      classification  rpart     <NA>
#> 5 decision_tree      classification  spark     <NA>
#> 6 decision_tree          regression  rpart     <NA>
#> 7 decision_tree          regression  spark     <NA>
```

juliasilge:
Closes #649 

After reworking this function a bit, there is more clarity in the note on the engine list.

For `linear_reg()`:

> Requires a parsnip extension package.

For `decision_tree()`:

> Requires a parsnip extension package for censored regression.

For `bag_tree()`:

> Requires a parsnip extension package for censored regression, classification, and regression.

### DIFF
--- a/R/engine_docs.R
+++ b/R/engine_docs.R
@@ -234,7 +234,8 @@ make_engine_list <- function(mod) {
         dplyr::filter(has_ext != "") %>%
         dplyr::pull(mode) %>%
         unique() %>%
-        knitr::combine_words(and = " and ")
+        sort() %>%
+        knitr::combine_words()
       notes <- paste0(
         notes, " ",
         cli::symbol$sup_2, " Requires a parsnip extension package for ",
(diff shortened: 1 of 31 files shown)


topepo:
Very nice. I also tested by adding the extra engines that treesnip will eventually provide.

topepo (approved):

PR #664, opened 2022-02-20
Re-registration of model information

### LINKED ISSUE #653: allow re-registration of same model information
To avoid issues running `devtools::test()` and/or `load_all()`. 

The idea is to allow extension packages to (silently) overwrite the model registry environments  _if the same data is being registered_. 

If other data is trying to overwrite, we should issue and error saying

 - If you are working on a different extension package, we don't allow this
 -  Otherwise, do a hard restart of the R session (i.e. don't `load_all()`)

topepo:
Closes #653 

Models components (e.g. fit, pred, etc) can now be re-registered as long as the information is the same. 

One caveat:  package dependencies cannot be check for the same values since those are set one value at a time.

### DIFF
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -674,12 +674,12 @@ set_dependency <- function(model, eng, pkg = "parsnip", mode = NULL) {
   check_eng_val(eng)
   check_pkg_val(pkg)
 
-  current <- get_model_env()
   model_info <- get_from_env(model)
   pkg_info <- get_from_env(paste0(model, "_pkgs"))
 
   # ----------------------------------------------------------------------------
   # Check engine
+
   has_engine <-
     model_info %>%
     dplyr::distinct(engine) %>%
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -540,7 +540,7 @@ set_new_model <- function(model) {
 
   current <- get_model_env()
 
-  set_env_val("models", c(current$models, model))
+  set_env_val("models", unique(c(current$models, model)))
   set_env_val(model, dplyr::tibble(engine = character(0), mode = character(0)))
   set_env_val(
     paste0(model, "_pkgs"),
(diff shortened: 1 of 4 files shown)


hfrick (approved):
no problems with censored or baguette - so nice to be able to use test() again 😀

R/aaa_models.R
```diff
@@ -750,37 +750,102 @@ get_dependency <- function(model) {
...
+  y <- unlist(y)
+  x_func <- purrr::map_lgl(x, is.function)
+  y_func <- purrr::map_lgl(y, is.function)
+  funcs <- unique(c(names(x)[x_func], names(y)[y_func]))
+  if (length(funcs) > 0) {
+    x[funcs] <- purrr::map(x[funcs], bland_func)
+    y[funcs] <- purrr::map(y[funcs], bland_func)
+  }
+  isTRUE(all.equal(x, y))
+}
```
DavisVaughan, line 818: I don't think this actually solves the `identical()` problem, because you switched to `all.equal()`. If you put `identical()` back in then it would still fail. I've determined this is a bug with `identical()`.

I think you can just use `all.equal()` on the tibbles directly?

topepo: No, `all.equal()` does not work. 

With this, we can load censored then `load_all()` again without errors. It's not pretty but it does work.

R/aaa_models.R
```diff
@@ -750,37 +750,102 @@ get_dependency <- function(model) {
...
+  new_encoding <- is.null(current) & component == "encoding"
+
+  if (new_encoding) {
+    return("new")
+  } else {
+    current <-  dplyr::filter(current, engine == eng & mode == !!mode)
+  }
+
+  if (component == "predict" & !is.null(pred_type)) {
+
```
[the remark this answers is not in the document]
topepo: please expand your comment

R/aaa_models.R
DavisVaughan: oh it seems like it was `all.equal(x, y, check.environment = FALSE)` that works

Issue #671, opened 2022-03-09
Integration with `sparklyr` is failing, starting with `parsnip` 2.0


edgararuiz:
Recent `sparklyr` GHA R checks are failing on `parsnip` tests.  These tests are new, and dev-version package testing is not yet working in GHA for `sparklyr`, so we didn't catch it until now.

reprex:
```r
library(sparklyr)
library(parsnip)

sc <- spark_connect("local")
iris_tbl <- sdf_copy_to(sc, iris)

decision_tree(engine = "spark") %>%
  set_mode("classification") %>%
  fit(Species ~ Sepal_Length + Petal_Length, iris_tbl)
#> Error in eval_tidy(env$formula[[2]], env$data): object 'Species' not found

spark_disconnect(sc)
```

It looks like this is the check in which it happens: https://github.com/tidymodels/parsnip/blob/b6db6763c82e4307978bdb54ed15d77607461a1f/R/fit_helpers.R#L9

Also confirmed here: https://github.com/sparklyr/sparklyr/runs/5483357141?check_suite_focus=true#step:19:1469

topepo:
I think that this would be fixed if we wrapped it in a 

```r
if (!inherits(env$data, "tbl_spark")) {
  check_outcome(eval_tidy(env$formula[[2]], env$data), object) 
}
```

PR #677, opened 2022-03-12
don't use check_outcome() when using sparklyr

EmilHvitfeldt:
To close https://github.com/tidymodels/parsnip/issues/671

``` r
library(sparklyr)
#> 
#> Attaching package: 'sparklyr'
#> The following object is masked from 'package:stats':
#> 
#>     filter
library(parsnip)

sc <- spark_connect("local")
iris_tbl <- sdf_copy_to(sc, iris)

decision_tree(engine = "spark") %>%
  fit(Species ~ Sepal_Length + Petal_Length, iris_tbl)
#> parsnip model object
#> 
#> Formula: Species ~ Sepal_Length + Petal_Length
#> 
#> DecisionTreeClassificationModel (uid=decision_tree_classifier__6a7e32a5_0707_461f_bfe5_d825c6b6f9f1) of depth 5 with 17 nodes

spark_disconnect(sc)
```

<sup>Created on 2022-03-11 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Should I add a corresponding test in https://github.com/tidymodels/extratests?

### DIFF
--- a/R/fit_helpers.R
+++ b/R/fit_helpers.R
@@ -6,7 +6,7 @@
 form_form <-
   function(object, control, env, ...) {
 
-    if (!inherits(env$data, "tbl_spark")) {
+    if (inherits(env$data, "data.frame")) {
       check_outcome(eval_tidy(env$formula[[2]], env$data), object)
     }
 
--- a/R/fit_helpers.R
+++ b/R/fit_helpers.R
@@ -6,7 +6,9 @@
 form_form <-
   function(object, control, env, ...) {
 
-    check_outcome(eval_tidy(env$formula[[2]], env$data), object)
+    if (!inherits(env$data, "tbl_spark")) {
+      check_outcome(eval_tidy(env$formula[[2]], env$data), object)
+    }
 
     # prob rewrite this as simple subset/levels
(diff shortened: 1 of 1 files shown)


juliasilge:
Unless I am misunderstanding, won't the existing Spark tests cover this?

topepo:
The existing tests should catch this (since they are currently failing).

R/fit_helpers.R
```diff
@@ -6,7 +6,9 @@
 form_form <-
   function(object, control, env, ...) {
 
-    check_outcome(eval_tidy(env$formula[[2]], env$data), object)
+    if (!inherits(env$data, "tbl_spark")) {
+      check_outcome(eval_tidy(env$formula[[2]], env$data), object)
+    }
```
DavisVaughan, line 11: So things like `levels_from_formula(env$formula, env$data)` used below work even with `env$data` as a spark tbl?

DavisVaughan, line 11: We could also consider flipping around the if statement. It could be `if (inherits(env$data, "data.frame"))` instead, which would avoid doing this check for spark and possibly for H2O, assuming that it wouldnt work for some kind of H2OFrame object either

PR #687, opened 2022-03-22
refine documentation of set_engine to make passing arguments to the engine fit function clearer

jmarshallnz:
I was confused with how to pass arguments directly to the fit function (see https://github.com/tidymodels/discrim/pull/48) so this refines the language of the `set_engine` documentation so that hopefully others might avoid the same confusion.


juliasilge (approved):
Thank you so much @jmarshallnz! 🙌

PR #688, opened 2022-03-22
export `xgb_pred()`

hfrick:
I'd like to export `xgb_pred()` for use in censored, similar to how `xgb_train()` is already exported.

For additional context, this is the PR in censored: https://github.com/tidymodels/censored/pull/163

Caveat: I currently can't run `document()` 😞


R/boost_tree.R
```diff
@@ -383,6 +383,10 @@ maybe_proportion <- function(x, nm) {
   }
 }
 
+#' @rdname xgb_train
+#' @param newdata A rectangular data object, such as a data frame.
```
hfrick, line 387: should this a) be called `new_data` and b) then inherit parameter documentation from `predict()`?

R/boost_tree.R
```diff
@@ -383,6 +383,10 @@ maybe_proportion <- function(x, nm) {
   }
 }
 
+#' @rdname xgb_train
```
DavisVaughan, line 386: Since the Description section of `?xgb_train` mentions what `xgb_train()` does, I think you should add a line there about what `xgb_predict()` does

DavisVaughan (commented):
I think we should figure out how to document this before merging, but otherwise looks good

R/boost_tree.R
DavisVaughan: me too

topepo: Ok, I'd like to go back to `xgb_pred()` even though `xgb_predict()` is better. The longer name will break models previously fit with parsnip but predicted using this version.

PR #692, opened 2022-03-25
case weights

### LINKED ISSUE #136: No support for weights
The `weights` argument in `fit` does nothing. Most modeling packages allow for case weights, so is this going to be supported in the future?

topepo:
Closes #136


topepo:
All models with case weights will be tested in extratests. 

Spark support is still outstanding because there are some API issues to work out.  Sparks models want a column name for the weights argument.

R/logistic_reg_data.R
```diff
@@ -220,7 +220,7 @@ set_fit(
   mode = "classification",
   value = list(
     interface = "matrix",
-    protect = c("x", "y", "wi"),
+    protect = c("x", "y"),
```
DavisVaughan, line 223: What was going on here?

R/misc.R
```diff
@@ -385,3 +385,18 @@ stan_conf_int <- function(object, newdata) {
 
   penalty
 }
+
+
+check_case_weights <- function(x, spec) {
+  if (is.null(x)) {
+    return(invisible(NULL))
+  }
+  if (!inherits(x, "hardhat_case_weights")) {
```
DavisVaughan, line 394:
```suggestion
  if (!hardhat::is_case_weights(x)) {
```

tests/testthat/test-case-weights.R
```diff
@@ -0,0 +1,57 @@
+
+test_that('case weights with xy method', {
+
+  skip_if_not_installed("C50")
+  skip_if_not_installed("modeldata")
+  library(hardhat)
```
DavisVaughan, line 6: Probably best not to `library(hardhat)`

tests/testthat/test-case-weights.R
```diff
@@ -0,0 +1,57 @@
...
+  data("two_class_dat", package = "modeldata")
+
+  wts <- runif(nrow(two_class_dat))
+  wts <- ifelse(wts < 1/5, 0, 1)
+  two_class_subset <- two_class_dat[wts != 0, ]
+  wts <- importance_weights(wts)
```
DavisVaughan, line 12: parsnip is one of the packages where I think it makes sense to re-export both `frequency_weights()` and `importance_weights()`. That way users get access to them and you can use them in tests without needing to library hardhat or prefix them with `hardhat::`

tests/testthat/test-case-weights.R
```diff
@@ -0,0 +1,57 @@
...
+      set_engine("C5.0") %>%
+      set_mode("classification") %>%
+      fit(Class ~ ., data = two_class_dat, case_weights = wts)
+  },
+  regexp = NA)
+
+  expect_output(
+    print(C5_bst_wt_fit$fit$call),
+    "weights = weights"
+  )
```
DavisVaughan, line 27: Is there anything inside the C5 model object that you can check for to prove that you are using case weights? That might be more robust

DavisVaughan (commented):
Looking pretty good so far!

tests/testthat/test-case-weights.R
topepo: Should the individual weight functions be reexported?

tests/testthat/test-case-weights.R
topepo: OK this answers ☝️

tests/testthat/test-case-weights.R
topepo: No, not really. There are a lot of models that don't keep records of the weights.  Also, with trees and earth models, the subsetting weights that I use don't produce the same results as the model when those rows are specifically absent. 

My thinking is, for those cases, I'll make sure that the weights were used in the call and the model didn't fail.

R/arguments.R
```diff
@@ -165,6 +169,13 @@ make_form_call <- function(object, env = NULL) {
   # sub in actual formula
   fit_args[[ unname(data_args["formula"]) ]]  <- env$formula
 
+  # Add in case weights symbol
+  if (uses_weights) {
+    fit_args[[ unname(data_args["weights"]) ]]  <- rlang::expr(weights)
+  }
```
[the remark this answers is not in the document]
topepo: I added a lot of comments (that should feed back into tidymodels.org) that explain what these elements do. I cleaned up the code a little too in the process.

DavisVaughan:
- [ ] Attempt to remove `convert_case_weights()` because it doesn't look to be actually used anymore

PR #696, opened 2022-04-04
xgboost changes for case weights 

topepo:
Related to tidymodels/censored#163


topepo:
Tests are in 99d1bc3

R/boost_tree.R
```diff
@@ -256,11 +257,11 @@ check_args.boost_tree <- function(object) {
...
-  x, y,
-  max_depth = 6, nrounds = 15, eta  = 0.3, colsample_bynode = NULL,
-  colsample_bytree = NULL, min_child_weight = 1, gamma = 0, subsample = 1,
-  validation = 0, early_stop = NULL, objective = NULL, counts = TRUE,
-  event_level = c("first", "second"), ...) {
+    x, y,
+    max_depth = 6, nrounds = 15, eta  = 0.3, colsample_bynode = NULL,
+    colsample_bytree = NULL, min_child_weight = 1, gamma = 0, subsample = 1,
+    validation = 0, early_stop = NULL, objective = NULL, counts = TRUE,
+    event_level = c("first", "second"), weights = NULL, ...) {
```
DavisVaughan, line 264: Why did this get tabbed over so much

DavisVaughan (approved):
Looks reasonable to me

R/boost_tree.R
topepo: I can put it closer to the other data args (if that's what you mean)...

hfrick (approved):
looks good to me!

We will probably need to pass the objective from `xgb_train()` to `as_xgb_data()` because the labels required differ between `survival:cox` and `survival:aft` but we'll figure that out in censored and then open another PR

R/boost_tree.R
DavisVaughan: I meant the added space before `max_depth`, `colsample_bytree`, and the other args that start their respective lines

It isn't important, just seemed weird

PR #705, opened 2022-04-20
Ensure that `fit_xy()` patches the formula environment with weights

DavisVaughan:
- Passed the weights through `xy_form()` to `.convert_xy_to_form_fit()`
- Made `.convert_xy_to_form_fit()` apply the same patch to `formula` as `fit()`

I think the tests are going to fail due to some MLP related things that also currently fail in the case weights PR.


tests/testthat/test-case-weights.R
```diff
@@ -78,5 +110,34 @@ test_that('case weights with formula method', {
...
+  set.seed(1)
+  wts <- runif(nrow(ames))
+  wts <- ifelse(wts < 1/5, 0L, 1L)
+  ames_subset <- ames[wts != 0, ]
+  wts <- frequency_weights(wts)
+
+  expect_error(
+    lm_wt_fit <-
+      linear_reg() %>%
+      fit_xy(
```
DavisVaughan, line 128: This is the main test that checks that this works. It goes through `fit_xy()` but the underlying model requires the formula method, so it has to use `xy_form()`

topepo (approved):

PR #706, opened 2022-04-20
transition to `vctrs::s3_register`

### LINKED ISSUE #652: remove `parsnip:::s3_register()`
We currently have a version of `s3_register()` in `R/zzz.R` and use that along with `vctrs::s3_register()` in `.onLoad()`. Can we use only one version (presumably the vctrs version)?

simonpcouch:
Following up on the release of vctrs this month, closes #652. Not sure if there are additional ad-hoc ways this is tested, but from what I could tell, the `tidy` methods interface in the same way as before this change!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -35,7 +35,7 @@ Imports:
     tibble (>= 2.1.1),
     tidyr (>= 1.0.0),
     utils,
-    vctrs (>= 0.2.0),
+    vctrs (>= 0.4.0),
     withr
 Suggests: 
     C50,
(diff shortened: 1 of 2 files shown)


DESCRIPTION
```diff
@@ -35,7 +35,7 @@ Imports:
     tibble (>= 2.1.1),
     tidyr (>= 1.0.0),
     utils,
-    vctrs (>= 0.2.0),
+    vctrs (>= 0.4.0),
```
DavisVaughan, line 38:
```suggestion
    vctrs (>= 0.4.1),
```

We had a few small bugs in 0.4.0 so bumping to 0.4.1 will hopefully force people to get the fixes

DavisVaughan (approved):

DESCRIPTION
simonpcouch: You got it👍

Issue #708, opened 2022-04-28
Inconsistency with return type for probability type predictions for two class GAM models 


oj713:
## The problem

I'm encountering an inconsistency with generating probability predictions for two class GAM models. The predictions are being returned as arrays within a tibble, rather than just a tibble of numeric values. I believe this issue arises because `mgcv` returns an *array* of predictions, rather than a vector.

Issue #541 covers a similar issue around GAM class predictions where `mgcv`'s prediction format was also the source of the problem. I believe the fix for this problem would closely mirror the fix for that issue (fix #542). 

## Reproducible example

``` r
library(parsnip)

set.seed(33)
twoClassSim <- caret::twoClassSim()

gam <- gen_additive_mod() |>
  set_mode("classification") |>
  fit(Class ~ Linear01, 
      data = twoClassSim)

pred <- predict(gam, head(twoClassSim), type = "prob")

# demonstrating error
class(pred$.pred_Class1)
#> [1] "array"
dplyr::glimpse(pred)
#> Rows: 6
#> Columns: 2
#> $ .pred_Class1 <dbl> <array[6]>
#> $ .pred_Class2 <dbl> <array[6]>

# session information 
sessionInfo()
#> R version 4.2.0 (2022-04-22)
#> Platform: aarch64-apple-darwin20 (64-bit)
#> Running under: macOS Monterey 12.3.1
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/4.2-arm64/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> other attached packages:
#> [1] parsnip_0.2.1
#> 
#> loaded via a namespace (and not attached):
#>  [1] Rcpp_1.0.8.3         lubridate_1.8.0      lattice_0.20-45     
#>  [4] listenv_0.8.0        tidyr_1.2.0          class_7.3-20        
#>  [7] digest_0.6.29        ipred_0.9-12         foreach_1.5.2       
#> [10] utf8_1.2.2           parallelly_1.31.1    R6_2.5.1            
[...]

PR #709, opened 2022-04-28
Fixing inconsistency for probability predictions for two class GAM models 

oj713:
This pull request contains a minor code change that addresses an inconsistency for probability predictions for two class GAM models -- see issue #708.


EmilHvitfeldt:
Looking good! Thank you @oj713 🙌

Issue #712, opened 2022-05-05
`usekernel` setting in Naive Bayes models with `engine = "klaR"`


deschen1:
Looking at the help for `?parsnip::details_naive_Bayes_klaR` it says:

> Note that usekernel is always set to TRUE for the klaR engine.

Does that mean parsnip doesn't allow to change the `usekernel` argument to FALSE (if so, why?) or does it only mean the default is TRUE and could easily be changed to FALSE? I'm a bit confused about it, because a) the klaR package itself does allow to set the usekernel argument for Naive Bayes models and b) using the alternative engine from package naivebayes doesn't come iwth this restriction as can be checked in `?parsnip::details_naive_Bayes_naivebayes`.

EmilHvitfeldt:
That note refers to the fact that `usekernel` has been set up `TRUE` by discrim, which is different than what the underlying engine function normally does. You can change this by setting it as an `engine` argument.

It appears that this line is missing from `?parsnip::details_naive_Bayes_naivebayes` and should be added, thanks for noticing the discrepancy.

``` r
library(discrim)
#> Loading required package: parsnip

naive_Bayes() %>% 
  set_engine("klaR") %>% 
  translate()
#> Naive Bayes Model Specification (classification)
#> 
#> Computational engine: klaR 
#> 
#> Model fit template:
#> discrim::klar_bayes_wrapper(x = missing_arg(), y = missing_arg(), 
#>     usekernel = TRUE)

naive_Bayes() %>% 
  set_engine("klaR", usekernel = FALSE) %>% 
  translate()
#> Naive Bayes Model Specification (classification)
#> 
#> Engine-Specific Arguments:
#>   usekernel = FALSE
#> 
#> Computational engine: klaR 
#> 
#> Model fit template:
#> discrim::klar_bayes_wrapper(x = missing_arg(), y = missing_arg(), 
#>     usekernel = FALSE)
```

<sup>Created on 2022-05-06 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

deschen1:
Ok, thanks for the explanation. Will try it out!

PR #714, opened 2022-05-06
add missing comment about usekernel in naive_Bayes naivebayes

EmilHvitfeldt:
Missing documentation has been added to close https://github.com/tidymodels/parsnip/issues/712.


juliasilge:
I think this documentation is still confusing and sounds like `usekernel` cannot be changed. Instead, can we say it is an engine argument, with a default of `TRUE`?

EmilHvitfeldt:
I did a quick rewording, let me know what you think

juliasilge:
Yes, more clear now! 🙌

PR #715, opened 2022-05-06
Update README.Rmd

PursuitOfDataScience:
`data` should equal to `dat` rather than `.`


EmilHvitfeldt:
Hello @PursuitOfDataScience 👋 

Would you be able to knit this document so the change will propagate to `README.md`? You can use `devtools::build_readme()` to do that.

PursuitOfDataScience:
Check it now.

PR #718, opened 2022-05-09
document elastic net regularization via mixture

### LINKED ISSUE #570: Document that linear_reg() and logistic_reg() can performs elastic net regularization
The `mixture` argument notes that we are doing a pure lasso model when `mixture = 1`, and ridge regression when `mixture = 0`. I think it would be helpful to mention that it is called elastic net when `1 > mixture > 0`. I'm not quite sure if elastic net is defined as being between lasso and ridge, or as encompassing.

simonpcouch:
Closes #570. In my experience, "interpolate" is the verbiage here. :)

Will need someone else to generate documentation for me—spent a while trying to debug, and eventually came across a set of errors in `knit_engine_docs`:

```
Failed with error:  'package 'parsnip' required by 'censored' could not be found'
Failed with error:  'package 'parsnip' required by 'discrim' could not be found'
# etc etc
```

that's not worth me troubleshooting if others are able to follow [inst/README-DOCS.md](https://github.com/tidymodels/parsnip/blob/main/inst/README-DOCS.md) just fine.🛫

### DIFF
--- a/R/linear_reg.R
+++ b/R/linear_reg.R
@@ -17,10 +17,14 @@
 #'  model is `"lm"`.
 #' @param penalty A non-negative number representing the total
 #'  amount of regularization (specific engines only).
-#' @param mixture A number between zero and one (inclusive) that is the
-#'  proportion of L1 regularization (i.e. lasso) in the model. When
-#'  `mixture = 1`, it is a pure lasso model while `mixture = 0` indicates that
-#'  ridge regression is being used (specific engines only).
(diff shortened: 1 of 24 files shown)


simonpcouch:
Aha—I think `load_all` got me here. Was able to knit engine docs after local install + restart. :)

> Where should we document this?

Follow-up from conversation with Max: This is indeed documented [here](https://github.com/tidymodels/parsnip/blob/main/inst/README-DOCS.md)!

EmilHvitfeldt (approved):
LGTM!

PR #728, opened 2022-05-20
Update svm_rbf.R

PursuitOfDataScience:
Added more explanation for `cost`.

### DIFF
--- a/R/svm_rbf.R
+++ b/R/svm_rbf.R
@@ -19,7 +19,8 @@
 #'  to use for fitting. Possible engines are listed below. The default for this
 #'  model is `"kernlab"`.
 #' @param cost A positive number for the cost of predicting a sample within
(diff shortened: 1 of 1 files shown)


juliasilge:
Thank you so much for this contribution @PursuitOfDataScience! 🙌 

I think that we could stand for a bit more documentation for `cost` here, like you have. However, it doesn't look like this suggestion will be that helpful for folks who haven't already learned about the relationship between cost, model complexity, how an SVM works, etc. It seems more like a reminder only focused the direction of changes in `cost`. 

Do you have a suggestion of about the same length that would be more helpful to a broader audience (including with less background knowledge)?

topepo:
Thanks for these contributions, but we are going to close them. They don't feel like a good fit for these particular docs.

PR #729, opened 2022-05-20
Update svm_linear.R

PursuitOfDataScience:
Similar change as #728


juliasilge:
In addition to my feedback on #728, it would be good to say here that `cost` doesn't make much difference for a linear SVM.

topepo:
Thanks for these contributions, but we are going to close them. They don't feel like a good fit for these particular docs.

Issue #544, opened 2021-08-19
Issue with parsnip survival_reg() funcion


davestr2:
I am trying to use the new survival_reg() function in parsnip 0.1.7. When I use this, I get an error when trying to do a fit on a workflow that uses this function. The older version surv_reg(), that is still available in 0.1.7, works as desired. 

If this is a duplicate issue, then sorry about that !

``` r
# Trying to run code found in Chapter 7 Section 7.4.1 of the
#     Tidy Modeling with R book (https://www.tmwr.org/workflows.html#special-model-formulas)

# Load the libraries

library(tidymodels)
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip
library(survival)

# Create  model for use with survival package:


# Use the "old" pre parsnip 0.1.6 function  -----   surv_reg()
#     As is done in the manual

parametric_model_Old_func <- 
  surv_reg() %>% 
  set_engine("survival")
#> Warning: `surv_reg()` was deprecated in parsnip 0.1.6.
#> Please use `survival_reg()` instead.

# Notice the warning about surv_reg() being depreciated and to use survival_reg()
#       See this  function being used below


# Run the output from above thru translate() function

surv_reg() %>% set_engine("survival") %>% translate()
#> Warning: `surv_reg()` was deprecated in parsnip 0.1.6.
#> Please use `survival_reg()` instead.
#> Parametric Survival Regression Model Specification (regression)
#> 
#> Computational engine: survival 
#> 
#> Model fit template:
#> survival::survreg(formula = missing_arg(), data = missing_arg(), 
#>     weights = missing_arg(), model = TRUE)

# Create a workflow

parametric_workflow_Old_func <- 
  workflow() %>% 
  # Pass the data along as-is: 
  add_variables(outcome = c(fustat, futime), predictors = c(age, rx)) %>% 
  add_model(parametric_model_Old_func, 
            # This formula is given to the model
            formula = Surv(futime, fustat) ~ age + strata(rx))

# fit the workflow

parametric_fit_Old_func <- fit(parametric_workflow_Old_func, data = ovarian)

# Take a look at the fitted model

parametric_fit_Old_func
#> ══ Workflow [trained] ══════════════════════════════════════════════════════════
#> Preprocessor: Variables
#> Model: surv_reg()
#> 
#> ── Preprocessor ────────────────────────────────────────────────────────────────
#> Outcomes: c(fustat, futime)
[...]

PR #732, opened 2022-05-22
prompt on unavailable extension packages

### LINKED ISSUE #731: `translate()` print method fails uninformatively when extension isn't loaded
``` r
library(parsnip)

C5_rules() %>% translate("C5.0")
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0 
#> 
#> Model fit template:
#> Error in `call2()`:
#> ! Can't create call to non-callable object

library(rules)

C5_rules() %>% translate("C5.0")
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0 
#> 
#> Model fit template:
#> rules::c5_fit(x = missing_arg(), y = missing_arg(), weights = missing_arg())
```

<sup>Created on 2022-05-22 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

simonpcouch:
Closes #731.

I've had a few moments of confusion as I've familiarized with parsnip where I unknowingly specify a model specification/engine/mode combination that requires an extension package, and only come across failures later on in a modeling pipeline. This seems to be the source of confusion in #544 as well.

This PR refines print methods to nudge users to load an extension package if it's needed.

``` r
# warns informatively when it ought to:
library(parsnip)

C5_rules()
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ The parsnip extension package rules implements support for this specification.
#> ℹ Please install (if needed) and load to continue.
#> 
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0

C5_rules(engine = "C5.0") %>% translate()
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ The parsnip extension package rules implements support for this specification.
#> ℹ Please install (if needed) and load to continue.
#> 
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0

# doesn't prompt and prints out the call as before if the implementing extension is loaded:
library(rules)

C5_rules()
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0

C5_rules(engine = "C5.0") %>% translate()
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0 
#> 
#> Model fit template:
#> rules::c5_fit(x = missing_arg(), y = missing_arg(), weights = missing_arg())

# otherwise behaves the same as before:
linear_reg() %>% translate()
#> Linear Regression Model Specification (regression)
#> 
#> Computational engine: lm 
#> 
#> Model fit template:
#> stats::lm(formula = missing_arg(), data = missing_arg(), weights = missing_arg())

linear_reg()
#> Linear Regression Model Specification (regression)
#> 
#> Computational engine: lm
```

<sup>Created on 2022-05-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

[EDIT: updated reprex and explanation to reflect changes re: Julia's edits]

### DIFF
--- a/R/bag_mars.R
+++ b/R/bag_mars.R
@@ -47,7 +47,7 @@ print.bag_mars <- function(x, ...) {
   cat("Bagged MARS Model Specification (", x$mode, ")\n\n", sep = "")
   model_printer(x, ...)
 
-  if (!is.null(x$method$fit$args) && has_loaded_implementation(class(x)[1], x$engine, x$mode)) {
+  if (is_printable_spec(x)) {
     cat("Model fit template:\n")
     print(show_call(x))
   }
--- a/R/bag_mars.R
+++ b/R/bag_mars.R
@@ -47,7 +47,7 @@ print.bag_mars <- function(x, ...) {
   cat("Bagged MARS Model Specification (", x$mode, ")\n\n", sep = "")
   model_printer(x, ...)
 
-  if (!is.null(x$method$fit$args) && has_loaded_implementation(x)) {
+  if (!is.null(x$method$fit$args) && has_loaded_implementation(class(x)[1], x$engine, x$mode)) {
     cat("Model fit template:\n")
     print(show_call(x))
   }
--- a/R/bag_mars.R
+++ b/R/bag_mars.R
@@ -47,7 +47,7 @@ print.bag_mars <- function(x, ...) {
   cat("Bagged MARS Model Specification (", x$mode, ")\n\n", sep = "")
   model_printer(x, ...)
 
-  if (!is.null(x$method$fit$args)) {
+  if (!is.null(x$method$fit$args) && has_loaded_implementation(x)) {
     cat("Model fit template:\n")
     print(show_call(x))
   }
--- a/R/bag_mars.R
+++ b/R/bag_mars.R
@@ -47,7 +47,7 @@ print.bag_mars <- function(x, ...) {
   cat("Bagged MARS Model Specification (", x$mode, ")\n\n", sep = "")
   model_printer(x, ...)
 
-  if (!is.null(x$method$fit$args) && has_loaded_implementation(x)) {
+  if (!is.null(x$method$fit$args)) {
     cat("Model fit template:\n")
     print(show_call(x))
   }
--- a/R/misc.R
+++ b/R/misc.R
@@ -74,9 +74,6 @@ model_printer <- function(x, ...) {
       }
     }
   }
-  if (!has_loaded_implementation(x)) {
-    cat(prompt_missing_implementation(x), fill = 80)
-  }
 }
 
 is_missing_arg <- function(x)
@@ -88,20 +85,18 @@ is_missing_arg <- function(x)
 #
 # return FALSE if:
 # * the model needs an extension and it is _not_ loaded
-has_loaded_implementation <- function(x) {
-  spec_ <- class(x)[1]
-  mode_ <- x$mode
-
+has_loaded_implementation <- function(spec_, engine_, mode_) {
   if (isFALSE(mode_ %in% c("regression", "censored regression", "classification"))) {
     mode_ <- c("regression", "censored regression", "classification")
   }
+  eng_cond <- if (is.null(engine_)) {TRUE} else {quote(engine == engine_)}
 
   avail <-
-    show_engines(spec_) %>%
-    dplyr::filter(mode %in% mode_)
+    get_from_env(spec_) %>%
+    dplyr::filter(mode %in% mode_, !!eng_cond)
   pars <-
     utils::read.delim(system.file("models.tsv", package = "parsnip")) %>%
-    dplyr::filter(model == spec_, mode %in% mode_, is.na(pkg))
+    dplyr::filter(model == spec_, !!eng_cond, mode %in% mode_, is.na(pkg))
 
   if (nrow(pars) > 0 || nrow(avail) > 0) {
     return(TRUE)
@@ -115,23 +110,19 @@ has_loaded_implementation <- function(x) {
 #
 # if there's a "pre-registered" extension supporting that setup,
 # nudge the user to install/load it.
-prompt_missing_implementation <- function(x) {
-  spec_ <- class(x)[1]
-  engine_ <- x$engine
-  mode_ <- x$mode
-
+prompt_missing_implementation <- function(spec_, engine_, mode_) {
   avail <-
     show_engines(spec_) %>%
-    dplyr::filter(mode == mode)
+    dplyr::filter(mode == mode_, engine == engine_)
   all <-
     utils::read.delim(system.file("models.tsv", package = "parsnip")) %>%
-    dplyr::filter(model == spec_, mode == mode, !is.na(pkg)) %>%
+    dplyr::filter(model == spec_, mode == mode_, engine == engine_, !is.na(pkg)) %>%
     dplyr::select(-model)
 
   msg <-
     glue::glue(
       "A parsnip implementation for `{spec_}` {mode_} model ",
-      "specifications using the `{engine_}` engine is not loaded."
+      "specifications using the `{engine_}` engine is not loaded. "
     )
 
   if (nrow(avail) == 0 && nrow(all) > 0) {
@@ -139,16 +130,17 @@ prompt_missing_implementation <- function(x) {
       glue::glue_collapse(c(
         msg,
         glue::glue_collapse(c(
-          "\nThe parsnip extension package {",
+          "The parsnip extension package {",
           cli::col_yellow(all$pkg[[1]]),
-          "} implements support for this specification/mode/engine combination. \n",
-          "Please install (if needed) and load to continue."
+          "} implements support for this specification/mode/engine combination. ",
+          "Please install (if needed) and load to continue.\n"
         ))
       ))
   }
 
   msg
 }
+
 
 #' Print the model call
 #'
@@ -242,9 +234,13 @@ update_dot_check <- function(...) {
 #' @export
 #' @keywords internal
 #' @rdname add_on_exports
-new_model_spec <- function(cls, args, eng_args, mode, method, engine) {
+new_model_spec <- function(cls, args, eng_args, mode, method, engine, new = TRUE) {
 
   check_spec_mode_engine_val(cls, engine, mode)
+
+  if ((!has_loaded_implementation(cls, engine, mode)) && new) {
+    rlang::inform(prompt_missing_implementation(cls, engine, mode))
+  }
 
   out <- list(args = args, eng_args = eng_args,
               mode = mode, method = method, engine = engine)
--- a/R/misc.R
+++ b/R/misc.R
@@ -241,11 +241,12 @@ update_dot_check <- function(...) {
 #' @export
 #' @keywords internal
 #' @rdname add_on_exports
-new_model_spec <- function(cls, args, eng_args, mode, method, engine, new = TRUE) {
+new_model_spec <- function(cls, args, eng_args, mode, method, engine,
+                           check_missing_spec = TRUE) {
 
   check_spec_mode_engine_val(cls, engine, mode)
 
-  if ((!has_loaded_implementation(cls, engine, mode)) && new) {
+  if ((!has_loaded_implementation(cls, engine, mode)) && check_missing_spec) {
     rlang::inform(prompt_missing_implementation(cls, engine, mode))
   }
 
--- a/R/misc.R
+++ b/R/misc.R
@@ -119,10 +119,14 @@ prompt_missing_implementation <- function(spec_, engine_, mode_) {
     dplyr::filter(model == spec_, mode == mode_, engine == engine_, !is.na(pkg)) %>%
     dplyr::select(-model)
 
+  if (identical(mode_, "unknown")) {
+    mode_ <- ""
+  }
+
   msg <-
     glue::glue(
-      "A parsnip implementation for `{spec_}` {mode_} model ",
-      "specifications using the `{engine_}` engine is not loaded."
+      "parsnip could not locate an implementation for `{spec_}` {mode_} model ",
+      "specifications using the `{engine_}` engine."
     )
 
   if (nrow(avail) == 0 && nrow(all) > 0) {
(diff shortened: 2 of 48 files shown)


juliasilge:
I think this would make the most sense in an `rlang::inform()` vs. printing (and we have a lot of tools for handling the rlang messages).

simonpcouch:
Ah, thank you @juliasilge! I agree that `rlang::inform` likely makes more sense.

Just pushed those changes—they're a bit more invasive now but do result in a better interface.

simonpcouch:
Wrapping up for the night, though this isn't quite there.🌝🌚

R/misc.R
```diff
@@ -89,18 +155,10 @@ is_missing_arg <- function(x)
 show_call <- function(object) {
   object$method$fit$args <-
     map(object$method$fit$args, convert_arg)
-  if (
```
simonpcouch, line 92: Gets rid of these conditionals as the code pointed to in the `is.null(object$method$fit$func["pkg"])` condition is what resulted in the `call2` errors noted in the issue reprex. Couldn't find coverage for those lines before this PR, so I think that code was otherwise unused.

R/bag_mars.R
```diff
@@ -47,7 +47,7 @@ print.bag_mars <- function(x, ...) {
   cat("Bagged MARS Model Specification (", x$mode, ")\n\n", sep = "")
   model_printer(x, ...)
 
-  if (!is.null(x$method$fit$args)) {
+  if (!is.null(x$method$fit$args) && has_loaded_implementation(class(x)[1], x$engine, x$mode)) {
```
simonpcouch, line 50: Most of the files changed are of this form—the additional condition guards us from printing `"Model fit template"` and then reaching the condition that would result in the `call2` error from the issue reprex.

simonpcouch (commented):
A quick self-review to give some context for changes. :)

R/misc.R
```diff
@@ -181,10 +238,14 @@ update_dot_check <- function(...) {
 #' @export
 #' @keywords internal
 #' @rdname add_on_exports
-new_model_spec <- function(cls, args, eng_args, mode, method, engine) {
+new_model_spec <- function(cls, args, eng_args, mode, method, engine, new = TRUE) {
```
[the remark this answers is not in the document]
juliasilge: I don't know if this is just a naming issue, but `new_model_spec(new = TRUE)` or even worse `FALSE` 😱 seems weird to me, and I am almost sure I will be super confused if I come back to this later. Do you have any ideas on this?

R/misc.R
```diff
@@ -79,6 +79,72 @@ model_printer <- function(x, ...) {
...
+  if (nrow(avail) == 0 && nrow(all) > 0) {
+    msg <-
+      glue::glue_collapse(c(
+        msg,
+        glue::glue_collapse(c(
+          "\nThe parsnip extension package {",
+          cli::col_yellow(all$pkg[[1]]),
+          "} implements support for this specification. \n",
+          "Please install (if needed) and load to continue.\n"
+        ))
```
juliasilge, line 141: For consistence we use the [tidyverse style guide](https://style.tidyverse.org/documentation.html#r-code) which has no special font or anything for package names:

```suggestion
        glue::glue_collapse(c(
          "\nThis specification is available in the  ",
          cli::col_yellow(all$pkg[[1]]),
          " package. \n"
        ))
```

I'm a bit worried about the `\n` in this message and how that may interact with the rlang messages. Can we use the built-in bulleting, etc instead of hardcoding new lines into the message? 

Why is this `glue_collapse()` instead of just `glue()`. Because there can be more than one? Maybe we should use the pluralization from cli?

R/bag_mars.R
juliasilge: How about a new little function `is_printable_spec(x)` (or something like that) to wrap this up? I worry about us having to make changes in so many places.

R/misc.R
juliasilge: I think this function should return a named character vector, with no name for the first element, `i` for the others, etc, to be used as `message` in `rlang::inform()`.

R/misc.R
```diff
@@ -79,6 +79,72 @@ model_printer <- function(x, ...) {
...
+    dplyr::select(-model)
+
+  if (identical(mode_, "unknown")) {
+    mode_ <- ""
+  }
+
+  msg <-
+    glue::glue(
+      "parsnip could not locate an implementation for `{spec_}` {mode_} model ",
+      "specifications using the `{engine_}` engine."
```
juliasilge, line 129:
```suggestion
      'parsnip could not locate an implementation for `{spec_}`, for {mode_} and the  `"{engine_}"` engine'
```

I like to write long `glue()` lines using `\\` rather than manually breaking with commas. One of the benefits of glue, I think!

R/misc.R
```diff
@@ -79,6 +79,72 @@ model_printer <- function(x, ...) {
...
+    mode_ <- c("regression", "censored regression", "classification")
+  }
+  eng_cond <- if (is.null(engine_)) {TRUE} else {quote(engine == engine_)}
+
+  avail <-
+    get_from_env(spec_) %>%
+    dplyr::filter(mode %in% mode_, !!eng_cond)
+  pars <-
+    utils::read.delim(system.file("models.tsv", package = "parsnip")) %>%
+    dplyr::filter(model == spec_, !!eng_cond, mode %in% mode_, is.na(pkg))
```
juliasilge, line 99: Is this an expensive check to do every time `new_model_spec()` is run (which is run a lot, I think)? The `env` is just there, but what about reading the file?

juliasilge (commented):
It will be a huge help to give more useful info in this situation! 🙌 

I've got a bit of specific feedback and then some questions about the approach.

R/bag_mars.R
simonpcouch: I agree that this conditional is a bit gnarly, especially given how many places it appears. Went ahead and implemented your change + I’m planning to follow up this PR with a refactor of print methods that should reduce this duplication quite a bit. :)

R/misc.R
simonpcouch: Thank you for the nudges on making better use of glue and inform—learned a lot about those functions from your comments.

The message now reads:

``` r
library(parsnip)

C5_rules()
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ  The parsnip extension package rules implements support for this specification.
#> ℹ  Please install (if needed) and load to continue.
#> 
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0

C5_rules(engine = "C5.0") %>% translate()
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ  The parsnip extension package rules implements support for this specification.
#> ℹ  Please install (if needed) and load to continue.
#> 
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0
```

<sup>Created on 2022-05-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

and the code to get there is a good bit cleaner.

(Acknowledgement that I’m a serial violator of the no-special-package-name-formatting convention.🫠  Sorry.)

R/misc.R
simonpcouch: Certainly more expensive than our other checks. But—

* It’s only run on initiation of new model types (i.e. never at `set_engine`, `update`, etc)
* It’s possibly negligible compared to fit times:

``` r
library(parsnip)

# running the check
system.time(mod <- C5_rules())
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ The parsnip extension package rules implements support for this specification.
#> ℹ Please install (if needed) and load to continue.
#>    user  system elapsed 
#>   0.038   0.001   0.039

# doesn't run the check
system.time(mod <- mod %>% set_engine("C5.0"))
#>    user  system elapsed 
#>   0.008   0.000   0.009

# model fit to a lil fella
library(rules)

mtcars$cyl <- as.factor(mtcars$cyl)

system.time(mod %>% fit(cyl ~ ., mtcars))
#>    user  system elapsed 
#>   0.445   0.020   0.467
```

<sup>Created on 2022-05-25 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

R/misc.R
```diff
@@ -79,6 +79,75 @@ model_printer <- function(x, ...) {
...
+    mode_ <- ""
+  }
+
+  msg <-
+    glue::glue(
+      "parsnip could not locate an implementation for `{spec_}` {mode_} model \\
+       specifications using the `{engine_}` engine."
+    )
+
+  if (nrow(avail) == 0 && nrow(all) > 0) {
```
juliasilge, line 137: Can we add a test that prints the first msg but not this bit?

R/misc.R
```diff
@@ -79,6 +79,75 @@ model_printer <- function(x, ...) {
...
+  !is.null(x$method$fit$args) &&
+  has_loaded_implementation(class(x)[1], x$engine, x$mode)
+}
+
+# construct a message informing the user that there are no
+# implementations for the current model spec / mode / engine.
+#
+# if there's a "pre-registered" extension supporting that setup,
+# nudge the user to install/load it.
+prompt_missing_implementation <- function(spec_, engine_, mode_) {
```
juliasilge, line 118:
```suggestion
inform_missing_implementation <- function(spec_, engine_, mode_) {
```
This will help for future reading, I think, to clarify what this is being used for

R/misc.R
juliasilge: The fact that it could be common to read this file multiple times seems like something to check and get other feedback on. We could read this file one time at `.onLoad` or something? Let's have one more person look at this part of the PR and see what they think.

R/misc.R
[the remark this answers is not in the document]
juliasilge: What do you think about `check_missing_spec = TRUE`? Since this argument is less about whether the spec is new or not and more about whether it will be checked?

juliasilge (changes requested):

R/misc.R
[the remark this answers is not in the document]
simonpcouch: That's definitely better—4dd0fb9.

R/misc.R
simonpcouch: Sure thing—6c06315!

R/misc.R
simonpcouch: I think that's a good observation. After [9918043](https://github.com/tidymodels/parsnip/pull/732/commits/9918043a135e0782d40f3233335642f23666c53f), parsnip now assigns the `models.tsv` system file as `model_info_table` internally so we can skip that load step on checks:

``` r
library(parsnip)
  
# now just assigns models.tsv to internal object:
head(parsnip:::model_info_table)
#>      model                mode engine      pkg
#> 1 bag_mars      classification  earth baguette
#> 2 bag_mars          regression  earth baguette
#> 3 bag_tree censored regression  rpart censored
#> 4 bag_tree      classification   C5.0 baguette
#> 5 bag_tree      classification  rpart baguette
#> 6 bag_tree          regression  rpart baguette

# running the check. used to be:
#>    user  system elapsed 
#>   0.038   0.001   0.039
system.time(mod <- C5_rules())
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ The parsnip extension package rules implements support for this specification.
#> ℹ Please install (if needed) and load to continue.
#>    user  system elapsed 
#>   0.033   0.001   0.033


# doesn't run the check. used to be:
#>    user  system elapsed 
#>   0.008   0.000   0.009
system.time(mod <- mod %>% set_engine("C5.0"))
#>    user  system elapsed 
#>   0.007   0.000   0.007


# model fit to a lil fella, for comparison
library(rules)

mtcars$cyl <- as.factor(mtcars$cyl)

system.time(mod %>% fit(cyl ~ ., mtcars))
#>    user  system elapsed 
#>   0.440   0.025   0.464
```

<sup>Created on 2022-05-27 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I opted not to ditch that `models.tsv` file entirely as it seems our machinery to keep it updated with the current docs system works well.

simonpcouch:
This should be ready for your review, @topepo! 

One unresolved discussion that we'd especially appreciate your eye on: https://github.com/tidymodels/parsnip/pull/732#discussion_r880971596

R/misc.R
juliasilge: Am I just missing seeing a test like this?

simonpcouch: Oh, woops—somehow missed this. Added in [51ffc03](https://github.com/tidymodels/parsnip/pull/732/commits/51ffc0371f3f9d398a2a27733202b5131f4b7073). :)

juliasilge (approved):

R/misc.R
```diff
@@ -79,6 +79,78 @@ model_printer <- function(x, ...) {
 is_missing_arg <- function(x)
   identical(x, quote(missing_arg()))
 
+model_info_table <-
+  utils::read.delim(system.file("models.tsv", package = "parsnip"))
```
topepo, line 83: Can you do the same without using this file? The parsnip model env is the best representation of when packages are required for an engine.

simonpcouch: @topepo I'm not sure we can.

Could you confirm that the needed extension package is not reflected in the model environment if that package is not loaded? e.g.

``` r
library(parsnip)
  
m_env <- parsnip::get_model_env()
m_env$C5_rules_pkgs
#> # A tibble: 0 × 3
#> # … with 3 variables: engine <chr>, pkg <list>, mode <chr>
m_env$C5_rules
#> # A tibble: 0 × 2
#> # … with 2 variables: engine <chr>, mode <chr>

C5_rules()
#> parsnip could not locate an implementation for `C5_rules` classification model specifications using the `C5.0` engine.
#> ℹ The parsnip extension package rules implements support for this specification.
#> ℹ Please install (if needed) and load to continue.
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0

library(rules)

m_env_ <- parsnip::get_model_env()
m_env_$C5_rules_pkgs
#> # A tibble: 1 × 3
#>   engine pkg       mode          
#>   <chr>  <list>    <chr>         
#> 1 C5.0   <chr [2]> classification

C5_rules()
#> C5.0 Model Specification (classification)
#> 
#> Computational engine: C5.0
```

<sup>Created on 2022-05-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Using the model environment instead wouldn't supply the information we need, I think—the purpose of this prompt is to nudge users to load a specific, needed extension package into their model environment if we can identify which one they need.

simonpcouch: (Discussed in meeting—thumbs up from Max here.)

simonpcouch:
Note from meeting: thumbs up from Max to merge!

PR #734, opened 2022-05-25
document `mtry` and `counts`, add `?details_boost_tree_lightgbm` doc

simonpcouch:
Also, updates `models.tsv` with new dev bonsai (dropping the censored regression entry).


man/rmd/template-mtry-prop.Rmd
```diff
@@ -0,0 +1,7 @@
+The `mtry` argument denotes the number of predictors that will be randomly sampled at each split when creating tree models.
```
simonpcouch, line 1: Note for review: this is the main body of changes, and many diffs reflect incorporating and re`document`ing with these lines.

EmilHvitfeldt: Thanks for the hint!

EmilHvitfeldt (approved):
I like it!

PR #735, opened 2022-05-26
Name translations

topepo:
Developer function to list names of tuning parameters

``` r
library(parsnip)
spec <- boost_tree(trees = tune("number of trees"), min_n = tune(), tree_depth = 3)

.model_param_name_key(spec)
#> # A tibble: 2 × 3
#>   user            parsnip engine          
#>   <chr>           <chr>   <chr>           
#> 1 number of trees trees   nrounds         
#> 2 min_n           min_n   min_child_weight

# A version to use with dplyr::rename()
.model_param_name_key(spec, as_tibble = FALSE)
#> $user_to_parsnip
#>             trees             min_n 
#> "number of trees"           "min_n" 
#> 
#> $parsnip_to_engine
#>          nrounds min_child_weight 
#>          "trees"          "min_n"
```

<sup>Created on 2022-05-26 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


juliasilge:
The function this PR creates isn't type stable, which we generally try to avoid. Would it be better to make two functions, `.model_param_name_key_dfr()` (or similar) and `.model_param_name_key_list()`?

juliasilge (commented):
The function this PR creates isn't type stable, which we generally try to avoid. Would it be better to make two functions, `.model_param_name_key_dfr()` (or similar) and `.model_param_name_key_list()`?

topepo:
The default value is type stable and I'm fine with that.

Issue #567, opened 2021-09-29
Should there be a helper function for print methods


EmilHvitfeldt:
All the print methods follow this pattern

https://github.com/tidymodels/parsnip/blob/af46f571c4e7ea07d7a707bc1a12e3567ff93e5e/R/boost_tree.R#L86-L95

What about we create a wrapper so we can do

```r
print_helper <- function(name, x, ...) {
  cat(name, " Specification (", x$mode, ")\n\n", sep = "")
  model_printer(x, ...)
  
  if (!is.null(x$method$fit$args)) {
    cat("Model fit template:\n")
    print(show_call(x))
  }
  invisible(x)
}

print.decision_tree <- function(x, ...) {
  print_helper("Decision Tree Model", x, ...)
}
```

juliasilge:
This would make it way easier to start using cli eventually, for example.

EmilHvitfeldt:
The null model print is substantially different and should be put back in line with the rest

https://github.com/tidymodels/parsnip/blob/27d5b6b17a1a89ab058ed793a005b7eaee0a6d0e/R/nullmodel.R#L79-L95

And in its current state, the print method doesn't print that it is a null model

``` r
library(parsnip)
null_model()
#> Model Specification (classification)
```

simonpcouch:
Ah, wasn't automatically closed with #739—doing so now. :)

PR #739, opened 2022-05-31
refactor model specification printing

simonpcouch:
While working on some previous PRs, I've noticed that all but one or two of our model spec print methods take the form:

```
print.bag_mars <- function(x, ...) {
  cat("Bagged MARS Model Specification (", x$mode, ")\n\n", sep = "")
  model_printer(x, ...)

  if (is_printable_spec(x)) {
    cat("Model fit template:\n")
    print(show_call(x))
  }
  invisible(x)
}
```

where the only line that changes method-to-method is the `cat` header summarizing the model type.

This PR refactors model specification print methods to all dispatch to `print.model_spec`, which now makes use of a `print_model_spec` helper that subsets a tibble giving the descriptions for each class to change that model type description in the header. Hopefully we find this to be more maintainable.

This change should be non-breaking for dependencies. 

* If they define engines for a model type that is already supported in parsnip and rely on parsnip's print methods, we've tested that the output doesn't change. (Exceptions noted in 135a544—`"Model "` is added preceding `"Specification"` in the header in 4 print methods.)
* If they supply their own print methods, `print.model_spec` won't be dispatched to.

This PR still exports the new `print_model_spec` helper, though, in case dependencies want to make use of it. That code would look like:

```
print_model_spec(new_type_object, desc = "New Type Description")
```

-----------

This PR will be hard to review as a whole (lots of deletions across many files), so I tried to commit in such a way to make reviewing commit-by-commit much easier. :)


hfrick (commented):
Nice one! Thanks for the commit order, that made it a lot easier indeed :raised_hands: 

Emil's comment https://github.com/tidymodels/parsnip/issues/567#issuecomment-932488860 about the print method for the null model is still open. If we want this PR to close #567, we should address this or turn it into its own issue so that it doesn't get lost.

simonpcouch:
Ah, thanks for pointing that out.

It looks like the output of `null_model` doesn't actually dispatch to the print method mentioned in #567. All of the methods for null models are defined for class `nullmodel`, but the output of `null_model` has first class `null_model`. With the current PR (that doesn't adjust null model internals):

``` r
library(parsnip)

null_mod <- null_model()

class(null_mod)
#> [1] "null_model" "model_spec"

methods(class = "null_model")
#> no methods found
methods(class = "nullmodel")
#> [1] predict print   tidy   
#> see '?methods' for accessing help and source code

# dispatches to new method :(
null_mod
#> null model Model Specification (classification)
```

<sup>Created on 2022-06-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

I could `print.null_model <- print.nullmodel` here, but I think a fresh issue re: null model dispatch is probably more appropriate.

hfrick (approved):

simonpcouch:
Thanks for the review, @hfrick! :) Ready to merge from our perspective, @topepo.

PR #741, opened 2022-06-01
remove test_by_col function and its use

EmilHvitfeldt:
The function `test_by_col()` was added in https://github.com/tidymodels/parsnip/commit/4b0808f9f5270f06a6d5e8e7d5162dd843ef9852 a little over 3 years ago. It should be fine to use `expect_equal()` instead.


hfrick (approved):
🧹 ✨

Issue #745, opened 2022-06-03
find_tune_id() errors if expression is namespaced


EmilHvitfeldt:
This bug was extracted from the following this SO question https://stackoverflow.com/questions/72482220/r-tidymodels-error-when-using-a-glm-model-within-a-workflow-set. Short-term fix would be to not namespace arguments or create object outside and pass that in. Long term, this should be fixed since disallowing name-spacing seems bad

``` r
library(parsnip)

ggg <- stats::gaussian("log")

glm_model_ns <- linear_reg() %>%
   set_engine("glm", family = stats::gaussian("log"))

tune::tune_args(glm_model_ns)
#> Error in if (rlang::call_name(x) == "tune") {: argument is of length zero

ggg <- stats::gaussian("log")

glm_model_ns2 <- linear_reg() %>%
   set_engine("glm", family = ggg)

tune::tune_args(glm_model_ns2)
#> # A tibble: 0 × 6
#> # … with 6 variables: name <chr>, tunable <lgl>, id <chr>, source <chr>,
#> #   component <chr>, component_id <chr>

glm_model <- linear_reg() %>%
   set_engine("glm", family = gaussian("log"))

tune::tune_args(glm_model)
#> # A tibble: 0 × 6
#> # … with 6 variables: name <chr>, tunable <lgl>, id <chr>, source <chr>,
#> #   component <chr>, component_id <chr>

glm_model$eng_args
#> <list_of<quosure>>
#> 
#> $family
#> <quosure>
#> expr: ^gaussian("log")
#> env:  0x127046da0

parsnip:::find_tune_id(glm_model$eng_args[[1]])
#> [1] NA

glm_model_ns$eng_args
#> <list_of<quosure>>
#> 
#> $family
#> <quosure>
#> expr: ^stats::gaussian("log")
#> env:  0x1137460c0

parsnip:::find_tune_id(glm_model_ns$eng_args[[1]])
#> Error in if (rlang::call_name(x) == "tune") {: argument is of length zero
```

<sup>Created on 2022-06-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #746, opened 2022-06-03
use rlang::is_call_simple

EmilHvitfeldt:
This PR aims to close https://github.com/tidymodels/parsnip/issues/745.

Before, what used to happen when a namespaced call came in, `rlang::call_name(x)` would return `NULL` and the if statement would break. This is happening because `bar::foo` is a complex name, see [rlang::call_name()](https://rlang.r-lib.org/reference/call_name.html).

``` r
library(parsnip)

glm_model_ns <- linear_reg() %>%
  set_engine("glm", family = stats::gaussian(tune("hello")))

glm_model_ns$eng_args
#> <list_of<quosure>>
#> 
#> $family
#> <quosure>
#> expr: ^stats::gaussian(tune("hello"))
#> env:  0x1177b49c8

parsnip:::find_tune_id(glm_model_ns$eng_args[[1]])
#> [1] "hello"

sg_call <- rlang::expr(stats::gaussian)

is.call(sg_call)
#> [1] TRUE

rlang::call_name(sg_call)
#> NULL

rlang::is_call_simple(sg_call)
#> [1] FALSE
```

This change doesn't break anything (that I found) by running extratests. It works even if `tune()` is namespaced

``` r
library(parsnip)

glm_model_ns <- linear_reg() %>%
  set_engine("glm", family = stats::gaussian(tune::tune("hello")))

parsnip:::find_tune_id(glm_model_ns$eng_args[[1]])
#> [1] "hello"
```

<sup>Created on 2022-06-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>


NEWS.md
```diff
@@ -14,6 +14,8 @@
 
 * Model type functions will now message informatively if a needed parsnip extension package is not loaded (#731).
 
+* Fixed bug where en error would be thrown if arguments to model functions were namespaced and find_tune_id() were used (#745).
```
DavisVaughan, line 17:
```suggestion
* Fixed a bug where an error would be thrown if arguments to model functions were namespaced (#745).
```

I don't think you need to mention the internal function in the news bullet. The PR link points to all the developer details

DavisVaughan (approved):
Makes sense to me after reading the `call_name()` docs. Seems smart to guard those calls with `is_call_simple()` if you need to require that the result is a single TRUE/FALSE, as we do here

PR #748, opened 2022-06-06
Engine docs for h2o

qiushiyan:


simonpcouch:
You may want to add "agua" to this vector:

https://github.com/tidymodels/parsnip/blob/78e8e16f822f74efaefef2b1d492c64f48979d09/R/engine_docs.R#L50-L53

 and re-document so that parsnip knows where the h2o support lives and can document it in `models.tsv`. :)

qiushiyan:
> 

Max already had it in https://github.com/tidymodels/parsnip/commit/1b3da0114a32608c768685cff161cfa1c9ae1f99. I have updated `inst/models.tsv` as well, thanks.

inst/models.tsv
```diff
@@ -9,9 +9,11 @@
 "bart"	"regression"	"dbarts"	NA
 "boost_tree"	"censored regression"	"mboost"	"censored"
 "boost_tree"	"classification"	"C5.0"	NA
+"boost_tree"	"classification"	"h2o"	NA
```
simonpcouch, line 12: Ah, I see. The `pkg` entry is NA for these because a duplicated `set_dependency` is needed in your model registration code in agua.

See [here](https://github.com/tidymodels/bonsai/blob/main/R/lightgbm_data.R#L16-L28) for an example—where you have `"h2o"` `"h2o"` in calls to `set_dependency` (like [here](https://github.com/tidymodels/agua/blob/main/R/model_engines.R#L102)), you also need an `"h2o"` `"agua"`.

Once that's situated, re-installing agua and then re-running `parsnip::update_model_info_file()` should update those missing values accordingly. :)

qiushiyan: Thanks! Fixed it in agua.

simonpcouch: Sweet, bravo!🌊

man/rmd/boost_tree_h2o.Rmd
```diff
@@ -0,0 +1,74 @@
...
+boost_tree(
+  mtry = integer(), trees = integer(), tree_depth = integer(), 
+  learn_rate = numeric(), min_n = integer(), loss_reduction = numeric(), stop_iter = integer()
+) %>% 
+  set_engine("h2o") %>% 
+  set_mode("classification") %>% 
+  translate()
+```
+
+## Other details
```
topepo, line 61: You can make an Rmd template that that has something like

```
## Initializing h2o

To use h2o with tidymodels, please run `h2o::h2o.init()` (blah blah blah)

To stop using h2o, run `h2o::h2o.shutdown()`
```

and include the template in the h2o engine files. 

We might also mention there that you can do multiple threads (by default) in `h2o.init()` but the usually parallel processing with tidymodels also works and that it will decrease tuning time.

man/rmd/mlp_h2o.Rmd
```diff
@@ -0,0 +1,84 @@
...
+  make_parameter_list(defaults)
+```
+
+This model has `r nrow(param)` tuning parameters:
+
+```{r h2o-param-list, echo = FALSE, results = "asis"}
+param$item
+```
+
+The naming of activation functions in [h2o::h2o.deeplearning()] differs from parsnip's conventions. Currently, only "relu" and "tanh" are supported and will be converted internally to "Rectifier" and "Tanh" passed to the fitting function.
```
topepo, line 25: 👍 on these parts

man/rmd/mlp_h2o.Rmd
```diff
@@ -0,0 +1,84 @@
...
+
+`penalty` corresponds to l2 penalty. [h2o::h2o.deeplearning()] also supports specifying the l1 penalty directly with the engine argument `l1`. 
+
+Other engine arguments of interest: 
+
+- `stopping_rounds` controls early stopping rounds based on the convergence of another engine parameter `stopping_metric`. By default, [h2o::h2o.deeplearning] stops training if simple moving average of length 5 of the stopping_metric does not improve for 5 scoring events.  This is mostly useful when used alongside the engine parameter `validation`, which is the __proportion__ of train-validation split, parsnip will split and pass the two data frames to h2o. Then [h2o::h2o.deeplearning] will evaluate the metric and early stopping criteria on the validation set. 
+
+- h2o uses a 50% dropout ratio controlled by `dropout` for hidden layers by default. [h2o::h2o.deeplearning()] provides an engine argument `input_dropout_ratio` for dropout ratios in the input layer, which defaults to 0.
```
topepo, line 33: I did not know this. Good catch

topepo (changes requested):
Looks great. One suggestion below. 

In the aqua PR, I mentioned making an engine for `poisson_reg()` so make sure to add that file here too.

man/rmd/boost_tree_h2o.Rmd
qiushiyan: template added [man/rmd/template-h2o-init.Rmd](https://github.com/tidymodels/parsnip/pull/748/files#diff-a383b8c5aa4b72f499c26045daeec384842cd337b923d6f08b1b19a42c4030ec)

PR #750, opened 2022-06-07
use `pillar::type_sum` rather than `tibble::type_sum`

### LINKED ISSUE #733: Use `pillar::type_sum()` rather than `tibble::type_sum()`
We'd like to _eventually_ remove `tibble::type_sum()`, which just reexports `pillar::type_sum()`. To do this, we need to switch over to using the pillar version.

We import the tibble version from here, we can just change this to importing the pillar version.
https://github.com/tidymodels/parsnip/blob/edf88ab9407e73013433aa5c09781ef005d7b459/R/parsnip-package.R#L24

simonpcouch:
Closes #733. :)


DavisVaughan (approved):
Looks great, thanks!

PR #767, opened 2022-07-13
address GHA failures

simonpcouch:
Attempting to address GHA failures. 

At the moment, looking like pak "install planning" failures. Looks like there's been a release of reticulate since we started using the dev version, so trying to install the CRAN version first.


simonpcouch:
Seems like this may have fixed it!

Tagging @EmilHvitfeldt as I want to make sure this doesn't "undo" any of these old-tensorglow checks that yall discussed in #639 and related threads. From what I can tell, this install always reflected "most recent version," so I think we're in a good spot.

PR #769, opened 2022-07-15
Make userfacing set_* functions generic

EmilHvitfeldt:
This PR makes the functions `set_engine()`, `set_mode()` and `set_args()` generic.

This is being done so [tidyclust](https://github.com/EmilHvitfeldt/tidyclust) can use these functions as well.


R/engines.R
```diff
@@ -105,10 +105,12 @@ load_libs <- function(x, quiet, attach = FALSE) {
 #'
 #' @export
 set_engine <- function(object, engine, ...) {
+  UseMethod("set_engine")
+}
+
+#' @export
+set_engine.model_spec <- function(object, engine, ...) {
   mod_type <- class(object)[1]
-  if (!inherits(object, "model_spec")) {
```
simonpcouch, line 109: Did we used to test these lines anywhere? Surprised there aren't any existing snaps or checks that needed adjusting.

simonpcouch (approved):
Looks great to me besides that one question about testing. :)

R/engines.R
EmilHvitfeldt: No prior tests: https://app.codecov.io/gh/tidymodels/parsnip/blob/master/R/engines.R
Screen Shot 2022-07-15 at 9 21 31 AM

simonpcouch: Gotcha, sounds good👍

PR #772, opened 2022-07-20
address dev dplyr join warnings

simonpcouch:
Related to, but doesn't fully fix, tidymodels/tune#526.

``` r
# devtools::install_github("tidyverse/dplyr")

library(tidymodels)

workflow(
  preprocessor = mpg ~ .,
  spec = linear_reg(engine = 'glmnet', penalty = tune(), mixture = 1)
) %>% 
  tune_grid(resamples = vfold_cv(mtcars))
#> # Tuning results
#> # 10-fold cross-validation 
#> # A tibble: 10 × 4
#>    splits         id     .metrics          .notes          
#>    <list>         <chr>  <list>            <list>          
#>  1 <split [28/4]> Fold01 <tibble [20 × 5]> <tibble [0 × 3]>
#>  2 <split [28/4]> Fold02 <tibble [20 × 5]> <tibble [0 × 3]>
#>  3 <split [29/3]> Fold03 <tibble [20 × 5]> <tibble [0 × 3]>
#>  4 <split [29/3]> Fold04 <tibble [20 × 5]> <tibble [0 × 3]>
#>  5 <split [29/3]> Fold05 <tibble [20 × 5]> <tibble [0 × 3]>
#>  6 <split [29/3]> Fold06 <tibble [20 × 5]> <tibble [0 × 3]>
#>  7 <split [29/3]> Fold07 <tibble [20 × 5]> <tibble [0 × 3]>
#>  8 <split [29/3]> Fold08 <tibble [20 × 5]> <tibble [0 × 3]>
#>  9 <split [29/3]> Fold09 <tibble [20 × 5]> <tibble [0 × 3]>
#> 10 <split [29/3]> Fold10 <tibble [20 × 5]> <tibble [0 × 3]>
```

<sup>Created on 2022-07-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

If this fix looks like the right move, we'll need to apply in [at least a few](https://github.com/search?q=org%3Atidymodels+param_key&type=code) other places, likely worth testing all `multi_predict` methods.

### DIFF
--- a/R/linear_reg.R
+++ b/R/linear_reg.R
@@ -243,7 +243,11 @@ multi_predict._elnet <-
     pred <- as_tibble(pred)
     pred$.row <- 1:nrow(pred)
     pred <- gather(pred, group, .pred, -.row)
-    pred <- full_join(param_key, pred, by = "group", multiple = "all")
+    if (utils::packageVersion("dplyr") >= "1.0.99.9000") {
+      pred <- full_join(param_key, pred, by = "group", multiple = "all")
+    } else {
+      pred <- full_join(param_key, pred, by = "group")
+    }
     pred$group <- NULL
     pred <- arrange(pred, .row, penalty)
     .row <- pred$.row
--- a/R/linear_reg.R
+++ b/R/linear_reg.R
@@ -243,7 +243,7 @@ multi_predict._elnet <-
     pred <- as_tibble(pred)
     pred$.row <- 1:nrow(pred)
     pred <- gather(pred, group, .pred, -.row)
-    pred <- full_join(param_key, pred, by = "group")
+    pred <- full_join(param_key, pred, by = "group", multiple = "all")
     pred$group <- NULL
     pred <- arrange(pred, .row, penalty)
     .row <- pred$.row
--- a/R/logistic_reg.R
+++ b/R/logistic_reg.R
@@ -296,7 +296,7 @@ multi_predict._lognet <-
         pred <- pred[, c(".row", "group", paste0(".pred_", object$lvl))]
       }
     }
-    pred <- full_join(param_key, pred, by = "group")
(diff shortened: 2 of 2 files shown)


simonpcouch:
Seems like CRAN dplyr doesn't trip up on the new `multiple` argument being passed through dots, so we can possibly just pass it along without requiring a specific version or passing the argument conditional on version.

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union

df1 <- tibble(x = 1:3)
df2 <- tibble(x = c(1, 1, 2), y = c("first", "second", "third"))
df1 %>% left_join(df2, multiple = "all")
#> Joining, by = "x"
#> # A tibble: 4 × 2
#>       x y     
#>   <dbl> <chr> 
#> 1     1 first 
#> 2     1 second
#> 3     2 third 
#> 4     3 <NA>
```

<sup>Created on 2022-07-20 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

DavisVaughan:
> CRAN dplyr doesn't trip up on the new multiple argument being passed through dots

Yea this is a bug, it really should yell at you for silently passing something through the dots that doesn't get used. The next version of dplyr will make this an error.

I dont particularly want to rely on this behavior, so I think we should go the route of doing something that is conditional on the version number (you can use `>= "1.0.99.9000"` as the version, I bumped to 99 when the join stuff was added).

simonpcouch:
Okay, fair enough. :) I'll do an "audit" of `multi_predict` methods today and PR as needed with conditional logic like you describe.

DavisVaughan (approved):

hfrick (approved):

PR #778, opened 2022-08-01
code style in checking internals

qiushiyan:
When `mode == 'unknown'`, `inform_missing_implementation` searches available models for all modes. 

Also had to use `new_model_spec(check_missing_spec = FALSE)` in `boost_tree` (and perhaps other models that involve extension packages) and  `new_model_spec(check_missing_spec = TRUE)` in `set_engine.model_spec()`, such that inform doesn't use the default engine

```
# before: parsnip could not locate an implementation for bag_tree model specifications using the `rpart` engine ... 
boost_tree() %>% 
  set_engine("C5.0")
```

Current error messages 

``` r
library(parsnip)

bag_tree() %>%
  set_engine("rpart") %>%
  testthat::capture_message()
#> <message/rlang_message>
#> Message:
#> ℹ parsnip could not locate an implementation for bag_tree model specifications using the `rpart` engine. The following parsnip extension packages implement support for this specification.
#> ℹ censored, baguette
#> ℹ Please install one of them (if needed) and load to continue.

bag_tree() %>% 
  set_mode("censored regression") %>% 
  set_engine("rpart") %>% 
  testthat::capture_message()
#> <message/rlang_message>
#> Message:
#> ℹ parsnip could not locate an implementation for bag_tree censored regression model specifications using the `rpart` engine. Please install `censored` (if needed) and load to continue.

bag_tree() %>% 
  set_engine("C5.0") %>% 
  testthat::capture_message()
#> <message/rlang_message>
#> Message:
#> ℹ parsnip could not locate an implementation for bag_tree model specifications using the `C5.0` engine. Please install `baguette` (if needed) and load to continue.

library(censored)

bag_tree() %>% 
  set_mode("censored regression") %>% 
  set_engine("rpart") %>% 
  testthat::capture_message()
#> NULL

library(baguette)
bag_tree() %>% 
  set_engine("C5.0") %>% 
  testthat::capture_message()
#> NULL
```

<sup>Created on 2022-08-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>
One remaining problem being if `set_mode` is called after `set_engine` then the user specified mode won't be reflected in the error message. If `set_mode` uses `new_model_spec(check_missing_arg = TRUE)`  as well then we get two sets of messages.


simonpcouch:
Thanks for making this happen, @qiushiyan! Feel free to mark this PR as ready for review when you feel good about it and I'll take a look. I had wondered if there were any conditions that this might happen when I put this together, and it looks like I might have missed one. The argument checking across model spec functions is indeed tricky. :)

qiushiyan:
Now compute inform message at print time in `print_model_spec()` with davis suggestions 🎉 . I think changes in `fit()` need more thoughts could be in a separte PR. @simonpcouch could you take a look and take this over?

simonpcouch:
Awesome, thanks for making this happen @qiushiyan. Will review and move this along next week. :)

R/engines.R
```diff
@@ -116,7 +116,6 @@ set_engine.model_spec <- function(object, engine, ...) {
     stop_missing_engine(mod_type)
   }
   object$engine <- engine
-  check_spec_mode_engine_val(mod_type, object$engine, object$mode)
```
simonpcouch, line 119: This is a larger piece of machinery than what what was added when `inform_missing_implementation` was added.

R/misc.R
```diff
@@ -67,31 +73,47 @@ is_printable_spec <- function(x) {
...
   if (nrow(avail) == 0 && nrow(all) > 0) {
-    msg <-
-      c(
-        msg,
-        i = paste0("The parsnip extension package ", all$pkg[[1]],
-                   " implements support for this specification."),
-        i = "Please install (if needed) and load to continue.",
-        ""
+    pkgs <- unique(all$pkg)
+    if (length(pkgs) == 1) {
```
simonpcouch, line 102: This conditional can be collapsed into one with use `cli` pluralization.

simonpcouch (commented):
Mostly reviewing for myself to point out notable changes.

Lots of helpful syntax consistency changes here, though the diffs are a bit difficult to navigate because of them.

Otherwise, as is, relocates `check_missing_spec` from `new_model_spec` to `print.model_spec`, shortens the missing extension package message, and makes needed adjustments! 

From here, as Qiushi mentioned, need to figure out actions that indicate defining a model spec is "finished" so we know we can check the `mode` vs `engine` vs `class`. `fit` is one of those places.

simonpcouch:
Splitting this PR into two, one for code style edits and one to improve messaging for missing implementations. This one is now the former, so will go ahead and merge!

PR #789, opened 2022-08-15
error informatively with bad `object` in `set_*` functions

### LINKED ISSUE #776: need a better error message for some parsnip functions
```r
> bag_tree %>% set_mode("classification")
Error in `check_spec_mode_engine_val()` at parsnip/R/arguments.R:83:2:
! 'classification' is not a known mode for model `function()`.
```

We should be able to say "you might need to load packages x, y, z for this engine" so some other information

simonpcouch:
Closes #776.

Worth noting that this error message has changed since the issue was filed, following up on #769.

``` r
library(parsnip)

bag_tree %>% set_mode("classification")
#> Error in UseMethod("set_mode"): no applicable method for 'set_mode' applied to an object of class "function"
```

<sup>Created on 2022-08-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

The errors now look like:

``` r
library(parsnip)

bag_tree %>% set_mode("classification")
#> Error in `set_mode()`:
#> ! `set_mode()` expected a model specification to be supplied to the
#>   `object` argument, but received a(n) `function` object.
#> ℹ Did you mistakenly pass `model_function` rather than `model_function()`?

bag_tree %>% set_engine("rpart")
#> Error in `set_engine()`:
#> ! `set_engine()` expected a model specification to be supplied to the
#>   `object` argument, but received a(n) `function` object.
#> ℹ Did you mistakenly pass `model_function` rather than `model_function()`?

# won't raise "info" part of error if not a parsnip-namespaced function
# not a function
1L %>% set_args(mode = "classification")
#> Error in `set_args()`:
#> ! `set_args()` expected a model specification to be supplied to the
#>   `object` argument, but received a(n) `integer` object.

# not from parsnip
cos %>% set_mode("classification")
#> Error in `set_mode()`:
#> ! `set_mode()` expected a model specification to be supplied to the
#>   `object` argument, but received a(n) `function` object.
```

<sup>Created on 2022-08-15 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Note that, in the case where a user supplies a parsnip-namespaced function that's not a model type function, the error message is misleading. If we wanted to make the logic for that part of the error more exact, we could class the model type functions with a subclass!


simonpcouch:
Requesting your review here, @EmilHvitfeldt, since this messaging is relevant to tidyclust, but no worries on reviewing this before the end of your vacation! These changes won't be merged until after the upcoming release of parsnip anyway, so reviewing in September is great. :)

I use "model specification" in the prompt rather than a specific object class so that this will hopefully generalize well to tidyclust. Feel free to suggest improvements if you think there's a more helpful way of phrasing this, though.

R/aaa_models.R
```diff
@@ -92,6 +92,22 @@ set_env_val <- function(name, value) {
 
 # ------------------------------------------------------------------------------
 
+error_set_object <- function(object, func) {
+  msg <-
+    "`{func}()` expected a model specification to be supplied to the \
+     `object` argument, but received a(n) `{class(object)[1]}` object."
+
+  if (inherits(object, "function") &&
+      isTRUE(environment(object)$.packageName == "parsnip")) {
```
EmilHvitfeldt, line 101: I like the general idea of nudging the user if they forgot the `()`. However this approach currently only works with parsnip. Not tidyclust, or any of the parsnip extensions. What is the rational of not having this behavior for any function, instead of just parsnip functions?

EmilHvitfeldt (commented):
Overall good implementation. 1 question in review

R/aaa_models.R
simonpcouch: > Not tidyclust, or any of the parsnip extensions.

Indeed for tidyclust, though any extension with model _type_ defined in parsnip (censored, bonsai, baguette, multilevelmod, poissonreg, most of agua) will have this part of the message included.

> What is the rational of not having this behavior for any function, instead of just parsnip functions?

I feel that a false positive (explicitly suggest supplying function output when the function is, say, `cos()`) is worse than a false negative (just give the first part of the message, saying that it received a function rather than a model spec) in this case.

EmilHvitfeldt: Good point. then it makes more sense, briefly forgot that we add some of model definitions in parsnip.

EmilHvitfeldt: Over time it might be worth it to add modeltime, and tidyclust to that list

simonpcouch:
Looping in @topepo before merging here. See Emil's comment above—does it make more sense to just drop the second bit of the prompt entirely? Or should we think about subclassing the model spec functions so that this logic is solid?

topepo:
I agree to leave it as-is (until we see it as a big issue)

PR #790, opened 2022-08-15
address knitr installation dependency

### LINKED ISSUE #785: knitr is an installation dependency under certain conditions
## The problem

Under certain conditions, installation of `parsnip` fails with the following error:

```
Error installing package 'parsnip':
===================================

* installing *source* package 'parsnip' ...
** package 'parsnip' successfully unpacked and MD5 sums checked
** using staged installation
** R
** data
*** moving datasets to lazyload DB
** inst
** byte-compile and prepare package for lazy loading
** help
*** installing help indices
  converting help for package 'parsnip'
    finding HTML links ... done
    C5.0_train                              html  
    C5_rules                                html  
    add_on_exports                          html  
    add_rowindex                            html  
    augment                                 html  
    autoplot.model_fit                      html  
    bag_mars                                html  
Error: bag_mars.Rd:36: there is no package called 'knitr'
* removing '/opt/R/4.2.1/lib/R/library/.renv/1/parsnip'
Error: install of package 'parsnip' failed [error code 1]
Traceback (most recent calls last):
12: renv::restore()
11: renv_restore_run_actions(project, diff, current, lockfile, rebuild)
10: renv_install_impl(records)
 9: renv_install_staged(records)
 8: renv_install_default(records)
 7: handler(package, renv_install_package(record))
 6: renv_install_package(record)
 5: withCallingHandlers(renv_install_package_impl(record), error = function(e) {
        vwritef("\tFAILED")
        writef(e$output)
    })
 4: renv_install_package_impl(record)
 3: r_cmd_install(package, path)
 2: r_exec_error(package, output, "install", status)
 1: stop(error)
Execution halted
executor failed running [/bin/sh -c Rscript -e "renv::restore()"]: exit code: 1
```

The error occurs for me when I install earth then parsnip, but I can only reproduce the error on CentOS 7. The error doesn't occur in Debian. If knitr is installed then the error goes away, of course, but I discovered this issue when using vetiver and renv, both of which have no way to automatically pick up on knitr as a dependency.

I believe the actual cause is as follows:

* install earth, and then attempt to install parsnip from source
* R attempts to convert the `.Rd` files into help files
* [The following line is called during this conversion](https://github.com/tidymodels/parsnip/blob/f0810910d0b69141af12c66ae210ec0880f95323/man/bag_mars.Rd#L36):
```tex
\Sexpr[stage=render,results=rd]{parsnip:::make_engine_list("bag_mars")}
```
* `parsnip:::make_engine_list("bag_mars")` picks up on the "earth" engine, and attempts to call `knitr::combine_words`.
* knitr is only a _suggested_ dependency of parsnip. If it's not already installed, R will error here.

## Reproducible example

I'm afraid this reprex requires a Dockerfile. [I've put up a minimal repository](https://github.com/mdneuzerling/parsnip-docker-build-error). The renv snapshot is produced by the below line, which is very similar to what vetiver would do:

```
renv::snapshot(packages = c("earth", "parsnip", "workflows"), prompt = FALSE)
```

Attempting to build the Docker image will error with the above, although because we're building from source the process may take 20 minutes. The alternative Debian-based Dockerfile doesn't error, and I just can't understand why.

## Possible solutions

`knitr::combine_words` is a very small and self-contained component of knitr. Could this function be copied into parsnip? 

I would be happy to contribute a PR, if you think this is a valid issue.

simonpcouch:
Copies `knitr::combine_words` and friends to close #785.

Could you confirm this does the trick for you, @mdneuzerling?


mdneuzerling:
Thank you so much for the quick response, @simonpcouch! My Docker image now builds successfully — without knitr — so I think we can consider #785 solved.

simonpcouch:
Awesome, glad to hear it!

hfrick (approved):

PR #791, opened 2022-08-16
Tests for survival models

hfrick:
This tests the error message for trying to use `proportional_hazards()` with the `glmnet` engine _without_ a penalty value because that's the thing the `translate()` method for PH models does in addition to the default translate() method.

The rest of the tests for `proportional_hazards()` and `survival_reg()` is more of a placeholder because parsnip does not have any engines for these models. The actual tests on primary and engine args for those models are in censored, see https://github.com/tidymodels/censored/pull/208.

### DIFF
--- a/tests/testthat/_snaps/translate.md
+++ b/tests/testthat/_snaps/translate.md
@@ -1895,6 +1895,13 @@
       env:  empty
       
 
+# arguments (survival_reg)
+
+    Code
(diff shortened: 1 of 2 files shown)


simonpcouch (approved):
This looks great! I think these examples are good rules of thumb for testing model types that don't have any engines implemented in parsnip itself. Green light from me on merging. :)

PR #792, opened 2022-08-17
Strip Unicode characters from engine docs

topepo:


topepo:
This uses pillar 1.7.0 to get around some of the issues. For example: r-lib/pillar#575

PR #793, opened 2022-08-17
improve prompting about missing implementations

simonpcouch:
Follows up on #732 and #778. The goal here is to tighten the logic for prompts about missing implementations. This PR eliminates false positives and makes sure we always suggest "registered" extension packages when they could be relevant.

To do so, we defer model specification checking for type / engine / mode combinations until we have some sort of clue that the user might be "done" specifying the model (If only we could `<-.model_spec`😭). This is at:

* `print`
* `fit`

...and functions that use either downstream. Some examples of how these prompts used to be raised:

``` r
library(parsnip)

# errors with a default engine, though doesn't call out baguette
bag_tree()
#> parsnip could not locate an implementation for `bag_tree` model specifications
#> using the `rpart` engine.
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: rpart

# bad: still includes default engine in error, bc triggered in bag_tree(),
# and still does not point to baguette
bag_tree() %>%
  set_engine("C5.0")
#> parsnip could not locate an implementation for `bag_tree` model specifications
#> using the `rpart` engine.
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: C5.0

# this should call out "censored regression" specifically, but does
# not because the error is triggered in bag_tree().
# also does not point to censored
bag_tree() %>%
  set_engine("rpart") %>%
  set_mode("censored regression")
#> parsnip could not locate an implementation for `bag_tree` model specifications
#> using the `rpart` engine.
#> Bagged Decision Tree Model Specification (censored regression)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: rpart

# this is well-defined:
library(censored)

bag_tree() %>% 
  set_mode("censored regression") %>% 
  set_engine("rpart")
#> Bagged Decision Tree Model Specification (censored regression)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: rpart

# this errors differently when just censored is loaded
# ought to note that an implementation does live in baguette
bag_tree() %>%
  set_engine("C5.0")
#> Error in `check_spec_mode_engine_val()`:
#> ! Engine 'C5.0' is not supported for `bag_tree()`. See `show_engines('bag_tree')`.

# once baguette is loaded, this is well-defined:
library(baguette)

bag_tree() %>%
  set_engine("C5.0")
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: C5.0
```

<sup>Created on 2022-08-17 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

Same code with the dev version:
``` r
library(parsnip)

bag_tree()
#> ! parsnip could not locate an implementation for `bag_tree` model
#>   specifications.
#> ℹ The parsnip extension packages censored and baguette implement support for
#>   this specification.
#> ℹ Please install (if needed) and load to continue.
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: rpart

# now includes correct engine in error
bag_tree() %>%
  set_engine("C5.0")
#> ! parsnip could not locate an implementation for `bag_tree` model
#>   specifications using the `C5.0` engine.
#> ℹ The parsnip extension package baguette implements support for this
#>   specification.
#> ℹ Please install (if needed) and load to continue.
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: C5.0

# calls out "censored regression" specifically
bag_tree() %>%
  set_engine("rpart") %>%
  set_mode("censored regression")
#> ! parsnip could not locate an implementation for `bag_tree` censored regression
#>   model specifications using the `rpart` engine.
#> ℹ The parsnip extension package censored implements support for this
#>   specification.
#> ℹ Please install (if needed) and load to continue.
#> Bagged Decision Tree Model Specification (censored regression)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: rpart

# this is well-defined:
library(censored)

bag_tree() %>% 
  set_mode("censored regression") %>% 
  set_engine("rpart")
#> Bagged Decision Tree Model Specification (censored regression)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: rpart

# this now prompts informatively, as before censored was loaded
bag_tree() %>%
  set_engine("C5.0")
#> ! parsnip could not locate an implementation for `bag_tree` model
#>   specifications using the `C5.0` engine.
#> ℹ The parsnip extension package baguette implements support for this
#>   specification.
#> ℹ Please install (if needed) and load to continue.
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: C5.0

library(baguette)

# prompt goes away once baguette is loaded
bag_tree() %>%
  set_engine("C5.0")
#> Bagged Decision Tree Model Specification (unknown)
#> 
#> Main Arguments:
#>   cost_complexity = 0
#>   min_n = 2
#> 
#> Computational engine: C5.0
```

<sup>Created on 2022-08-30 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

This also catches bugs related to how loading one unrelated extension would affect the prompts related to another. PR to extratests to make sure we test this [here](https://github.com/tidymodels/extratests/pull/53)! There are also additional tests with decision trees there.

On `fit()`, this same message is elevated to an error.

Open to ideas if we have any other cues that a user may be "done" putting together a model spec—adding to a workflow is another instance, though we'll get "automatic" workflow support otherwise in that it wraps parsnip's print and fit methods. Same with [parameter extraction](https://community.rstudio.com/t/rule-fit-using-tidy-models-error/145913/2).


simonpcouch:
Note: [this community post](https://community.rstudio.com/t/rule-fit-using-tidy-models-error/145913) points out additional locations where we might want to drop these checks, none of which live in parsnip—`workflows::add_model` and `workflows::extract_parameter_set_dials.workflow`.  May be worth exporting helpers.

R/aaa_models.R
```diff
@@ -37,6 +37,11 @@ pred_types <-
 
 # ------------------------------------------------------------------------------
 
+model_info_table <-
```
topepo, line 40: This should probably be an internal function and not in open code. We only use it once (inside `check_spec_mode_engine_val()`) and the function should be invoked there.

R/misc.R
```diff
@@ -27,33 +27,78 @@ is_missing_arg <- function(x) {
...
+#
+# checks against the union of
+# 1) the current model environment and
+# 2) the model_info_table of "pre-registered" model types
+#
+# if spec \in 1, it will pass checks, and
+# if spec \in 2, it ought to bypass checks
+#    and defer to `prompt_missing_implementation` in case the
+#    the needed implementation is not yet installed/loaded
+implementation_exists_somewhere <- function(cls,
```
topepo, line 59: Informative name but excessive length. Can you rename?

topepo: Also, `cls` can be derived from `object` so maybe just have `object` and `user_specified_engine` as arguments?

R/bag_mars.R
```diff
@@ -37,8 +37,10 @@ bag_mars <-
       args = args,
       eng_args = NULL,
       mode = mode,
+      user_specified_mode = !missing(mode),
```
topepo, line 40: These changes will have to propagate to the external extension packages too (e.g. model time)?

We'll also need to update the docs on tidymodels.org too (I believe).

R/engines.R
```diff
@@ -116,7 +116,12 @@ set_engine.model_spec <- function(object, engine, ...) {
     stop_missing_engine(mod_type)
   }
   object$engine <- engine
-  check_spec_mode_engine_val(mod_type, object$engine, object$mode)
+
+  if (!implementation_exists_somewhere(mod_type,
```
topepo, line 120: Can you add some comments (here?) about the strategy we are using to check the mode?

R/fit.R
```diff
@@ -120,12 +120,14 @@ fit.model_spec <-
     dots <- quos(...)
 
     if (length(possible_engines(object)) == 0) {
-      rlang::abort(
-        glue::glue(
-          "No engines were loaded for `{class(object)[1]}`.",
-          " Please load a parsnip extension package that provides one.",
-          " See https://www.tidymodels.org/find/parsnip/ "
-        )
+      prompt_missing_implementation(
```
topepo, line 123: Named arguments please.

R/misc.R
```diff
@@ -27,33 +27,78 @@ is_missing_arg <- function(x) {
   identical(x, quote(missing_arg()))
 }
 
-model_info_table <-
-  utils::read.delim(system.file("models.tsv", package = "parsnip"))
+engine_filter_condition <- function(engine, user_specified_engine) {
```
topepo, line 30: It's hard to read this and know what it does. It is returning an expression (or `TRUE`) that is used in `filter()`?

R/misc.R
```diff
@@ -62,45 +107,59 @@ has_loaded_implementation <- function(spec_, engine_, mode_) {
...
-    mode_ <- ""
-  }
+  if (!user_specified_mode) {mode <- ""}
 
-  msg <-
-    glue::glue(
-      "parsnip could not locate an implementation for `{spec_}` {mode_} model \\
-       specifications using the `{engine_}` engine."
+  msg <- c(
+    "!" = glue::glue(
```
topepo, line 143: Can we put the package name in single quotes (or something else)? Due to the weird way I name packages, this can look like a typo or something.

topepo (changes requested):
I'll run this with TMwR and our workshop notes to test more. 

Can you run revdeps?

simonpcouch:
Thanks for the review! Running revdeps and addressing comments ATM.

R/bag_mars.R
simonpcouch: `engine_filter_condition` and `mode_filter_condition` (to be renamed re: comment), where these slots are processed, anticipate model objects from external extension packages not having this slot—in those cases, we bypass the additional prompts so that those external extensions fall back to the same errors as before. Those extensions can add the slot and get the prompting introduced in this PR "for free," since these prompts look to a full join of the model environment and `model_info_table`.

Good thought on the tidymodels.org docs. Will follow-up with a PR there.

simonpcouch:
No new problems in revdeps, though I found a backward-compatibility issue that wasn't caught.

R/misc.R
simonpcouch: Renamed to `spec_is_possible`! Correct that the arguments could be derived from a complete model object, but we use this to test a new "proposed" object in `set_mode` and `set_engine`. e.g. in `set_mode.model_spec`:

```
  if (!spec_is_possible(cls,
                        object$engine, object$user_specified_engine,
                        mode, user_specified_mode = TRUE))
```

where `mode` was supplied to `set_mode`.

Maybe we can set defaults based on `object` slots and then override them where needed?

R/aaa_models.R
simonpcouch: Implemented in [e99dfd5](https://github.com/tidymodels/parsnip/pull/793/commits/e99dfd55115ca4dc36098527faea2da1dd2a710f)!

R/misc.R
simonpcouch: Added comments in [e99dfd5](https://github.com/tidymodels/parsnip/pull/793/commits/e99dfd55115ca4dc36098527faea2da1dd2a710f):

```
# return a condition for use in `dplyr::filter()` on model info.
# if the user specified an engine and the model object reflects that in
# the `user_specified_engine` slot, filter the model info down to
# those that the user specified. if not, don't filter the model info at all.
```

R/misc.R
simonpcouch: I wrapped parsnip with the same cli `.pkg` tag as the recommended packages. Does that address this, or were you referring to the recommended packages as well?

R/fit.R
simonpcouch: Addressed in [8145467](https://github.com/tidymodels/parsnip/pull/793/commits/8145467e3eee9aa498009509f0a2791653ebf2e7)!

R/engines.R
simonpcouch: Addressed in [8145467](https://github.com/tidymodels/parsnip/pull/793/commits/8145467e3eee9aa498009509f0a2791653ebf2e7)! Commented in `set_mode`, `set_engine`, and `new_model_spec`.

simonpcouch:
The most recent commit exports and documents the relationship between the different checking functions more thoroughly:


> The helpers `spec_is_possible()`, `spec_is_loaded()`, and
> `prompt_missing_implementation()` provide tooling for checking
> model specifications. In addition to the `cls`, `engine`, and `mode`
> arguments, the functions take arguments `user_specified_engine` and
> `user_specified_mode`, denoting whether the user themselves has
> specified the engine or mode, respectively.
> 
> `spec_is_possible()` checks against the union of
> 
> * the current parsnip model environment and
> * the `model_info_table` of "pre-registered" model specifications
> 
> to determine whether a model is well-specified. See
> `parsnip:::read_model_info_table()` for this table.
> 
> `spec_is_loaded()` checks only against the current parsnip model environment.
> 
> `spec_is_possible()` is executed automatically on `new_model_spec()`,
> `set_mode()`, and `set_engine()`, and `spec_is_loaded()` is executed
> automatically in `print.model_spec()`, among other places. `spec_is_possible()`
> should be used when a model specification is still "in progress" of being
> specified, while `spec_is_loaded` should only be called when parsnip or an
> extension receives some indication that the user is "done" specifying a model
> specification: at print, fit, addition to a workflow, or `extract_*()`, for
> example.
> 
> When `spec_is_loaded()` is `FALSE`, the `prompt_missing_implementation()`
> helper will construct an informative message to prompt users to load or
> install needed packages. It's `prompt` argument refers to the prompting
> function to use, usually `cli::cli_inform` or `cli::cli_abort`, and the
> ellipses are passed to that function.

R/misc.R
topepo: Nope, your changes are perfect

PR #799, opened 2022-08-24
update snaps to remove tibble print suggestion

simonpcouch:
Looks like this won't print in non-interactive sessions post-pillar 1.8.1!


simonpcouch:
Looks like soome build systems don't grab the newest pillar version yet—will wait a couple days to re-run Actions.

PR #801, opened 2022-08-30
Make fit_xy() fail when mode is unknown

topepo:
As `fit()` currently does.

### DIFF
--- a/parsnip.Rproj
+++ b/parsnip.Rproj
@@ -17,5 +17,6 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
+PackageCleanBeforeInstall: Yes
 PackageInstallArgs: --no-multiarch --with-keep.source
 PackageRoxygenize: rd,collate,namespace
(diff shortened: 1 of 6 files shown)


DavisVaughan (approved):
It is probably worth testing this against the book and the workshop notes? It seems like something that people could definitely accidentally rely on.

Also, I think you should be able to remove every usage of `check_mode()`, because mode is now required in all cases and is never inferred "automatically"
https://github.com/tidymodels/parsnip/search?q=check_mode

parsnip.Rproj
```diff
@@ -17,5 +17,6 @@ StripTrailingWhitespace: Yes
 
 BuildType: Package
 PackageUseDevtools: Yes
+PackageCleanBeforeInstall: Yes
```
DavisVaughan, line 20: Looks like you need to update your version of the IDE. Neither the "released" version of the IDE nor the current daily use this anymore

topepo:
Checked against TMwR and revdeps. Everything looks good.

DavisVaughan:
Ok, I think we are good to move forward with this then. I'll go ahead and release workflows with this change.

PR #771, opened 2022-07-20
PR to allow dataframe as `validation` arg in `xgboost`

joeycouse:
Just a draft of allowing user to pass a dataframe as the `validation` arg for `xgboost`. From what I can tell, the main issue that prevents from passing a df with extra columns is the fact that when `y` is passed to `as_xgb_data()`:
https://github.com/tidymodels/parsnip/blob/f8505bdf6f039076c1eca5aa37a103a6e072887d/R/boost_tree.R#L380

The `y` vector is unnamed, so you can't just select the columns from `x` and which column corresponds to `y` to pass to
`xgb.DMatrix()`

Also case weights weren't being passed to the internal validation set in the current implementation. 

Would be happy to further develop this PR with some advice from your team

# Examples
``` r
> # Works 
> reg_fit <-
+   boost_tree(trees = 10, mode = "regression") %>%
+   set_engine("xgboost", 
+              eval_metric = "mae", 
+              validation = mtcars[1:3,],
+              verbose = 1) %>%
+   fit(mpg ~ ., data = mtcars)
[1]	validation-mae:14.428947 
[2]	validation-mae:10.406661 
[3]	validation-mae:7.885722 
[4]	validation-mae:5.953925 
[5]	validation-mae:4.558322 
[6]	validation-mae:3.400241 
[7]	validation-mae:2.560631 
[8]	validation-mae:1.757373 
[9]	validation-mae:1.187967 
[10]	validation-mae:0.787739 
> 
> reg_fit <-
+   boost_tree(trees = 10, mode = "regression") %>%
+   set_engine("xgboost", 
+              eval_metric = "mae", 
+              validation = 0.2,
+              verbose = 1) %>%
+   fit(mpg ~ ., data = mtcars)
[1]	validation-mae:14.203231 
[2]	validation-mae:9.799767 
[3]	validation-mae:7.313254 
[4]	validation-mae:5.302311 
[5]	validation-mae:3.932890 
[6]	validation-mae:2.969718 
[7]	validation-mae:2.525703 
[8]	validation-mae:2.261265 
[9]	validation-mae:2.169489 
[10]	validation-mae:2.301616 
> 
> 
> # Errors
> car_rec <- 
+   recipe(mpg ~ disp + hp + cyl, data = mtcars) |> 
+   update_role(cyl, new_role = 'id')
> 
> 
> reg_fit <-
+     boost_tree(trees = 10, mode = "regression") %>%
+     set_engine("xgboost", 
+                eval_metric = "mae", 
[...]

PR #803, opened 2022-08-30
pass weights to xgboost internal validation set

joeycouse:
In response to https://github.com/tidymodels/parsnip/pull/771#issuecomment-1231991055

This pull request passes case weights to the interval validation set of `xgboost`. This causes test failures here:

https://github.com/tidymodels/parsnip/blob/e1eb30a6f6704bd0a3cf61736ea1d26e1d8eb081/tests/testthat/test_boost_tree_xgboost.R#L409

and here.

https://github.com/tidymodels/parsnip/blob/e1eb30a6f6704bd0a3cf61736ea1d26e1d8eb081/tests/testthat/test_boost_tree_xgboost.R#L449

Seems like the original intention was to not pass case weights to the internal validation set?

### DIFF
--- a/R/boost_tree.R
+++ b/R/boost_tree.R
@@ -432,7 +432,12 @@ as_xgb_data <- function(x, y, validation = 0, weights = NULL, event_level = "fir
       # Split data
       m <- floor(n * (1 - validation)) + 1
       trn_index <- sample(1:n, size = max(m, 2))
-      val_data <- xgboost::xgb.DMatrix(x[-trn_index,], label = y[-trn_index], missing = NA)
+      val_info_list <- list(label = y[-trn_index])
+      if (!is.null(weights)) {
+        val_info_list$weight <- weights[-trn_index]
+      }
+
+      val_data <- xgboost::xgb.DMatrix(x[-trn_index,], info = val_info_list, missing = NA)
       watch_list <- list(validation = val_data)
 
       info_list <- list(label = y[trn_index])


topepo:
This depends on the type of weight: 

- Importance weights only affect the model estimation and supervised recipes steps. They are not used with yardstick functions for calculating measures of model performance.

- Frequency weights are used for all parts of the preprocessing, model fitting, and performance estimation operations.

(This is form the [blog post](https://www.tidyverse.org/blog/2022/05/case-weights/) but it should be better documented in hardhat). 

So we should do this but just for frequency weights (which are less likely to be used with boosting.

joeycouse:
@topepo @simonpcouch I've updated the PR to only pass freq weights to the internal validation set. This is done by delaying the conversion of the weights to numeric/integer till pass to `as_xgb_data()` 

``` r
library(parsnip)

freq_weights <- hardhat::frequency_weights(1:32)

mtcar_x <- mtcars[, -1]
mtcar_mat <- as.matrix(mtcar_x)

set.seed(1)
val_freq_wts <- parsnip:::as_xgb_data(mtcar_mat, mtcars$mpg, weights = freq_weights, validation = 1/10)
xgboost::getinfo(val_freq_wts$watchlist$validation, "weight")
#> [1]  3 17 26


imp_wts <- hardhat::importance_weights(1:32)

mtcar_x <- mtcars[, -1]
mtcar_mat <- as.matrix(mtcar_x)

set.seed(1)
val_freq_wts <- parsnip:::as_xgb_data(mtcar_mat, mtcars$mpg, weights = imp_wts, validation = 1/10)
xgboost::getinfo(val_freq_wts$watchlist$validation, "weight")
#> NULL
```

<sup>Created on 2022-09-01 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

PR #807, opened 2022-09-13
set `call` in internal model registration errors

simonpcouch:
The names of internal functions are shown in the message for a few different user-facing model registration errors. This PR proposes updating the call information in a few of those errors to reduce cognitive load + improve readability.

On CRAN:

``` r
library(parsnip)

linear_reg() %>%
  set_engine("boop") %>%
  set_mode("regression")
#> Error in `check_spec_mode_engine_val()`:
#> ! Engine 'boop' is not supported for `linear_reg()`. See `show_engines('linear_reg')`.

linear_reg(engine = "boop") %>%
  set_mode("regression")
#> Error in `check_spec_mode_engine_val()`:
#> ! Engine 'boop' is not supported for `linear_reg()`. See `show_engines('linear_reg')`.
```

<sup>Created on 2022-09-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

With this PR:

``` r
library(parsnip)

linear_reg() %>%
  set_engine("boop") %>%
  set_mode("regression")
#> Error in `set_engine()`:
#> ! Engine 'boop' is not supported for `linear_reg()`. See `show_engines('linear_reg')`.

linear_reg(engine = "boop") %>%
  set_mode("regression")
#> Error in `linear_reg()`:
#> ! Engine 'boop' is not supported for `linear_reg()`. See `show_engines('linear_reg')`.
```

<sup>Created on 2022-09-14 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

[EDIT: deleted an outdated comment about tests and updated reprex with new functionality.]

### DIFF
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -181,7 +181,7 @@ stop_incompatible_mode <- function(spec_modes, eng = NULL, cls = NULL) {
     msg,
     glue::glue_collapse(glue::glue("'{spec_modes}'"), sep = ", ")
   )
-  rlang::abort(msg)
+  rlang::abort(msg, call = NULL)
 }
 
 stop_incompatible_engine <- function(spec_engs, mode) {
@@ -189,7 +189,7 @@ stop_incompatible_engine <- function(spec_engs, mode) {
     "Available engines for mode {mode} are: ",
     glue::glue_collapse(glue::glue("'{spec_engs}'"), sep = ", ")
   )
-  rlang::abort(msg)
+  rlang::abort(msg, call = NULL)
 }
 
 stop_missing_engine <- function(cls) {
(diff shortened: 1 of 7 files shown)


hfrick (commented):
I really like this sort of thing 🎉  Two questions:

I agree that we don't want to have `check_*()` popping up in the error message. Right now, it doesn't show any call though and I think we generally want it to show the high-level call e.g., `set_mode()`. Have you thought about this already or would it be possible to pass the high-level call through to the check function without it being too complex? (Happy to add examples if you'd like!)

Re the tests: why are we only testing on the error and not capturing a snapshot? I feel like this would be a good use case for snapshot tests.

simonpcouch:
Agreed that displaying the high-level call would be helpful! I wasn't sure how to weigh the additional complexity of passing those calls with the improvement in interface, but am definitely up for that change.

Just pushed those changes, as well as testing updates (separately so we can be sure tests pass in both cases). This PR is a bit larger in scope now, but still only sets the `call` argument in user-facing machinery called inside of `check_spec_mode_engine_val`.

topepo:
@hfrick how do these changes look?

hfrick (approved):
Looks great now! 🙌

PR #813, opened 2022-09-21
print `(unknown mode)` instead of `(unknown)`

simonpcouch:
The model spec printing method calls out the spec’s mode:

``` r
library(parsnip)
  
boost_tree(mode = "regression")
#> Boosted Tree Model Specification (regression)
#> 
#> Computational engine: xgboost
```

When the `mode` is `"unknown"`, the output `(unknown)` beside the model spec’s type is ambiguous.

``` r
boost_tree()
#> Boosted Tree Model Specification (unknown)
#> 
#> Computational engine: xgboost
```

This PR proposes printing `(unknown mode)` instead:

``` r
boost_tree()
#> Boosted Tree Model Specification (unknown mode)
#> 
#> Computational engine: xgboost
```

<sup>Created on 2022-09-21 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

### DIFF
--- a/R/print.R
+++ b/R/print.R
@@ -13,7 +13,9 @@ print_model_spec <- function(x, cls = class(x)[1], desc = get_model_desc(cls), .
     prompt_missing_implementation(spec = structure(x, class = cls), prompt = cli::cli_inform)
   }
 
-  cat(desc, " Model Specification (", x$mode, ")\n\n", sep = "")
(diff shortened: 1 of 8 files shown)


simonpcouch:
This change would be quite breaking for snapshots—wouldn't anticipate it presents any challenges for getting a new release onto CRAN, but more snapshot updates in extension packages' future. Hm.

PR #833, opened 2022-10-28
add `fit_xy.decision_tree()`

hfrick:
This is a joint PR with https://github.com/tidymodels/censored/pull/228 - the parsnip one should be merged first.

With current CRAN versions we can't use `fit_xy()` for a decision tree with the rpart engine for censored regression. The reason is that `pec::pecRpart()` calls `prodlim::EventHistory.frame()`, which expects a call to `Surv()` _in_ the formula on the left-hand side while we make the Surv object outside of the formula. 

This has already been fixed in the dev version of prodlim but as long as that's not on CRAN yet, I'd want `fit_xy()` to fail properly for this model spec, hence the addition of `fit_xy.decision_tree()`. We discussed in the team meeting that this is better placed in parsnip than censored since it gives it better visibility for the other extension packages working with `decsion_tree()`.


simonpcouch (approved):
This looks great! You're welcome to merge—I'll leave that to you, re: order/branch sensitivity.

One more bit of paper trail for passersby: this method hadn't yet made it to CRAN censored yet—this transfers the method from one dev version to the other.

PR #835, opened 2022-10-28
check outcome class conditional on whether `y` is atomic

### LINKED ISSUE #834: Make `check_outcome()` faster
`check_outcome()` currently checks _each element_ of a vector to be numeric for the regression mode or to be a factor for the classification mode but we only need it to check the vector.

https://github.com/tidymodels/parsnip/blob/025cc59b89ad87fa0971f4f4e94d40bb0ceaa027/R/misc.R#L326 

and

https://github.com/tidymodels/parsnip/blob/025cc59b89ad87fa0971f4f4e94d40bb0ceaa027/R/misc.R#L330

simonpcouch:
😮

With CRAN parsnip:

``` r
library(tidymodels)

set.seed(1)
dat <- tibble(x = rnorm(1e7), y = x + rnorm(1e7, sd = .2))

system.time({
  linear_reg() %>%
    fit(y ~ x, dat)
})
#>    user  system elapsed 
#>   4.108   0.158   4.267
```

<sup>Created on 2022-10-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

With this PR:

``` r
library(tidymodels)

set.seed(1)
dat <- tibble(x = rnorm(1e7), y = x + rnorm(1e7, sd = .2))

system.time({
  linear_reg() %>%
    fit(y ~ x, dat)
})
#>    user  system elapsed 
#>   1.251   0.187   1.442
```

<sup>Created on 2022-10-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

Closes #834.

`check_outcome` is a bit conditional-heavy, though I have trouble imagining how these could be converted to switch- or tibbled-based logic elegantly.

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -322,21 +322,29 @@ new_model_spec <- function(cls, args, eng_args, mode, user_specified_mode = TRUE
 check_outcome <- function(y, spec) {
   if (spec$mode == "unknown") {
     return(invisible(NULL))
-  } else if (spec$mode == "regression") {
+  }
+
+  if (spec$mode == "regression") {
     outcome_is_numeric <- if (is.atomic(y)) {is.numeric(y)} else {all(map_lgl(y, is.numeric))}
     if (!outcome_is_numeric) {
       rlang::abort("For a regression model, the outcome should be numeric.")
     }
-  } else if (spec$mode == "classification") {
+  }
+
+  if (spec$mode == "classification") {
     outcome_is_factor <- if (is.atomic(y)) {is.factor(y)} else {all(map_lgl(y, is.factor))}
     if (!outcome_is_factor) {
       rlang::abort("For a classification model, the outcome should be a factor.")
     }
-  } else if (spec$mode == "censored regression") {
-    if (!inherits(y, "Surv")) {
+  }
+
+  if (spec$mode == "censored regression") {
+    outcome_is_surv <- inherits(y, "Surv")
+    if (!outcome_is_surv) {
       rlang::abort("For a censored regression model, the outcome should be a `Surv` object.")
     }
   }
+
   invisible(NULL)
 }
 
--- a/R/misc.R
+++ b/R/misc.R
@@ -323,11 +323,13 @@ check_outcome <- function(y, spec) {
   if (spec$mode == "unknown") {
     return(invisible(NULL))
   } else if (spec$mode == "regression") {
-    if (!all(map_lgl(y, is.numeric))) {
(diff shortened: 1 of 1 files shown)


R/misc.R
```diff
@@ -323,11 +323,13 @@ check_outcome <- function(y, spec) {
   if (spec$mode == "unknown") {
     return(invisible(NULL))
   } else if (spec$mode == "regression") {
```
hfrick, line 325: since you commented on the code being condition-heavy: this could just be `if` since we only get there if we don't return early

R/misc.R
```diff
@@ -323,11 +323,13 @@ check_outcome <- function(y, spec) {
...
+    if (!outcome_is_numeric) {
       rlang::abort("For a regression model, the outcome should be numeric.")
     }
   } else if (spec$mode == "classification") {
-    if (!all(map_lgl(y, is.factor))) {
+    outcome_is_factor <- if (is.atomic(y)) {is.factor(y)} else {all(map_lgl(y, is.factor))}
+    if (!outcome_is_factor) {
       rlang::abort("For a classification model, the outcome should be a factor.")
     }
   } else if (spec$mode == "censored regression") {
```
hfrick, line 335: what do you think about extending that pattern of first making `outcome_is_*`  to the censored regression condition?

hfrick (approved):
Such a speed-up, nice! 🏎️

R/misc.R
```diff
@@ -323,11 +323,13 @@ check_outcome <- function(y, spec) {
   if (spec$mode == "unknown") {
     return(invisible(NULL))
   } else if (spec$mode == "regression") {
-    if (!all(map_lgl(y, is.numeric))) {
+    outcome_is_numeric <- if (is.atomic(y)) {is.numeric(y)} else {all(map_lgl(y, is.numeric))}
+    if (!outcome_is_numeric) {
       rlang::abort("For a regression model, the outcome should be numeric.")
     }
   } else if (spec$mode == "classification") {
```
hfrick, line 330: if we assume that there will ever only be a single mode at this point, the other `else if`s could also just be `if`s, no?

R/misc.R
simonpcouch: sure thing! this and other suggestions implemented in [9b6a00f](https://github.com/tidymodels/parsnip/pull/835/commits/9b6a00f146f1880a4c12f24f5dd53123baa3574b). :)

hfrick (approved):

PR #838, opened 2022-11-02
make lightgbm's `num_leaves` engine arg tunable

hfrick:
This PR makes use of https://github.com/tidymodels/dials/pull/256 to make the `num_leaves` engine arg of lightgbm tunable. 

Having the news bullet here instead of bonsai feels a little off but I guess it's the best place?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -41,7 +41,7 @@ Imports:
 Suggests: 
     C50,
     covr,
-    dials (>= 0.1.0),
+    dials (>= 1.0.0.9001),
     earth,
     ggrepel,
     keras,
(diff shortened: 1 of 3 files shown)


simonpcouch (commented):
This looks great!

> Having the news bullet here instead of bonsai feels a little off but I guess it's the best place?

I will be making some [related changes](https://github.com/tidymodels/dials/pull/256#issuecomment-1300369893) in bonsai anyway and can send a release out with NEWS there. I'd say we're fine to include the NEWS entry here as well!

----

Just "Comment"ing rather than approving merge since we'll discuss release timeline for parsnip in the group meeting tomorrow. If we plan on parsnip going out in coming days, consider this an "Approve" once that release is out. :)

PR #852, opened 2022-12-13
unexport unused check helpers

simonpcouch:
Revdepcheck results:

```
We checked 59 reverse dependencies, comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 0 new problems
```

🕺


EmilHvitfeldt (approved):
Very nice!

PR #854, opened 2022-12-22
Tests for `poisson_reg()`

hfrick:
We generally try to put tests on the basic model function (like `poisson_reg()`) in parsnip since the model definitions now live here. This PR moves some of those tests from poissonreg to here.

The corresponding PR to remove them from poissonreg is: https://github.com/tidymodels/poissonreg/pull/55

### DIFF
--- a/tests/testthat/test-poisson_reg.R
+++ b/tests/testthat/test-poisson_reg.R
@@ -1,0 +1,13 @@
+test_that("updating", {
+  expect_snapshot(
+    poisson_reg(penalty = 1) %>%
+      set_engine("glmnet", lambda.min.ratio = 0.001) %>%
+      update(mixture = tune())
+  )
+})
+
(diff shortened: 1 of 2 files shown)


tests/testthat/test-poisson_reg.R
```diff
@@ -0,0 +1,13 @@
+test_that("updating", {
+  expect_snapshot(
+    poisson_reg(penalty = 1) %>%
+      set_engine("glmnet", lambda.min.ratio = 0.001) %>%
+      update(mixture = tune())
+  )
+})
+
+test_that("bad input", {
+  expect_snapshot(error = TRUE, poisson_reg(mode = "bogus"))
```
simonpcouch, line 10: `mode = "bogus"`😆

simonpcouch (approved):
Looks great!

PR #855, opened 2023-01-12
estimate censoring probabilities

topepo:
For censored regression models, fit a KM curve with the status field reversed. 

These calculations may be used to measure model performance. 

Unit tests will be in extratests.


R/fit_helpers.R
```diff
@@ -198,4 +198,39 @@ xy_form <- function(object, env, control, ...) {
...
+
+  if (!is.null(eval_env$weights)) {
+    cl <- rlang::call_modify(cl, caseweights = rlang::expr(eval_env$weights))
+  }
+  rkm <- try(rlang::eval_tidy(cl), silent = TRUE)
+  if (!inherits(rkm, "try-error")) {
+    attr(f, ".Environment") <- rlang::base_env()
+  }
+  attr(rkm$formula, ".Environment") <- rlang::base_env()
+  list(formula = f, model = rkm)
```
topepo, line 233: Should we include a character vector for `required_pkgs` in here?

topepo: Also, should we embed the call to make predictions on new samples? It might change in the future if we open this up to other models/formulas

topepo: We could also make a `censoring_prob` parsnip model type (but that's potentially a lot for right now)

hfrick: Isn't the answer to the second question yes because we need to calculate the inverse probability weight for the `new_data` when calculating performance? Do we need the formula environments then as well?

hfrick: I'd leave a `censoring_prob` model type for later when we better understand what we gain for the added complexity.

hfrick: what situations would a `required_pkgs` addition help with? is that for running things in parallel?

topepo: > Isn't the answer to the second question yes because we need to calculate the inverse probability weight for the `new_data` when calculating performance? Do we need the formula environments then as well?

Right now, we only predict using the time of analysis (or the observed event time, depending on the type of weight). 

This assumes uninformative censoring, so adding covariates to that model would extend to informative censoring situations, which is one of the upsides to IPCW in general. 

That could be in a second version of this, but we should plan on generalizing the interface.  

> what situations would a `required_pkgs` addition help with? is that for running things in parallel? 

Mainly to ensure that any additional packages are around to use this functionality. We Suggests `prodlim` but should check for it in the usual way (which uses `required_pkgs()`).

topepo:
I added more elements to the object, made a constructor, added classes, and made `print()`/`predict()` methods. 

@hfrick Can you give these changes a look? Most of the changes are in 2126975

It's not completely future-proof but should give us a lot of latitude if we decide to extend this for informative censoring. 

Unit tests in https://github.com/tidymodels/extratests/pull/67 will be merged after this is finalized.

PR #866, opened 2023-02-08
Improve readability of docs for `predict.model_fit()`

hfrick:
A lot of what `predict.model_fit()` takes as arguments and what it returns depends on the type of prediction. To make this relatively long and detailed page easier to skim, I've replaced various different descriptions of the prediction type, with mentions of `type = "<whatever the type>"`.

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -55,7 +55,7 @@
 #' have the opposite sign as what the underlying model's `predict()` method
 #' produces. Set `increasing = FALSE` to suppress this behavior.
 #'
-#' @return With the exception of `type = "raw"`, the results of
+#' @return With the exception of `type = "raw"`, the result of
 #'  `predict.model_fit()`
 #'
 #'  * is a tibble
--- a/R/predict.R
+++ b/R/predict.R
@@ -4,44 +4,41 @@
 #'  `predict()` can be used for all types of models and uses the
 #'  "type" argument for more specificity.
 #'
-#' @param object An object of class `model_fit`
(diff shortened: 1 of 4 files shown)


R/predict.R
```diff
@@ -59,36 +56,41 @@
 #' produces. Set `increasing = FALSE` to suppress this behavior.
 #'
 #' @return With the exception of `type = "raw"`, the results of
```
simonpcouch, line 58:
```suggestion
#' @return With the exception of `type = "raw"`, the result of
```

otherwise reads "the results of `predict.model_fit()` is a tibble," etc.

man/predict.model_fit.Rd
```diff
@@ -13,74 +13,80 @@
...
+the standard error of fit or prediction (on the scale of the
+linear predictors). Default value is \code{FALSE}.
+\item \code{quantile}: for \code{type} equal to \code{quantile}, the quantiles of the
+distribution. Default is \code{(1:9)/10}.
+\item \code{time}: for \code{type} equal to \code{"survival"} or \code{"hazard"}, the
+time points at which the survival probability or hazard is estimated.
 }}
 }
 \value{
 With the exception of \code{type = "raw"}, the results of
```
simonpcouch, line 52:
```suggestion
With the exception of \code{type = "raw"}, the result of
```

changing in the `.rd` directly as well to hopefully save you time.

simonpcouch (approved):
I think all of these changes are improvements—love it. :)

PR #868, opened 2023-02-09
Consolidate glmnet predict methods

hfrick:
This PR consolidates the different flavors of predict methods (`predict()`, `predict_<type>()` and `multi_predict()`) for different glmnet model types into one function per predict method flavor. 

The goal is to bring it all in one place to more easily maintain/apply changes. I'm currently not touching the class of the glmnet `model_fit` objects (yet?).

The glmnet tests in extratests run fine locally with this change. They'll run on CI when this in `main`.

### DIFF
--- a/R/glmnet.R
+++ b/R/glmnet.R
@@ -1,0 +1,146 @@
+# glmnet call stack using `predict()` when object has
+# classes "_<glmnet-class>" and "model_fit":
+#
+#  predict()
+# 	predict._<glmnet-class>(penalty = NULL)    <-- checks and sets penalty
+#    predict.model_fit()                       <-- checks for extra vars in ...
+#     predict_numeric()
+#      predict_numeric._<glmnet-class>()
+#       predict_numeric.model_fit()
+#        predict.<glmnet-class>()
+
+
+# glmnet call stack using `multi_predict` when object has
+# classes "_<glmnet-class>" and "model_fit":
+#
+# 	multi_predict()
+#    multi_predict._<glmnet-class>(penalty = NULL)
+#      predict._<glmnet-class>(multi = TRUE)   <-- checks and sets penalty
+#       predict.model_fit()                    <-- checks for extra vars in ...
+#        predict_raw()
+#         predict_raw._<glmnet-class>()
+#          predict_raw.model_fit(opts = list(s = penalty))
+#           predict.<glmnet-class>()
+
+
+predict_glmnet <- function(object,
+                           new_data,
+                           type = NULL,
+                           opts = list(),
+                           penalty = NULL,
+                           multi = FALSE,
+                           ...) {
+
+  if (any(names(enquos(...)) == "newdata")) {
+    rlang::abort("Did you mean to use `new_data` instead of `newdata`?")
+  }
+
+  # See discussion in https://github.com/tidymodels/parsnip/issues/195
+  if (is.null(penalty) & !is.null(object$spec$args$penalty)) {
+    penalty <- object$spec$args$penalty
+  }
+
+  object$spec$args$penalty <- .check_glmnet_penalty_predict(penalty, object, multi)
+
+  object$spec <- eval_args(object$spec)
+  predict.model_fit(object, new_data = new_data, type = type, opts = opts, ...)
+}
+
+predict_numeric_glmnet <- function(object, new_data, ...) {
+  if (any(names(enquos(...)) == "newdata"))
+    rlang::abort("Did you mean to use `new_data` instead of `newdata`?")
+
+  object$spec <- eval_args(object$spec)
+  predict_numeric.model_fit(object, new_data = new_data, ...)
+}
+
+predict_class_glmnet <- function(object, new_data, ...) {
+  if (any(names(enquos(...)) == "newdata"))
+    rlang::abort("Did you mean to use `new_data` instead of `newdata`?")
+
+  object$spec <- eval_args(object$spec)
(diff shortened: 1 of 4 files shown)


R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+    if (!(type %in% c("class", "prob", "link", "raw"))) {
+      rlang::abort("`type` should be either 'class', 'link', 'raw', or 'prob'.")
+    }
+    if (type == "prob")
+      dots$type <- "response"
+    else
+      dots$type <- type
+  }
+
+  pred <- predict(object, new_data = new_data, type = "raw",
```
hfrick, line 128: this is now a call the the generic `predict()` rather than directly calling the method for the glmnet class at hand, `predict._<glmnet-class>()`, so that it works for all classes

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+                  opts = dots, penalty = penalty, multi = TRUE)
+
+  model_type <- class(object$spec)[1]
```
hfrick, line 131: This isn't necessarily the final form for this. My current thinking is that it makes more sense to make this dependent on the prediction type, rather than the model type, but goal here is to make it more straightforward to consolidate all functions into one.

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+
+  if (any(names(enquos(...)) == "newdata")) {
+    rlang::abort("Did you mean to use `new_data` instead of `newdata`?")
+  }
+
+  if (object$spec$mode == "classification") {
+    if (is_quosure(penalty)) {
+      penalty <- eval_tidy(penalty)
+    }
+  }
```
hfrick, line 101: This block was part of the methods for logistic and multinomial regression so it's under "classification" here, same for the block in line 116-126

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+
+  dots <- list(...)
+
+  object$spec <- eval_args(object$spec)
```
hfrick, line 105: this line was right before the call to `predict()` for the classification model but I think it's okay to evaluate here already

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
+# glmnet call stack using `predict()` when object has
```
EmilHvitfeldt, line 1: I think it would be helpful to include where `predict_glmnet()` would sit in this diagram

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+
+predict_glmnet <- function(object,
+                           new_data,
+                           type = NULL,
+                           opts = list(),
+                           penalty = NULL,
+                           multi = FALSE,
+                           ...) {
+
+  if (any(names(enquos(...)) == "newdata")) {
```
EmilHvitfeldt, line 34: As another issue, we should properly consolidate this logic into a helper function if possible as it appears across the package. So we should open another issue

for this PR, there are some inconsistencies, some places it is

```r
  if (any(names(enquos(...)) == "newdata")) {
    rlang::abort("Did you mean to use `new_data` instead of `newdata`?")
  }
```

other places it is 

```r
  if (any(names(enquos(...)) == "newdata"))
    rlang::abort("Did you mean to use `new_data` instead of `newdata`?")
```

to follow the [style guide](https://style.tidyverse.org/syntax.html#inline-statements) they should all include `{}`

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+  object$spec <- eval_args(object$spec)
+
+  opts$s <- object$spec$args$penalty
+
+  predict_raw.model_fit(object, new_data = new_data, opts = opts, ...)
+}
+
+
+
+
```
EmilHvitfeldt, line 86: this is more of a personal preference, but unless it signifies something, I would avoid these duplicate newlines

R/glmnet.R
```diff
@@ -0,0 +1,146 @@
...
+    if (is.null(type))
+      type <- "class"
+    if (!(type %in% c("class", "prob", "link", "raw"))) {
+      rlang::abort("`type` should be either 'class', 'link', 'raw', or 'prob'.")
+    }
+    if (type == "prob")
+      dots$type <- "response"
+    else
+      dots$type <- type
+  }
```
EmilHvitfeldt, line 126: I find this logic quite hard to read at first glance. I think it is because we are have a handful of logic back to back with no `{}`. I like this suggestion, but I also wouldn't hate a newline between each logic block

```suggestion
  if (object$spec$mode == "classification") {
    if (is.null(type)) {
      type <- "class"
    }
    if (!(type %in% c("class", "prob", "link", "raw"))) {
      rlang::abort("`type` should be either 'class', 'link', 'raw', or 'prob'.")
    }
    if (type == "prob") {
      dots$type <- "response"
    } else {
      dots$type <- type
    }
  }
```

R/glmnet.R
EmilHvitfeldt: I think this is the right call

EmilHvitfeldt (commented):
I like the effort. all of this glmnet is quite unwieldy. Keeping it in a separate file will make it a lot easier organizational

R/glmnet.R
hfrick: Totally agree, I'm planning to do that next. `predict.model_fit()` calls `check_pred_type_dots()` which also checks for `newdata`. I think that check might have originated here but since it's not glmnet-specific it's better place in the `model_fit` method anyway.

EmilHvitfeldt:
Approve!

PR #869, opened 2023-02-14
register `tune_args()` and `tunable()` methods unconditionally

simonpcouch:
Similar rationale and implementation as tidymodels/workflows#193. :)


EmilHvitfeldt (approved):

R/tune_args.R
```diff
@@ -1,6 +1,10 @@
+#' @importFrom generics tune_args
+#' @export
+generics::tune_args
 
-# Lazily registered in .onLoad()
-tune_args_model_spec <- function(object, full = FALSE, ...) {
+#' @method tune_args model_spec
```
simonpcouch, line 5: For whatever reason, this one required the explicit `@method` tag to probably document as an S3 method.

R/tunable.R
```diff
@@ -1,7 +1,13 @@
-# Lazily registered in .onLoad()
+
 # Unit tests are in extratests
 # nocov start
-tunable_model_spec <- function(x, ...) {
+
+#' @importFrom generics tunable
```
hfrick, line 5: this is already imported in `parsnip-package.R` so let's remove one of them. I'd remove this one here because I also think that it shouldn't be (re)exported (see https://github.com/tidymodels/workflows/pull/193#pullrequestreview-1300378027)

R/tune_args.R
```diff
@@ -1,6 +1,10 @@
+#' @importFrom generics tune_args
```
hfrick, line 1: similar comment on import/export and where it should sit

hfrick (commented):
same thinking around reexporting the generics for `tune_args()` and `tunable()`

R/zzz.R
```diff
@@ -9,45 +9,9 @@
   vctrs::s3_register("generics::tidy", "_fishnet")
   vctrs::s3_register("generics::glance", "model_fit")
   vctrs::s3_register("generics::augment", "model_fit")
-  vctrs::s3_register("generics::required_pkgs", "model_fit")
-  vctrs::s3_register("generics::required_pkgs", "model_spec")
```
hfrick, line 13: should this stay in this PR? given that we register it without vctrs in `R/required_pkgs.R`, it should definitely go, it's more a question of "should this be its own PR"? Up to you!

hfrick (approved):
Approve! 🙌

R/zzz.R
simonpcouch: I'm with you that this probably ought to live in its own PR! Reverting.

PR #879, opened 2023-02-22
remove check on newdata typo in glmnet functions

### LINKED ISSUE #877: Remove checks on `newdata` typo in glmnet-related functions
The various glmnet-related predict methods include a check if the user may have used `newdata` instead of `new_data`. This is not glmnet-specific and gets checked [inside of `predict.model_fit()`](https://github.com/tidymodels/parsnip/blob/a482442da20b6be271062e198a69f9d422523ab1/R/predict.R#L149) via [`check_pred_type_dots()`](https://github.com/tidymodels/parsnip/blob/a482442da20b6be271062e198a69f9d422523ab1/R/predict.R#L333-L335) so we can remove it here.

hfrick:
closes  #877 

`predict_glmnet()` calls `predict.model_fit()` which checks this via `check_pred_type_dots()` so we don't need it in `predict_glmnet()`.

The `predict_<type>()` methods don't run this check. I think generally users are supposed to use `predict(type = <type>)` rather than `predict_<type>()` but the `predict_<type>()` functions are exported. If we think this check is important to have also when directly calling `predict_<type>()`, I'd suggest we add it to the method for `model_fit` objects but since it's not glmnet-specific, I have removed it here. 

`multi_predict()` does not have a central `model_fit` method (because it doesn't apply to all models) so I have left the check in the glmnet-specific method.

The tests for glmnet engines live in extratests, those checks pass locally and will be run on CI once this is merged.

### DIFF
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -423,6 +423,7 @@ spec_has_pred_type <- function(object, type) {
   possible_preds <- names(object$spec$method$pred)
   any(possible_preds == type)
 }
+
 check_spec_pred_type <- function(object, type) {
   if (!spec_has_pred_type(object, type)) {
     possible_preds <- names(object$spec$method$pred)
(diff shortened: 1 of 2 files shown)


EmilHvitfeldt (approved):
Good catch!

topepo (approved):

PR #881, opened 2023-02-22
Fix test on translating `path_values` to `lambda` for glmnet

hfrick:
The snapshot previously just captured the error message about the missing penalty value, now it shows that `path_values` gets translated to `lambda` as it should.


simonpcouch (approved):
Groovy!🐇

topepo (approved):

PR #888, opened 2023-02-23
make `will_make_matrix()` behave consistently for 0-length factors

### LINKED ISSUE #883: uninformative "outcome should be a factor" error with missing predictors
Hi! I want to use xgboost for classification but are getting a (simple?) error stating that my outcome variable is not a factor which it indeed is. Is there another better way of making the outcome variable be interpreted as a factor in the model?

All help is much appreciated.
```
train_test_split = rsample::initial_split(df_num, prop=train_set_prop, strata=kombucha_intake_num)
train = training(train_test_split)
test = testing(train_test_split)
train$kombucha_intake_num =as.factor(train$kombucha_intake_num)
test$kombucha_intake_num = as.factor(test$kombucha_intake_num)

xgb_mod <- 
  boost_tree(trees=50) %>%
  set_engine("xgboost") %>%
  fit(kombucha_intake_num ~ ., data=train)`
```

Error: Error in `check_outcome()`:
! For a classification model, the outcome should be a factor.

simonpcouch:
Closes #883, more context there!🪲

With CRAN version:

``` r
parsnip:::will_make_matrix(factor("a"))
#> [1] FALSE
parsnip:::will_make_matrix(factor(character()))
#> [1] TRUE
```

With this PR:

``` r
parsnip:::will_make_matrix(factor("a"))
#> [1] FALSE
parsnip:::will_make_matrix(factor(character()))
#> [1] FALSE
```

<sup>Created on 2023-02-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -324,7 +324,7 @@ make_formula <- function(x, y, short = TRUE) {
 
 
 will_make_matrix <- function(y) {
-  if (is.matrix(y) | is.vector(y))
+  if (is.matrix(y) | is.vector(y) | (is.factor(y) & length(y) == 0))
     return(FALSE)
   cls <- unique(unlist(lapply(y, class)))
   if (length(cls) > 1)
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -324,7 +324,7 @@ make_formula <- function(x, y, short = TRUE) {
 
 
 will_make_matrix <- function(y) {
-  if (is.matrix(y) | is.vector(y) | (is.factor(y) & length(y) == 0))
+  if (is.matrix(y) | is.vector(y) | is.factor(y))
     return(FALSE)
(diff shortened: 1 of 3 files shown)


R/convert_data.R
```diff
@@ -324,7 +324,7 @@ make_formula <- function(x, y, short = TRUE) {
 
 
 will_make_matrix <- function(y) {
-  if (is.matrix(y) | is.vector(y))
+  if (is.matrix(y) | is.vector(y) | (is.factor(y) & length(y) == 0))
```
simonpcouch, line 327: Another possible change here is to just add `| is.factor(y)`. This does change where in the function that case used to return from, but I believe gives the same results.

hfrick: I would also think that we only need `| is.factor(y)` 🤔

hfrick (approved):

PR #890, opened 2023-02-23
Support glmnet models with base-R families

### LINKED ISSUE #483: handle quasi-likelihood glmnet models better
These don't return an object class specific to their model: 

``` r
library(tidymodels)
data("two_class_dat")

logistic_reg(penalty = 0.1) %>% 
  set_engine("glmnet") %>% 
  fit(Class ~ ., data = two_class_dat) %>% 
  pluck("fit") %>% 
  class()
#> [1] "lognet" "glmnet"

logistic_reg(penalty = 0.1) %>% 
  set_engine("glmnet", family = quasibinomial(link = "logit")) %>% 
  fit(Class ~ ., data = two_class_dat) %>% 
  pluck("fit") %>% 
  class()
#> [1] "glmnetfit" "glmnet"
```

<sup>Created on 2021-05-04 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0.9000)</sup>

We should add class to the `parsnip` model or restrict this family for this engine.

### LINKED ISSUE #738: Unable to tune penalty for `glmnet` with non-default family
## The problem

I'm unable to tune the penalty hyperparameter for a `glmnet` model specification with non-default family engine argument. I've traced the error down to the `multi_predict` generic not having a method for generalized model fits (class `_glmnetfit`). 

I have also tried passing specific `path_values` as an engine argument (which I think will need to be done to correctly compare penalty values) but that didn't resolve the underlying issue with `multi_predict`.

## Reproducible example

``` r
library(tidyverse)
library(tidymodels)

recipe <- recipe(hp ~ ., data = mtcars)
lasso <- linear_reg(
  mixture = 1,
  penalty = !!tune()
) %>% 
  set_engine(
    engine = 'glmnet',
    family = gaussian(link = 'log')
  )
wflow <- workflow(recipe, lasso)
grid <- grid_max_entropy(penalty(), size = 10)
folds <- vfold_cv(mtcars)

res <- tune_grid(
  object = wflow,
  resamples = folds,
  grid = grid
)
#> x Fold01: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold02: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold03: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold04: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold05: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold06: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold07: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold08: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold09: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> x Fold10: preprocessor 1/1, model 1/1 (predictions): Error in `parsnip::multi_pred...
#> Warning: All models failed. See the `.notes` column.

res %>% collect_notes %>% distinct(note) %>% reduce(c) %>% cli::cli_ul()
#> • Error in `parsnip::multi_predict()`: ! No `multi_predict` method exists for
#> objects with classes '_glmnetfit', 'model_fit'
```

<sup>Created on 2022-05-31 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

hfrick:
Closes #483 and closes #738

Since version 4.0, glmnet supports base-R families for the `family` argument. The resulting fitted model objects all get a new class `glmnetfit` which is not supported in parsnip yet. 

We need several predict methods for this new class that check and set the penalty argument to work with the rest of parsnip.

We also need to translate between parsnip's prediction type (`"numeric"`, `"class"`, `"prob"`, etc) and glmnet's prediction type (`"response"`, `"class"`, etc). The gotcha moment here is that `type = "class"` exists for logistic and multinomial regression models fitted with glmnet's built-in families, resulting in objects of classes `lognet` and `multnet`, but does not exist for models fitted with any of the base-R families, resulting in objects of class `glmnetfit`. 

For `glmnetfit` models, we do have `type = "response"` available, which, for logistic and multinomial regression, gives us the class probabilities which we can translate to hard class predictions. 

We already do that for `predict()` on logistic regression, see [here](https://github.com/tidymodels/parsnip/blob/25881814167ff4de6139d55866c57bbcca5ba688/R/logistic_reg_data.R#L170) and [here](https://github.com/tidymodels/parsnip/blob/25881814167ff4de6139d55866c57bbcca5ba688/R/logistic_reg.R#L180), but not for `multi_predict()`. For multinomial regression, we use `type = "class"` for both [`predict()`](https://github.com/tidymodels/parsnip/blob/25881814167ff4de6139d55866c57bbcca5ba688/R/multinom_reg_data.R#L65) and [`multi_predict()`](https://github.com/tidymodels/parsnip/blob/25881814167ff4de6139d55866c57bbcca5ba688/R/glmnet.R#L131).

I have contemplated if we should try to use `type = "class"` wherever we can and then special-case the `glmnetfit` models or if we should lean the other way and use `type = "response"` consistently and translate from probabilities to class predictions even in cases where we could get that from glmnet. I have landed in a pragmatic middle with this PR. 

For multinomial regression, I don't see a corresponding suitable base-R family, so there is no added complexity via `glmnetfit` objects. I have therefore left this as is for the time being.

For logistic regression, this PR leans into "use type 'response' everywhere" because special-casing `glmnetfit` object would be a particular challenge: For `predict()`, the translation between parsnip's prediction type and glmnet's prediction type happens when `predict_<type>.model_fit()` [constructs the call](https://github.com/tidymodels/parsnip/blob/25881814167ff4de6139d55866c57bbcca5ba688/R/predict_class.R#L29) to glmnet's prediction method, using the type defined in `set_pred()`. There isn't really a (good?) way to set that conditional on the class of the model fit object. The (only? canonical?) way to manually take control of the translation between the prediction types is what `multi_predict()` does: it uses the parsnip `type = "raw"` and sets glmnet's `type` as part of the `opts` which are passed down to glmnet's predict method. Trying to do this for `predict()` in addition to `multi_predict()` would defy the whole purpose of parsnip's predict machinery. Hence, no special-casing the `glmnetfit` objects but rather `type = "response"` everywhere and taking control of how the predictions are formatted at the end (aka turn probabilities into class predictions or not). 

The corresponding tests are with the other glmnet-related tests in extratests: https://github.com/tidymodels/extratests/pull/77


R/logistic_reg.R
```diff
@@ -216,34 +216,41 @@ predict._lognet <- predict_glmnet
...
   pred <- as_tibble(pred)
-  pred$.row <- 1:nrow(pred)
-  pred <- gather(pred, group, .pred_class, -.row)
+  pred$.row <- seq_len(nrow(pred))
+  pred <- tidyr::pivot_longer(pred, -.row, names_to = "s", values_to = ".pred")
+
   if (type == "class") {
-    pred[[".pred_class"]] <- factor(pred[[".pred_class"]], levels = lvl)
+    pred <- pred %>%
+      dplyr::mutate(.pred_class = dplyr::if_else(.pred >= 0.5, lvl[2], lvl[1]),
```
hfrick, line 230: apart from this line adding the translation from probabilities to classes, the changes to this function are only to make use of dplyr and tidyr

simonpcouch (approved):
Do-er of diligence! Effectuator of exactness! Agent of accuracy!

This PR is super solid—no comments. Thanks for the helpful description.🍄

R/glmnet.R
```diff
@@ -135,13 +138,13 @@ multi_predict_glmnet <- function(object,
...
 
-  model_type <- class(object$spec)[1]
+
   res <- switch(
     model_type,
     "linear_reg" = format_glmnet_multi_linear_reg(pred, penalty = penalty),
     "logistic_reg" = format_glmnet_multi_logistic_reg(pred,
                                                       penalty = penalty,
-                                                      type = dots$type,
+                                                      type = type,
```
simonpcouch, line 147: A note to help speed up @topepo's review: `format_glmnet_multi_logistic_reg()` used to take whichever `type` was passed along to glmnet, and its internals read:

```
if (type == "class") {
  # ...
} else {
  # ...
}
```

That helper is now supplied the "parsnip" type instead, in this case one of `"class"` or `"prob"`, so that the above conditional has only one possible value for the `else`. The `format_glmnet_multi_multinom_reg()` helper does this already.

hfrick: @simonpcouch Thanks for adding this! 🙌

PR #893, opened 2023-02-27
Survival helpers

topepo:
Move some code from tune to parsnip since we will relocate the IPCW code here. This PR has basic helper functions. Unit tests that require actual `Surv` objects will be in extratests


EmilHvitfeldt:
Could this be moved to {hardhat} instead? They would be helpful in {yardstick} which doesn't import/suggests {parsnip}

topepo:
After talking with @DavisVaughan about moving them to hardhat, he recommended putting them in a portable file that can be added to other packages via `usethis::use_standalone("tidymodels/parsnip")`

R/ipcw.R
```diff
@@ -0,0 +1,34 @@
...
+# censored regression models
+
+# ------------------------------------------------------------------------------
+# Simple helpers for computing the probability of censoring
+
+# For avoiding extremely large, outlier weights
+trunc_probs <- function(probs, trunc = 0.01) {
+  complt_prob <- !is.na(probs)
+  probs_non_zero <- probs[complt_prob]
+  non_zero_min <- min(probs_non_zero[probs_non_zero > 0])
```
hfrick, line 12: I was mildly confused by `probs_non_zero` because those are not actually non-zero but non-missing. Wouldn't this do what you want?
```suggestion
  non_zero_min <- min(probs[probs > 0], na.rm = TRUE)
```

tests/testthat/test-ipcw.R
```diff
@@ -0,0 +1,41 @@
...
+  expect_equal(
+    parsnip:::.filter_eval_time(times_1),
+    times_1
+  )
+  expect_equal(
+    parsnip:::.filter_eval_time(times_1),
+    times_1
+  )
+  expect_snapshot_error(parsnip:::.filter_eval_time(-1))
+})
```
hfrick, line 36:
```suggestion
  expect_null(parsnip:::.filter_eval_time(NULL))
})
```
Since there is a code comment on this, I'd say it's worth a test.

tests/testthat/test-ipcw.R
```diff
@@ -0,0 +1,41 @@
...
+
+  expect_equal(
+    parsnip:::.filter_eval_time(times_1),
+    times_1
+  )
+  expect_equal(
+    parsnip:::.filter_eval_time(times_1),
+    times_1
+  )
+  expect_snapshot_error(parsnip:::.filter_eval_time(-1))
```
hfrick, line 35:
```suggestion
  expect_snapshot(error = TRUE, parsnip:::.filter_eval_time(-1))
```
This adds the code that generated the error to the snapshot. Since the code isn't long, I find that helpful when looking at the snapshots directly.

R/standalone-survival.R
```diff
@@ -0,0 +1,85 @@
...
+  attr(surv, "type")
+}
+
+.check_cens_type <- function(surv, type = "right", fail = FALSE) {
+  .is_surv(surv)
+  obj_type <- .extract_surv_type(surv)
+  good_type <- all(obj_type %in% type)
+  if (fail && !good_type) {
+    c_list <- paste0(type, collapse = ", ")
+    msg <- glue::glue("For this usage, the allowed censoring types are: {c_list}")
```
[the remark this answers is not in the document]
hfrick: we could also use `cli::cli_abort()` but that works ever so slightly differently from `rlang::abort()` (with how the `call` arg is handled, I think) so I left it at making the message first, then raising the error.

hfrick (commented):
I like the idea of the standalone file! There are suggestions for you to take or leave, I just want to point you to the one `.extract_surv_status()` and when it converts the status to `0`/`1`

R/ipcw.R
topepo: I can rename to make it more clear but I think the issue is that, if there are missing values, `probs > 0` gives us missing values.

PR #896, opened 2023-02-28
reuse results of long-running checks

simonpcouch:
``` r
library(tidymodels)
```

Within the call:

``` r
res <- fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars, 3))
```

The internal `check_spec_mode_engine_val()` helper is called 6 times, twice per resample. It’s always passed the same arguments, and as such will always give the same answer across resamples. Because of an inner join and some manipulation of environments, calls to this check end up accounting for \~10% of the runtime of the above code.

This PR is a proof-of-concept for some tools to hook into the new `tune_env` machinery in order to reuse intermediate results, like those outputted by `check_spec_mode_engine_val()`. With main dev tune and main dev parsnip:

``` r
bench::mark(
  fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars, 100)),
  iterations = 10
) %>%
  select(median, mem_alloc)
#> → A | warning: prediction from a rank-deficient fit may be misleading
#> There were issues with some computations   A: x1
#> 
#> # A tibble: 1 × 2
#>     median mem_alloc
#>   <bch:tm> <bch:byt>
#> 1    8.73s     167MB
```

With main dev tune and this PR to parsnip:

``` r
bench::mark(
  fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars, 100)),
  iterations = 10
) %>%
  select(median, mem_alloc)
#> → A | warning: prediction from a rank-deficient fit may be misleading
#> There were issues with some computations   A: x1
#> 
#> # A tibble: 1 × 2
#>     median mem_alloc
#>   <bch:tm> <bch:byt>
#> 1    7.86s     123MB
```

There are quite a few analogous changes that could be made to other internal helpers—a few percent here and there could go a long way to reducing the “tidymodels overhead” drastically.

None of this machinery kicks in when tune isn't installed or there's no active `tune_env`.

<sup>Created on 2023-02-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -52,6 +52,7 @@ Suggests:
     MASS,
     Matrix,
     mgcv,
+    mockr,
     modeldata,
     nlme,
     prodlim,
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -46,6 +46,79 @@ read_model_info_table <- function() {
 
 # ------------------------------------------------------------------------------
 
+# for checks and other intermediate steps that are redundant across fits to
+# resamples, we can skip computationally intensive ones after the first run
+# by hooking into tune's `tune_env` and storing their result.
+#
+# we can precede some long-running chunk of code with the conditional:
+#
+# if (has_intermediate_result("unique_label")) {
+#   return(get_intermediate_result("unique_label"))
+# }
+#
+# the first time this conditional is run, `has_intermediate_result` will
+# return `FALSE` and the long-running chunk will execute, resulting in a
+# `result`. That result can then be stored as:
+#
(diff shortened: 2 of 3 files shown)


topepo:
Can you look at what Edgar did with [probably](https://github.com/tidymodels/probably/blob/main/tests/testthat/helper-cal.R#L5:L39)? It gets run once, cached, then reused later.

simonpcouch:
Let's table this until we've made some more simple optimizations and revisit after then!

simonpcouch:
Closing in favor of more granular optimizations. :)

PR #899, opened 2023-03-02
Be more verbose about prediction types in `check_pred_type()`

hfrick:
This PR makes `check_pred_type()` equally verbose about the prediction type for censored regression as it already is for classification.

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -198,7 +198,7 @@ check_pred_type <- function(object, type, ...) {
   if (type == "numeric" & object$spec$mode != "regression") {
     rlang::abort("For numeric predictions, the object should be a regression model.")
   }
-  if (type == "class" & object$spec$mode != "classification"){
+  if (type == "class" & object$spec$mode != "classification") {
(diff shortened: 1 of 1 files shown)


EmilHvitfeldt (commented):
i generally like this style of control-flow for readability. Is this function run a lot of times? it might be worth it time wise to use a if-else / switch()

R/predict.R
```diff
@@ -197,14 +195,27 @@ check_pred_type <- function(object, type, ...) {
         glue_collapse(pred_types, sep = ", ", last = " and ")
       )
     )
-  if (type == "numeric" & object$spec$mode != "regression")
+  if (type == "numeric" & object$spec$mode != "regression") {
```
EmilHvitfeldt, line 198:
```suggestion
  if (type == "numeric" && object$spec$mode != "regression") {
```

This suggestion applies to all `if (...)`

hfrick:
Good point - it's run with every `predict()` plus it's a bit more readable with `switch()` now. Thanks!

EmilHvitfeldt (approved):
Love it!

PR #910, opened 2023-03-08
`read_model_info_table()` only once

simonpcouch:
The `read_model_info_table()` helper is called 3 times per fit, and is read from file each time.

``` r
library(tidymodels)

bt <- bootstraps(mtcars, 100)

bm <- 
  bench::mark(
    total = 
      fit_resamples(linear_reg(), mpg ~ ., bt),
    read_model_info_table = 
      replicate(300, parsnip:::read_model_info_table()),
    check = FALSE
  )
#> Warning: Some expressions had a GC in every iteration; so filtering is
#> disabled.

bm
#> # A tibble: 2 × 6
#>   expression                 min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>            <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 total                    5.94s    5.94s     0.168      57MB     8.92
#> 2 read_model_info_table 215.87ms 219.19ms     4.57     25.1MB     3.04
```

As a percentage of total time, the check takes:

``` r
#> [1] 3.887301
```

This PR proposes, instead, assigning the function output directly into the namespace once and then referring to its output from then on.

<sup>Created on 2023-03-08 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -37,14 +37,8 @@ pred_types <-
 
 # ------------------------------------------------------------------------------
 
-read_model_info_table <- function() {
-  model_info_table <-
-    utils::read.delim(system.file("models.tsv", package = "parsnip"))
-
-  model_info_table
-}
-
-model_info_table <- read_model_info_table()
+model_info_table <-
+  utils::read.delim(system.file("models.tsv", package = "parsnip"))
 
 # ------------------------------------------------------------------------------
 
--- a/R/aaa_models.R
+++ b/R/aaa_models.R
@@ -44,6 +44,8 @@ read_model_info_table <- function() {
   model_info_table
 }
 
+model_info_table <- read_model_info_table()
+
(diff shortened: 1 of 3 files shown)


R/aaa_models.R
```diff
@@ -44,6 +44,8 @@ read_model_info_table <- function() {
   model_info_table
 }
 
+model_info_table <- read_model_info_table()
```
EmilHvitfeldt, line 47: I would remove `read_model_info_table()` and do this code here instead

```r
model_info_table <- utils::read.delim(
  system.file("models.tsv", package = "parsnip")
)
```

EmilHvitfeldt (commented):
Great find! 1 suggetion

R/aaa_models.R
simonpcouch: you got it!

PR #912, opened 2023-03-09
create helper for newdata error

### LINKED ISSUE #525: Create function for newdata alert
Should the following code be wrapped up in an exported helper function?

https://github.com/tidymodels/parsnip/blob/37b07cbe609b6799c4d2a2b2bc3c5430f6b1f835/R/linear_reg.R#L236-L237

It is used a couple of handfuls of times in this package along with other add-on packages. 

``` r
newdata_abort <- function(...) {
  if (any(names(rlang::enquos(...)) == "newdata")) 
   rlang::abort("Did you mean to use `new_data` instead of `newdata`?") 
}

test <- function(...) {
  newdata_abort(...)
  list(...)
}

test(new_data = "hello")
#> $new_data
#> [1] "hello"


test(newdata = "hello")
#> Error: Did you mean to use `new_data` instead of `newdata`?
```

<sup>Created on 2021-07-03 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

simonpcouch:
Closes #525.


EmilHvitfeldt (commented):
Love it!

Is there a specific reason why you put the check in all the `multi_predict.*()` functions instead of in `multi_predict()` itself?

https://github.com/tidymodels/parsnip/blob/main/R/aaa_multi_predict.R#L20-L26

simonpcouch:
I did so to mirror exactly the existing functionality. We do do this check in the `predict.model_fit()` method, though, via `check_pred_type_dots()`, so maybe we should do the same in `multi_predict()`? I'm up for it if you are.

EmilHvitfeldt:
I'm up!

simonpcouch:
Worth mentioning that packages that re-export the `multi_predict()` generic will now receive this check as well, and this is probably worth revdepchecking and dropping in NEWS. Some other packages use a (now-)redundant check:

* bonsai @ https://github.com/tidymodels/bonsai/blob/aab79d511f1ae1d320da6a59335e8ed1f3e50be6/R/lightgbm.R#L354-L356
* rules @ https://github.com/tidymodels/rules/blob/1b27a2f229e7ab355eb3b837e264f655abbbc568/R/cubist_rules.R#L191-L193
* poissonreg @ https://github.com/tidymodels/poissonreg/blob/2ca5869df31db3bee845ac353201e1bc1b1bc78b/R/poisson_reg-glmnet.R#L29-L31
* rules @ https://github.com/tidymodels/rules/blob/1b27a2f229e7ab355eb3b837e264f655abbbc568/R/rule_fit.R#L196-L198
* censored @ https://github.com/tidymodels/censored/blob/f8d2dc1eb15893a8f3c29aa5770035d5be71e22a/R/proportional_hazards-glmnet.R#L252-L253

I do think dropping this into the generic is best for cutting down on technical debt in the long run, but will require a bit more effort. @hfrick, how do you feel about this change? While we're dropping this into a helper, would you make any revisions to the wording of the error?

R/misc.R
```diff
@@ -472,3 +472,13 @@ check_case_weights <- function(x, spec) {
   }
   invisible(NULL)
 }
+
+# -----------------------------------------------------------------------------
+check_for_newdata <- function(..., call = rlang::caller_env()) {
+  if (any(names(rlang::enquos(...)) == "newdata")) {
```
hfrick, line 478:
```suggestion
  if (any(names(list(...)) == "newdata")) {
```
I think we don't need to use quosures here, all we want are the names of things passed in the dots, not the things themselves.

R/misc.R
```diff
@@ -472,3 +472,13 @@ check_case_weights <- function(x, spec) {
...
+    rlang::abort(
+      "Did you mean to use `new_data` instead of `newdata`?",
```
hfrick, line 480:
```suggestion
      "Pleaes use `new_data` instead of `newdata`.",
```
I don't feel particularly strongly about this but it would avoid assuming anything about the user's intent other than which arg they are trying to pass.

hfrick (approved):
I think the generic is a good place for this check to live in 👍 +1 also to revdeps and NEWS item

R/misc.R
simonpcouch: You got it!

PR #916, opened 2023-03-09
warn with >2 levels on `logistic_reg()` fit

### LINKED ISSUE #545: parsnip should raise an error when users use logistic_reg() for multiclass classification
Currently parsnip does not throw an error if a user chooses `logistic_reg()` to fit a multiclass classification model, but the result is wrong and misleading. Moreover, the documentation doesn't mention that `logistic_reg()` should only be used for binary classification. It took me a while to figure out that I should use `multinom_reg()` instead. It would be nice if parsnip can prevent this usage explicitly.

``` r
library(parsnip)
fit <- logistic_reg() %>%
  set_engine("glm") %>%
  fit(Species ~ ., data = iris)
#> Warning: glm.fit: algorithm did not converge
#> Warning: glm.fit: fitted probabilities numerically 0 or 1 occurred
predict(fit, new_data = iris)
#> # A tibble: 150 x 1
#>    .pred_class
#>    <fct>      
#>  1 setosa     
#>  2 setosa     
#>  3 setosa     
#>  4 setosa     
#>  5 setosa     
#>  6 setosa     
#>  7 setosa     
#>  8 setosa     
#>  9 setosa     
#> 10 setosa     
#> # … with 140 more rows
```

<sup>Created on 2021-08-23 by the [reprex package](https://reprex.tidyverse.org) (v2.0.0)</sup>

simonpcouch:
Closes #545.🐧 Warns rather than errors since some engines bin the non-first-level outcome levels as the non-event.

Locating this check inside of `check_outcome` rather than `translate.logistic_reg()` since we don't yet have access to the outcome at the point. :/

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,7 +9,7 @@
 * Fixed bug with prediction from a boosted tree model fitted with `"xgboost"` using a custom objective function (#875).
 
 * Several internal functions (to help work with `Surv` objects) were added as a standalone file that can be used in other packages via `usethis::use_standalone("tidymodels/parsnip")`. 
-
+* `logistic_reg()` will now warn at `fit()` when the outcome has more than two levels (#545).
 
 # parsnip 1.0.4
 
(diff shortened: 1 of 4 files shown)


hfrick (approved):
Whoop! 🐬

PR #917, opened 2023-03-09
correct documentation for `...` in `keras_mlp()`

### LINKED ISSUE #777: Custom loss function for keras (or mlp)
## Feature

(1) When training a mlp() (with 'keras' engine), it would be great if we could provide a custom loss function, similar way as we set metric.
This would open up the opportunity for different customizations or event uncertainty quantification with different objective functions.

(2) I saw a post about introducing keras callbacks (https://github.com/tidymodels/parsnip/issues/70#issuecomment-426459570), as early_stopping, I couldn't figure out how to do this. It just throws an error for me that the "others" is a unused argument.

Thank you!

simonpcouch:
Closes #777.🐊

### DIFF
--- a/R/mlp.R
+++ b/R/mlp.R
@@ -195,7 +195,9 @@ class2ind <- function (x, drop2nd = FALSE) {
 #' @param activation A character string for the type of activation function between layers.
 #' @param seeds A vector of three positive integers to control randomness of the
 #'  calculations.
-#' @param ... Currently ignored.
+#' @param ... Additional arguments to pass to `keras::compile()` or
+#'  `keras::fit()`. Arguments will be sorted automatically---pass these
+#'  directly as main arguments to [set_engine()].
 #' @return A `keras` model object.
 #' @keywords internal
 #' @export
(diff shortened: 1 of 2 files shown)


R/mlp.R
```diff
@@ -195,7 +195,9 @@ class2ind <- function (x, drop2nd = FALSE) {
 #' @param activation A character string for the type of activation function between layers.
 #' @param seeds A vector of three positive integers to control randomness of the
 #'  calculations.
-#' @param ... Currently ignored.
+#' @param ... Additional arguments to pass to `keras::compile()` or
+#'  `keras::fit()`. Arguments will be sorted automatically---pass these
```
EmilHvitfeldt, line 199: I'm not sure i understand the ---? 😄

EmilHvitfeldt (commented):
Looks good! 1 note

R/mlp.R
simonpcouch: They evaluate to an em dash in markdown!

Screen Shot 2023-03-09 at 8 10 47 PM

But... this could probably be worded more straightforwardly. I'll update the wording in a moment!

PR #918, opened 2023-03-09
document that main arguments are captured as quosures

### LINKED ISSUE #662: Unable to pass NULL to boost_tree model
Passing NULL into a boost_tree model results in an error. For the underlying reason and solution see the [rstudio community post](https://community.rstudio.com/t/unable-to-pass-null-to-boost-tree-model/129566). I ran into this issue because I wanted to tune my R model in Python using the `reticulate` package. I was wondering if this could somehow be handled better or mentioned in the documentation as finding what caused these errors took a long time. 

```
library(tidyverse)
library(tidymodels)
library(xgboost)

set.seed(123)
#Load data
df <- as_tibble(iris)
# Split data
split <- initial_split(df, strata = Species)
train <- training(split)
test <- testing(split)
folds <- vfold_cv(train)

depth <- NULL # Defining an integer works fine
is.null(depth)

#Model
xg_model <-
  boost_tree(
    trees = 1000, 
    tree_depth = depth # Directly passing in NULL works fine
  ) %>%
  set_engine("xgboost") %>%
  set_mode("classification")

#Recipe
xg_recipe <-
  recipe(Species ~., data=split)

#Workflow
xg_workflow <-
  workflow() %>%
  add_recipe(xg_recipe) %>%
  add_model(xg_model)

#Fit
xg_fit <- 
  xg_workflow %>%
  last_fit(split) 

xg_fit_rs <- 
  xg_workflow %>% 
  fit_resamples(folds)
accuracy_xg <- xg_fit_rs %>% collect_metrics()
accuracy_xg
```
> [1] TRUE
> x train/test split: preprocessor 1/1, model 1/1: Error in xgb.iter.update(bst$handle, dtrain, iteration - 1, obj): Invalid Parameter format for max_de...

**Solution**

Use injection operator from rlang.

```
xg_model <-
  boost_tree(
    trees = 1000, 
    tree_depth = !!depth
  ) %>%
```

simonpcouch:
Closes #662.🦀

If we think the `model_type()` pseudocode may be confusing for folks, we could make this into a proper template that could take the name of the model type function as an input. :)

### DIFF
--- a/man-roxygen/spec-details.R
+++ b/man-roxygen/spec-details.R
@@ -6,3 +6,12 @@
 #'
 #' The model is not trained or fit until the [`fit()`][fit.model_spec()] function is used
 #' with the data.
+#'
+#' Each of the arguments defaulting to `NULL` in this function are captured
+#' as [quosures][rlang::`topic-quosure`]. To pass values programmatically,
+#' use the [injection operator][rlang::`!!`] like so:
+#'
+#' ``` r
+#' value <- 1
+#' model_type(argument = !!value)
+#' ```
--- a/man/C5_rules.Rd
+++ b/man/C5_rules.Rd
@@ -45,6 +45,14 @@ arguments.
 
 The model is not trained or fit until the \code{\link[=fit.model_spec]{fit()}} function is used
 with the data.
+
+Each of the arguments defaulting to \code{NULL} in this function are captured
+as \link[rlang:`topic-quosure`]{quosures}. To pass values programmatically,
+use the \link[rlang:`!!`]{injection operator} like so:
+
+\if{html}{\out{<div class="sourceCode r">}}\preformatted{value <- 1
+model_type(argument = !!value)
+}\if{html}{\out{</div>}}
 }
 \examples{
 show_engines("C5_rules")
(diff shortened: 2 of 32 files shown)


man-roxygen/spec-details.R
```diff
@@ -6,3 +6,12 @@
 #'
 #' The model is not trained or fit until the [`fit()`][fit.model_spec()] function is used
 #' with the data.
+#'
+#' Each of the arguments defaulting to `NULL` in this function are captured
```
hfrick, line 10: I think the pattern is "every arg apart from `mode` and `engine`" but I haven't checked all of them. I did hit `bag_tree()` though which has non-NULL defaults for some main args and all of them get captured as quosures:
https://github.com/tidymodels/parsnip/blob/332de182d87ce3d74d1e41cdf334ee3159cbbb37/R/bag_tree.R#L25-L37

hfrick (commented):
Hm, I think we should change something about that `model_type()` snippet. It using the actual modelling function would be nice. A possibly easier solution is to explicitly say that this is pseudo-code. Either solution is fine with me.

man-roxygen/spec-details.R
simonpcouch: Ah, good point. Thank you!

simonpcouch:
I'm with ya. Let's make this into a proper template and substitute the function name in.

PR #922, opened 2023-03-13
`format_<prediction type>()` functions only format when necessary

### LINKED ISSUE #263: naming prediction columns
`parsnip` gives proper names of predictor columns (per our conventions) but assumes that the incoming columns do not follow the convention. As a results, we end up with column names like `.pred_.pred_Class1`. 

We should detect if the columns are already named correctly.

hfrick:
closes  #263

The main part was done in #268 but since the issue was still open, this PR now extends it to newer prediction types.

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -222,37 +222,19 @@ check_pred_type <- function(object, type, ...) {
 #' @export
 
 format_num <- function(x) {
-  if (inherits(x, "tbl_spark"))
+  if (inherits(x, "tbl_spark")) {
     return(x)
-
-  if (isTRUE(ncol(x) > 1) | is.data.frame(x)) {
-    x <- as_tibble(x, .name_repair = "minimal")
-    if (!any(grepl("^\\.pred", names(x)))) {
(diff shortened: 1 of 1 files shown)


EmilHvitfeldt (commented):
I like the consistancy upgrade 🙌 

Are these similar enough that we can have a helper function do it?

```r
this_would_need_a_name <- function(x, col_name) {
  if (isTRUE(ncol(x) > 1) | is.data.frame(x)) {
    x <- as_tibble(x, .name_repair = "minimal")
    if (!any(grepl(paste0("^\\", col_name), names(x)))) {
      names(x) <- col_name
    }
  } else {
    x <- tibble(unname(x))
    names(x) <- col_name
    x
  }
}
```

hfrick:
Good thought - I'd say so, yes 🙌  I've introduced a helper here and added a follow-up issue #927. We might get to a general `format_parsnip()` helper but I didn't want to overload this PR.

EmilHvitfeldt:
Very nice! I think this is a good improvement

Issue #608, opened 2021-11-20
Move check_control() to control_parsnip.R


EmilHvitfeldt:
This function would be more at home in https://github.com/tidymodels/parsnip/blob/main/R/control_parsnip.R

https://github.com/tidymodels/parsnip/blob/dd36a954d58c73b67a118e0b79f79f770e9ce703/R/fit.R#L318-L330

PR #928, opened 2023-03-14
move check_control() to where it is used

EmilHvitfeldt:
What the title says

To close https://github.com/tidymodels/parsnip/issues/608


simonpcouch (approved):
Perfect!

PR #929, opened 2023-03-14
speed up `tune_tbl()` helper

simonpcouch:
`tune_tbl()` is called each resample fit, and is a light wrapper around `tibble::tibble()`. The function is only called inside of `tune_args()`, so we know what kinds of inputs the helper will receive:

<https://github.com/tidymodels/parsnip/blob/d8f273cd3eeac3cb2c130109a785fb291bc9e58c/R/tune_args.R#L21-L29>

``` r
bm <- 
  bench::mark(
    total = fit_resamples(linear_reg(), mpg ~ ., bootstraps(mtcars, 100)),
    tune_tbl = replicate(100, parsnip:::tune_tbl()),
    check = FALSE
  )

bm
#> # A tibble: 2 × 6
#> 1 total         3.25s    3.25s     0.308    47.7MB     8.92
#> 2 tune_tbl    43.77ms  47.91ms    22.6        77KB    11.3
```

As a percentage of total time, the helper takes:

``` r
#> [1] 1.474154
```

Since we can recycle vectors as needed before inputting to the helper, we can use `new_tibble(list())` instead of `tibble()`, which will drastically cut down on the time this helper takes.

### DIFF
--- a/R/tune_args.R
+++ b/R/tune_args.R
@@ -18,13 +18,15 @@ tune_args.model_spec <- function(object, full = FALSE, ...) {
   res <- c(arg_id, eng_arg_id)
   res <- ifelse(res == "", names(res), res)
 
+  len_res <- length(res)
+
   tune_tbl(
     name = names(res),
     tunable = unname(!is.na(res)),
     id = res,
-    source = "model_spec",
-    component = model_type,
-    component_id = NA_character_,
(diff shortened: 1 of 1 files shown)


EmilHvitfeldt (approved):
Looks clean!

topepo (approved):

PR #930, opened 2023-03-16
Refactor formatting functions for glmnet methods for `multi_predict()`

hfrick:
This refactoring doesn't change how the functions work but aims to increase readability. The function for logistic regression models already uses this style.

Extratests includes tests on the exact format of predictions (and those run fine).

### DIFF
--- a/R/glmnet-engines.R
+++ b/R/glmnet-engines.R
@@ -223,8 +223,8 @@ multi_predict_glmnet <- function(object,
     "multinom_reg" = format_glmnet_multi_multinom_reg(pred,
                                                       penalty = penalty,
                                                       type = type,
-                                                      n_rows = nrow(new_data),
-                                                      lvl = object$lvl)
+                                                      lvl = object$lvl,
+                                                      n_obs = nrow(new_data))
   )
 
   res
@@ -270,7 +270,6 @@ format_glmnet_multi_linear_reg <- function(pred, penalty) {
 }
 
 format_glmnet_multi_logistic_reg <- function(pred, penalty, type, lvl) {
-
   type <- rlang::arg_match(type, c("class", "prob"))
 
   penalty_key <- tibble(s = colnames(pred), penalty = penalty)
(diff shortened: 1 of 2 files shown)


R/glmnet-engines.R
```diff
@@ -248,26 +248,28 @@ multi_predict._multnet <- multi_predict_glmnet
 multi_predict._glmnetfit <- multi_predict_glmnet
 
 format_glmnet_multi_linear_reg <- function(pred, penalty) {
-  param_key <- tibble(group = colnames(pred), penalty = penalty)
+  penalty_key <- tibble(s = colnames(pred), penalty = penalty)
```
simonpcouch, line 251: Mostly just leaving this here for posterity, as I don't think it's worth the change, but since we know this 1d mapping from the get-go we could just assign to `colnames(pred)` and avoid the join:

``` r
  colnames(pred) <- penalty
  pred <- as_tibble(pred)
  pred$.row <- seq_len(nrow(pred))
  pred <-
    tidyr::pivot_longer(
      pred,
      -.row,
      names_to = "penalty",
      values_to = ".pred",
      names_transform = list(penalty = as.numeric)
    )
  pred <- dplyr::arrange(pred, penalty)
  pred <- tidyr::nest(pred, .by = .row, .key = ".pred")
  pred <- select(pred, -.row)
```

This is surely less readable than what you've written, though, and only a smidge faster—the `as_tibble()` will account for most of the eval time regardless.

simonpcouch (approved):
Looks great--this is definitely an improvement. :)

hfrick:
Just remembered to bump the purrr version - we need >= 1.0.0 for `list_rbind()` which I'm using here. It is already in use in ipcw.R: 
https://github.com/tidymodels/parsnip/blob/e2b9baa8a203da0563440da2b99a719d840a17e1/R/ipcw.R#L224

R/glmnet-engines.R
```diff
@@ -248,26 +248,28 @@ multi_predict._multnet <- multi_predict_glmnet
...
+
   pred <- as_tibble(pred)
-  pred$.row <- 1:nrow(pred)
-  pred <- gather(pred, group, .pred, -.row)
```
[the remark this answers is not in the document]
hfrick: it's been a while since this was last touched

R/glmnet-engines.R
hfrick: Thanks for leaving the speed-up comments here! 🙏  If we're feeling some pain regarding speed here, I'd be all for changing it again. I tried to have a bridge between old and new in each step of changing it to make reviewing (and re-reading) easier, but happy to pick it up again if called for!

PR #932, opened 2023-03-17
speed up `spec_is_possible()` helper

simonpcouch:
My old helper was slow🦥 Instead of joining, we can perform the filtering on the separate datasets, and with vctrs instead of dplyr. For a quick model fit on a small dataset, this cuts down fit time by ~22%!

With `main` dev:

``` r
library(parsnip)

bench::mark(
  total = fit(linear_reg(), mpg ~ ., mtcars),
  iterations = 25
)
#> # A tibble: 1 × 6
#> 1 total        7.65ms   8.04ms      120.    7.19MB     67.6
```

With this PR:

``` r
#> # A tibble: 1 × 6
#> 1 total        6.09ms   6.28ms      155.    7.12MB     60.3
```

<sup>Created on 2023-03-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

There are additional tests for this infrastructure [in extratests](https://github.com/tidymodels/extratests/blob/main/tests/testthat/test-parsnip-extension-messaging.R)—they are fine. :)

### DIFF
--- a/R/misc.R
+++ b/R/misc.R
@@ -97,25 +97,40 @@ spec_is_possible <- function(spec,
                              user_specified_mode = spec$user_specified_mode) {
   cls <- class(spec)[[1]]
 
-  all_model_info <-
-    dplyr::full_join(
-      model_info_table,
-      rlang::env_get(get_model_env(), cls) %>% dplyr::mutate(model = cls),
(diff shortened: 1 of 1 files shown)


EmilHvitfeldt (approved):
looking clean!

PR #935, opened 2023-03-20
speed up `prepare_data()` helper

simonpcouch:
Together with #934, puts us under 100% overhead in the causal estimate bootstrapping example. :)

With `main` dev:

``` r
library(tidymodels)

lr <- fit(logistic_reg(), Class ~ ., two_class_dat)

bench::mark(
  old = predict(lr, two_class_dat, type = "prob")
)
#> # A tibble: 1 × 6
#> 1 old          3.01ms   3.12ms      305.    3.62MB     6.27
```

With this PR:

``` r
#> # A tibble: 1 × 6
#> 1 new          2.14ms   2.44ms      398.   84.78KB     6.19
```

<sup>Created on 2023-03-20 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -394,12 +394,12 @@ prepare_data <- function(object, new_data) {
     new_data <- .convert_xy_to_form_new(object$preproc, new_data)
   }
 
-  remove_intercept <- get_encoding(class(object$spec)[1])
+  encodings <- get_encoding(class(object$spec)[1])
   remove_intercept <-
     vctrs::vec_slice(
-      remove_intercept$remove_intercept,
-      remove_intercept$mode == object$spec$mode &
-        remove_intercept$engine == object$spec$engine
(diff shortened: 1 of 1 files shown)


R/predict.R
```diff
@@ -394,12 +394,16 @@ prepare_data <- function(object, new_data) {
     new_data <- .convert_xy_to_form_new(object$preproc, new_data)
   }
 
+  remove_intercept <- get_encoding(class(object$spec)[1])
```
hfrick, line 397:
```suggestion
  encodings <- get_encoding(class(object$spec)[1])
```

hfrick (commented):
Looks good, I would just suggest changing the name on one object since `remove_intercept` a) suggests a boolean and b) is only a part of that object (the whole tibble on encodings).

PR #936, opened 2023-03-22
`time` to `eval_time`

### LINKED ISSUE #892: Change argument name for evaulation time of predicted survival probabilities
We have settled on `eval_time` as the argument name for the time point(s) at which to evaluate a dynamic (time-dependent) performance metric. For the prediction types `"survival"` and `"hazard"`, we have an argument named `time` here in parsnip which we'll rename to `eval_time` for consistency.

(Note: this has implications in censored.)

hfrick:
Closes #892

This PR deprecates the `time` argument to the prediction functions in favor of the `eval_time` argument name.

- `predict()` only takes this argument in the ellipsis, so I updated only the documentation. I replaced `time` with `eval_time`, i.e. there is no mention of any deprecation but instead a direct swap. Should this talk about deprecation as well? I was trying not to overload that help page since it contains so much already.
- `predict_survival()` and `predict_hazard()` both had a `time` argument. Now they have both `time` and `eval_time` as arguments, and `time` is deprecated, including updates to the documentation.

This PR starts the deprecation at `deprecate_warn()` rather than a soft-deprecation. Does that sound good?

Tests on this deprecation are going to have to sit in censored since we need an engine for those prediction types. parsnip does contain `surv_reg()`, but that uses the mode `"regression"` for which we don't predict survival or hazard probabilities.

### DIFF
--- a/R/predict.R
+++ b/R/predict.R
@@ -31,7 +31,7 @@
 #'            linear predictors). Default value is `FALSE`.
 #'     \item `quantile`: for `type` equal to `quantile`, the quantiles of the
 #'            distribution. Default is `(1:9)/10`.
-#'     \item `time`: for `type` equal to `"survival"` or `"hazard"`, the
+#'     \item `eval_time`: for `type` equal to `"survival"` or `"hazard"`, the
 #'            time points at which the survival probability or hazard is estimated.
 #'  }
 #' @details For `type = NULL`, `predict()` uses
@@ -48,7 +48,7 @@
 #'
 #'  ## Censored regression predictions
 #'
-#' For censored regression, a numeric vector for `time` is required when
+#' For censored regression, a numeric vector for `eval_time` is required when
 #' survival or hazard probabilities are requested. Also, when
 #' `type = "linear_pred"`, censored regression models will by default be
 #' formatted such that the linear predictor _increases_ with time. This may
@@ -83,11 +83,11 @@
 #'
 #' For `type = "survival"`, the tibble has a `.pred` column, which is
 #'  a list-column. Each list element contains a tibble with columns
-#'  `.time` and `.pred_survival` (and perhaps other columns).
+#'  `.eval_time` and `.pred_survival` (and perhaps other columns).
 #'
 #' For `type = "hazard"`, the tibble has a `.pred` column, which is
 #'  a list-column. Each list element contains a tibble with columns
-#'  `.time` and `.pred_hazard` (and perhaps other columns).
+#'  `.eval_time` and `.pred_hazard` (and perhaps other columns).
 #'
 #' Using `type = "raw"` with `predict.model_fit()` will return
 #'  the unadulterated results of the prediction function.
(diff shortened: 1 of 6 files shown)


R/predict.R
```diff
@@ -342,7 +343,15 @@ check_pred_type_dots <- function(object, type, ..., call = rlang::caller_env())
...
+  # places where eval_time should not be given
+  if (any(nms == "eval_time") & !type %in% c("survival", "hazard")) {
+    rlang::abort(
+      paste(
+        "`eval_time` should only be passed to `predict()` when `type` is one of:",
+        paste0("'", c("survival", "hazard"), "'", collapse = ", ")
+      )
+    )
+  }
   if (any(nms == "time") & !type %in% c("survival", "hazard")) {
```
EmilHvitfeldt, line 355: If I understand it correctly, then the error message below, should be updated to also include that `time` is not longer used and `eval_time` is now the thing

EmilHvitfeldt (approved):

EmilHvitfeldt:
> Should this talk about deprecation as well?

I don't think it would be needed. We error pretty well, and it will clog up the documentation. Also, although we have been working on this for a while, it is also fairly experimental and this argument name change shouldn't be too unexpected

R/predict.R
hfrick: I assume that that the `type` that the user provides is the better indication of what they are trying to do than the fact that they supplied a `time` arg. Based on that, I would leave the message as "remove the time arg" rather than give additional info which I assume is not directly relevant in this context.

topepo:
Can you do a quick `bench::mark()` on prediction with and without the new deprecation notices? I want to avoid negating any speed gains.

hfrick:
If you trigger the deprecation message, it is slower, as expected. If you switch to the new argument name, avoiding to trigger that warning, the speed is the same.

With `main` for both parsnip and censored:

``` r
library(parsnip)
library(censored)

bench::mark(
  main = fit(survival_reg(), Surv(time, status) ~ age + sex, data = lung) %>% 
    predict(lung, type = "survival", time = c(100, 200)), 
  iterations = 25
)
#> # A tibble: 1 × 6
#> 1 main         47.5ms   50.7ms      19.6      13MB     18.0
```

<sup>Created on 2023-03-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

With the PRs for both parsnip and censored:

``` r
library(parsnip)
library(censored)

bench::mark(
  dev_with_warning = fit(survival_reg(), Surv(time, status) ~ age + sex, data = lung) %>%
    predict(lung, type = "survival", time = c(100, 200)),
  dev_without_warning = fit(survival_reg(), Surv(time, status) ~ age + sex, data = lung) %>% 
    predict(lung, type = "survival", eval_time = c(100, 200)), 
  iterations = 25
)
#> Warning: The `time` argument of `predict_survival()` is deprecated as of parsnip
#> 1.0.4.9005.
#> ℹ Please use the `eval_time` argument instead.
#> ℹ The deprecated feature was likely used in the parsnip package.
#>   Please report the issue at <https://github.com/tidymodels/parsnip/issues>.
#> This warning is displayed once every 8 hours.
#> Call `lifecycle::last_lifecycle_warnings()` to see where this warning was
#> generated.
#> # A tibble: 2 × 6
#>   expression               min   median `itr/sec` mem_alloc `gc/sec`
#>   <bch:expr>          <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl>
#> 1 dev_with_warning      54.3ms   56.9ms      17.5    16.7MB     16.2
#> 2 dev_without_warning   48.2ms   49.6ms      19.6     967KB     15.4
```

<sup>Created on 2023-03-23 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

topepo (approved):

PR #942, opened 2023-03-29
transition from deprecated `map_dfr()`

simonpcouch:
In favor of `map() %>% list_rbind()`. The deprecation machinery there had resulted in some slowdowns. Note that we already require the needed purrr version:

https://github.com/tidymodels/parsnip/blob/07525af8506b16e0bc963bcfd936b6d02eae07ff/DESCRIPTION#L33


EmilHvitfeldt (approved):
Looking good! clean usage

PR #943, opened 2023-03-29
transition v0.1.6 and v0.1.8 `soft`-deprecations to `warn`

simonpcouch:
It has now been at least a year since each of these were soft-deprecated. :)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,8 @@
 
 * The minimum version for R is now 3.5 (#926).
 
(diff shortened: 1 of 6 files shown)


hfrick (approved):

simonpcouch:
Wanted to make sure we weren't introducing a hitch right before we start the release process, but looks like we're all good!

```
We checked 49 reverse dependencies, comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 0 new problems
```

🥑

PR #944, opened 2023-03-29
don't use colon programatically

EmilHvitfeldt:
I noticed there are a couple of places where we use `:` that could be replaced with `seq_*()` to make the code a little more robust


hfrick (approved):
Yes! 🌊

PR #955, opened 2023-04-28
add helper for bridging causal fits

simonpcouch:
PR 1/3 to address tidymodels/tune#652. This PR implements the generic and the `model_fit` method--note that this method isn't dispatched to from the `workflow` method (and thus the `tune_results` method).


R/weight_propensity.R
```diff
@@ -0,0 +1,90 @@
...
+                                        .treated = object$lvl[2],
+                                        ...,
+                                        data) {
+  if (rlang::is_missing(wt_fn) || !is.function(wt_fn)) {
+    abort("`wt_fn` must be a function.")
+  }
+
+  if (rlang::is_missing(data) || !is.data.frame(data)) {
+    abort("`data` must be the data supplied as the data argument to `fit()`.")
+  }
```
simonpcouch, line 75: The only checks this PR makes on inputted data is that it's a data frame. This gives a window for folks to supply different data to `fit()` and `weight_propensity()`, probably resulting in uninformative errors at `weight_propensity()`. We could implement some functionality to "fingerprint" the training data, noting dims/column names or some other coarse set of identifying features, as a full hash would be too expensive to justify. Note that this is not an issue for the `tune_results` method (and its use of the workflow method), which accesses the training data via the underlying split and does not take a `data` argument.

R/weight_propensity.R
```diff
@@ -0,0 +1,90 @@
...
+    "i" = "Supply `object` to `fit()` before generating propensity weights."
+  ))
+}
+
+#' @rdname weight_propensity.model_fit
+#' @method weight_propensity model_fit
+#' @export
+weight_propensity.model_fit <- function(object,
+                                        wt_fn,
+                                        .treated = object$lvl[2],
```
simonpcouch, line 66: By default, propensity assumes that the treatment level of the exposure is the second level, counter to the `event_level` default. I've tried out a bunch of different interfaces to this argument and am unsure of how we can best handle this. We could alternatively add an `event_level` argument with the usual parameterization and then "translate" it to the right `.treated` level when interfacing with propensity. Seeing if we could have that behavior adjusted in propensity might also make this feel better.

Note that the current form of that argument is not checked / tested, pending a decision on how we want it to feel.

R/weight_propensity.R
```diff
@@ -0,0 +1,90 @@
...
+#' propensity and outcome models in causal workflows. **The main documentation
+#' for this function lives in the tune package at** `?tune::weight_propensity`.
+#'
+#' @param object The object containing the model fit(s) that will generate
+#' predictions used to calculate propensity weights. Currently, either a
+#' [parsnip model fit][parsnip::fit.model_spec()], fitted
+#' [workflow][workflows::workflow()], or
+#' tuning results (`?tune::fit_resamples`) object. If a tuning result, the
+#' object must have been generated with the control argument
+#' (`?tune::control_resamples`) `extract = identity`.
```
simonpcouch, line 14: Another option would be to instead require `save_pred = TRUE`, but we couldn't make use of `weight_propensity.workflow` in that case. This approach is a bit more DRY.

hfrick: This feels like a little bit of a rough edge to me. I'm not sure we need to sand over it right now in terms of the interface but I would add more documentation, especially on the "main" doc page in tune, which currently only mentions this in an example. What about adding a sentence to the Details section, explaining _why_ this needs to be set like this? I think that would help people remember.

R/weight_propensity.R
hfrick: I think that's worth discussing with Lucy and Malcolm

hfrick (commented):
Nice! Most of this interface/bridge now feels buttery-smooth, well done!

R/weight_propensity.R
malcolmbarrett: This is all awesome!

A few comments:

1) we noticed that in propensity we also use the `.treated` terminology but now think that's a poor idea because not everything is a treatment. So, we're going to change that to `.exposed`, and I think it should probably be that here, too.
2) One issue with that language and with that default value for what is currently `.treated` is that it only applies to binary and multiclass variables. For continuous variables, there won't be a default level. If this really becomes an issue to make it fit nicely, we're actually moving away from PS models for continuous exposures because of some mathematical issues with them.
3) As for the default value, we do pick the second level because we assume 0 is unexposed and 1 is exposed, but that's mostly to do with what a common logistic model spec looks like. I'm going back to propensity soon and would be happy to work with you all to make this all consistent.

PR #964, opened 2023-05-10
Add `call` arg to checking functions in standalone-survival

hfrick:
This is possibly a smidge easier to read in commit-order instead of aggregated:

- the first commit just moves functions into an order in which they build on each other, for readability
- the second commit adds a call arg to functions that may call `rlang::abort()` so that we can pass the call through

The tests are in extratest and don't change because they either don't capture those error messages (yet) or because the call does not get passed on (yet).


EmilHvitfeldt (commented):
Very nice!

PR #970, opened 2023-05-15
Remove redundant test on `trunc_probs()`

hfrick:
I think the test removed in this PR is mainly to test aspects which are already covered by the other test on `trunc_probs()`. I'll comment with links :)

### DIFF
--- a/tests/testthat/test-survival-censoring-weights.R
+++ b/tests/testthat/test-survival-censoring-weights.R
@@ -22,21 +22,6 @@ test_that("probability truncation via trunc_probs()", {
   )
 })
 
-test_that("trunc_probs()", {
-  probs_1 <- (0:10) / 20
-  probs_2 <- probs_1
-  probs_2[3] <- NA_real_
-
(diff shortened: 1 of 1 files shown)


tests/testthat/test-survival-censoring-weights.R
```diff
@@ -22,21 +22,6 @@ test_that("probability truncation via trunc_probs()", {
   )
 })
 
-test_that("trunc_probs()", {
-  probs_1 <- (0:10) / 20
-  probs_2 <- probs_1
-  probs_2[3] <- NA_real_
-
-  expect_equal(parsnip:::trunc_probs(probs_1, 0), probs_1)
```
hfrick, line 30: this covers "leave probs untouched" and "check the full vector", both of which are covered by https://github.com/tidymodels/parsnip/blob/ab424095b1314b4aaf5136bd150b296646456455/tests/testthat/test-survival-censoring-weights.R#L18-L22

PR #972, opened 2023-05-18
Make `.check_censor_model()` error for non-censored-regression models

### LINKED ISSUE #971: `check_censor_model()` should error for models without mode `"censored regression"`
`.censoring_weights_graf.model_fit()` checks the model fit via `.check_censor_model()` and since censoring weights only make sense for models with mode `"censored regression"`, the check function should error for models with a different mode.

``` r
library(parsnip)
library(survival)

lung2 <- lung %>%
  dplyr::mutate(surv = Surv(time, status), .keep = "unused")

not_a_censor_model <- fit(linear_reg(), mpg ~ ., data = mtcars)

# this should error
parsnip:::.check_censor_model(not_a_censor_model)

# because it's used in and this now gives a different error
.censoring_weights_graf(not_a_censor_model, lung2)
#> Error:
#> ! The input should have a list column called `.pred`.
#> Backtrace:
#>     ▆
#>  1. ├─parsnip::.censoring_weights_graf(not_a_censor_model, lung2)
#>  2. └─parsnip:::.censoring_weights_graf.model_fit(...) at parsnip/R/survival-censoring-weights.R:188:2
#>  3.   └─parsnip:::.check_pred_col(predictions) at parsnip/R/survival-censoring-weights.R:223:2
#>  4.     └─rlang::abort(...) at parsnip/R/survival-censoring-weights.R:48:4
```

<sup>Created on 2023-05-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

hfrick:
closes #971

Regression and classification models also have `$censor_probs` even though it's just an empty list, so this check function didn't error before.

Test is to be updated in extratests.

With this PR:
``` r
library(parsnip)
library(survival)

lung2 <- lung %>%
  dplyr::mutate(surv = Surv(time, status), .keep = "unused")

not_a_censor_model <- fit(linear_reg(), mpg ~ ., data = mtcars)

# this now errors
parsnip:::.check_censor_model(not_a_censor_model)
#> Error in `parsnip:::.check_censor_model()`:
#> ! The model needs to be for mode 'censored regression', not for mode
#>   'regression'.

# so also this errors now with the better message
.censoring_weights_graf(not_a_censor_model, lung2)
#> Error in `.check_censor_model()` at parsnip/R/survival-censoring-weights.R:226:2:
#> ! The model needs to be for mode 'censored regression', not for mode
#>   'regression'.
```

<sup>Created on 2023-05-18 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: parsnip
 Title: A Common API to Modeling and Analysis Functions
-Version: 1.1.0.9002
+Version: 1.1.0.9003
 Authors@R: c(
     person("Max", "Kuhn", , "max@posit.co", role = c("aut", "cre")),
     person("Davis", "Vaughan", , "davis@posit.co", role = "aut"),
--- a/R/survival-censoring-weights.R
+++ b/R/survival-censoring-weights.R
@@ -63,6 +63,11 @@ trunc_probs <- function(probs, trunc = 0.01) {
 }
 
 .check_censor_model <- function(x) {
+  if (x$spec$mode != "censored regression") {
+    cli::cli_abort(
+      "The model needs to be for mode 'censored regression', not for mode '{x$spec$mode}'."
+    )
+  }
   nms <- names(x)
   if (!any(nms == "censor_probs")) {
     rlang::abort("Please refit the model with parsnip version 1.0.4 or greater.")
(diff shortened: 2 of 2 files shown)


hfrick:
An alternative is to make changes in `fit()` and `fit_xy()` to only fit the additional model for censoring probabilities if the main model has mode censored regression. On the one hand, the `model_fit` objects then only carry what they need. On the other hand, the `model_fit` objects look somewhat different on what the mode is, was there hesitation around type-stability there? Do you have a preference for either of the two options?

PR #973, opened 2023-05-19
survival time to binary conversion

topepo:
This exports three functions in the standalone file (one function is new). It exports them in a separate file that uses some razzle-dazzle from the sources for with::defer. 

Unit tests are here; no more needed in extratests for the new function.


hfrick:
Cool, I'll do that and solve the merge conflict in NEWS. Once that's done, I'll merge this PR 👍

PR #981, opened 2023-06-14
add to `predict()` docs how to get the IPCW

### LINKED ISSUE #979: documentation for dynamic survival probabilities
* `predict.model_fit()` will generate a data frame with a list column with predictions at each time point.
* `censoring_weights_graf.model_fit()` can take the prediction results and add columns related to IPCW
* `augment.model_fit()` will do both (if there is an outcome column). 

We need to document this in the functions to reduce confusion. The main point to make related to `predict()` is that this S3 method does not ever involve the outcome (which is needed to compute censoring weights). 

If there is an outcome, we should document that the required columns for yardstick can be made using two steps (`predict()` then `censoring_weights_graf()`) or with one step (`augment()`).

hfrick:
closes #979

I've not included links to `.censoring_weights_graf()` like in the issue description because that function is `@keyword internal` and I think it would be too much detail on the predict help page. If we think it's super important to talk about that internal function in the docs then I would add it to the help page for `augment()` (and consider making the function not internal) - but I tend to think that `augment()` is the right place to point a user to.

### DIFF
--- a/R/augment.R
+++ b/R/augment.R
@@ -27,12 +27,15 @@
 #'
 #' If survival predictions are created and `new_data` contains a
 #' [survival::Surv()] object, additional columns are added for inverse
-#' probability of censoring weights (IPCW) are also created. This enables the
-#' user to compute performance metrics in the \pkg{yardstick} package.
+#' probability of censoring weights (IPCW) are also created (see `tidymodels.org`
+#' page in the references below). This enables the user to compute performance
+#' metrics in the \pkg{yardstick} package.
 #'
 #' @param new_data A data frame or matrix.
 #' @param ... Not currently used.
 #' @rdname augment
+#' @references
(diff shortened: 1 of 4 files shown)


simonpcouch (approved):
Very clearly phrased—thumbs up from me!

I agree that we ought not to link to internal functions in these docs, though am not familiiar enough with the context to have an opinion on whether `censoring_weights_graf.model_fit()` ought to be made user-facing.

topepo:
Internal or not is a grey area for that particular function. I think that I'd prefer to leave it as internal. 

Would it help to link to the (upcoming) tidymodels.org articles on the dynamic predictions?

hfrick:
I agree with leaving it as internal for now. Looks like the CI failures are unrelated. Should we merge?

topepo:
I'll give them an hour or two and then see if `curl` is working again.

PR #997, opened 2023-08-08
Make testing more quiet

EmilHvitfeldt:
This isn't the biggest deal in the world. But the tests can sometimes be a little noisy. This PR fixes what I found.

For example this is what I'm trying to avoid:

Screenshot 2023-08-08 at 10 30 50 AM

### DIFF
--- a/R/engines.R
+++ b/R/engines.R
@@ -41,7 +41,7 @@ load_libs <- function(x, quiet, attach = FALSE) {
     if (!attach) {
       suppressPackageStartupMessages(requireNamespace(pkg, quietly = quiet))
     } else {
-      library(pkg, character.only = TRUE)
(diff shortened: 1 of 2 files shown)


R/engines.R
```diff
@@ -41,7 +41,7 @@ load_libs <- function(x, quiet, attach = FALSE) {
     if (!attach) {
       suppressPackageStartupMessages(requireNamespace(pkg, quietly = quiet))
     } else {
-      library(pkg, character.only = TRUE)
+      library(pkg, character.only = TRUE, quietly = quiet)
```
EmilHvitfeldt, line 44: I assume this is okay? we are passing `quiet` to `requireNamespace()` so i think it is okay to do it to `library()` as well

hfrick (approved):
Shhhhhh 🤫

Issue #996, opened 2023-08-08
Adding dann 


gmcmacran:
I would like to add my modeling package to the tidymodels ecosystem. I am working on a major release that refactors the code base how tidymodel's hardhat suggests. Models will have a recipe method for training, and predict returns tidymodel's preferred prediction format. I am holding off on submitting to cran until I get it added to tidymodels.  

Github link: https://github.com/gmcmacran/dann/tree/hardhat

My question are:
1: Are you guys open to the addition of two new models? dann and sub_dann.
2: Should I issue a PR myself?
3: What package should I issue it in? Should I make a PR in parsnip or create a new package similar to poissonreg?
4: Is there a blog post that serves as a good starting point for tidymodels development? Otherwise, I will use poissonreg as a reference.

simonpcouch:
Thanks for the issue! dann looks like a neat package.

This sounds like a great idea for a parsnip extension package. parsnip's infrastructure allows for anyone to create their own package that integrates with parsnip (and thus the rest of the tidymodels ecosystem)! We have some documentation on how to create custom model specifications [on the tidymodels website](https://www.tidymodels.org/learn/develop/models/). We'd recommend creating a new package of your own—poissonreg would indeed be a great place to see how those specifications are situated in a package, and modeltime is an example of another community-contributed package.

Feel free to holler if you run into any issues while building the package or have a finished product you'd like us to feature [on the tidymodels website](https://www.tidymodels.org/find/parsnip/).

gmcmacran:
Excellent! I will take a look at the links and get started.

gmcmacran:
My tidydann package defines an S3 method tunable. https://github.com/gmcmacran/tidydann/blob/cf7208a3a6894ffc4bcc0d4d09f2f1dc81947cb4/R/tidy_dann.R#L50-L112

With my package loaded, tune_grid fails. If I define the same exact function within the R script (outside of the package), tune_grid  works as expected. https://github.com/gmcmacran/tidydann/blob/cf7208a3a6894ffc4bcc0d4d09f2f1dc81947cb4/R/question.R#L18-L37

Can you explain why this is happening and how to fix the problem?

hfrick:
I suspect that might be because `tune_grid()` also has to know about (and load) the package which contains the model and the engine, so `tidydann` itself. That should be declared as one of the dependencies for the engine, in addition to `dann` here:
[...]

PR #1001, opened 2023-10-06
Add dann models

gmcmacran:
This pull request adds the nearest_neighbor_adaptive to parsnip. Engines are the tidydann package.


gmcmacran:
https://github.com/tidymodels/parsnip/issues/996

gmcmacran:
https://github.com/gmcmacran/tidydann/issues/7

simonpcouch:
Thank you for the PR! Your [comment re: dependencies](https://github.com/gmcmacran/tidydann/issues/7#issuecomment-1749949015) makes sense given your interpretation of Max's issue—implementing this approach is indeed tricky.

Max is OOO for a bit now, so I'll be the first touch point. I believe his intent was to suggest that the tidydann package receives the suggested changes rather than parsnip.

By your interpretation, this would involve renaming the `dann()` and `sub_dann()` model types to `nearest_neighbor_adaptive()`, with `"dann"` and `"sub_dann"` being engines. Those changes would happen in tidydann.

A lower-friction way to implement that might be to make `"dann"` and `"sub_dann"` engines for the existing `nearest_neighbor()` model type from parsnip. [Here is an example](https://github.com/tidymodels/bonsai/blob/main/R/lightgbm_data.R) of adding a new engine to an existing model type, where the engine is `"lightgbm"` and the model type is `boost_tree()`. Once you've done so, parsnip itself needs no changes—loading the tidydann package will automatically hook into parsnip's machinery to recognize the engine.

gmcmacran:
Thanks for feedback. I will move the model functions back into tidydann. Closing PR.

PR #1016, opened 2023-11-03
improve error when outcome is missing

### LINKED ISSUE #1003: Better error message if outcome has 0 columns
Extracted from: https://stackoverflow.com/questions/77305226/why-am-i-receiving-the-error-object-not-found

If someone uses `step_dummy()` or some other ways in {recipes} to remove the outcome, then `parsnip:::check_outcome()` throws a wonderfully bad error 😄 

``` r
library(tidymodels)

rec <- recipe(Species ~ ., data = iris) %>%
  step_dummy(Species)

svm_spec <- 
  svm_rbf() %>% 
  set_engine("kernlab") %>% 
  set_mode("classification")

wflow <- workflow() %>% 
  add_model(svm_spec) %>% 
  add_recipe(rec)

fit(wflow, iris)
#> Error: object '.' not found
```

<sup>Created on 2023-10-16 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

simonpcouch:
Closes #1003. :)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # parsnip (development version)
 
+* Improved errors in cases where the outcome column is mis-specified.
+
 # parsnip 1.1.1
 
 * Fixed bug where prediction on rank deficient `lm()` models produced `.pred_res` instead of `.pred`. (#985)
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # parsnip (development version)
 
-* Improved errors in cases where the outcome column is mis-specified.
+* Improved errors in cases where the outcome column is mis-specified. (#1003)
 
 # parsnip 1.1.1
 
--- a/R/misc.R
+++ b/R/misc.R
@@ -354,6 +354,16 @@ check_outcome <- function(y, spec) {
     return(invisible(NULL))
   }
 
+  has_no_outcome <- if (is.atomic(y)) {is.null(y)} else {length(y) == 0}
+  if (isTRUE(has_no_outcome)) {
+    cli::cli_abort(
+      c("!" = "{.fun {class(spec)[1]}} was unable to find an outcome.",
+        "i" = "Ensure that you have specified an outcome column and that it \\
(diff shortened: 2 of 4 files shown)


R/misc.R
```diff
@@ -354,6 +354,16 @@ check_outcome <- function(y, spec) {
     return(invisible(NULL))
   }
 
+  has_no_outcome <- if (is.atomic(y)) {is.null(y)} else {length(y) == 0}
```
simonpcouch, line 357: For `is.atomic(y)`, note `is.atomic(NULL)` is `TRUE`, and for `length(y) == 0`, `y` can either be a list or tibble.

R/misc.R
```diff
@@ -354,6 +354,16 @@ check_outcome <- function(y, spec) {
...
+  if (isTRUE(has_no_outcome)) {
+    cli::cli_abort(
+      c("!" = "{.fun {class(spec)[1]}} was unable to find an outcome.",
+        "i" = "Ensure that you have specified an outcome column and that it \\
+               hasn't been removed in pre-processing."),
+      call = NULL
```
simonpcouch, line 363: Plumbing the `call` here is a bit tricky. We _could_ pass this call all the way along from wherever it came from in parsnip (`fit()` or `fit_xy()`), but that's not necessarily the function called by the user (could be any function from tune, workflowsets, etc). Instead, I opt to call out the name of the model spec, which the user definitely wrote out at some point.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # parsnip (development version)
 
+* Improved errors in cases where the outcome column is mis-specified.
```
EmilHvitfeldt, line 3:
```suggestion
* Improved errors in cases where the outcome column is mis-specified. (#1003)
```

R/misc.R
EmilHvitfeldt: Yes this seems wise. Because there can be many reasons why an outcome didn't get passed to the model

EmilHvitfeldt (approved):

PR #1021, opened 2023-11-06
Document what `mtry = NULL` means for xgboost

### LINKED ISSUE #1017: Clarify xgboost mtry default
I've looked in dials and parsnip and I can't find what mtry actually defaults to. Does it with `count=F` correspond to xgboost's `colsample_bynode`? 
I believe for random forests, mtry default is sqrt(# features), but xgboost by default uses all columns.

hfrick:
closes  #1017


simonpcouch (approved):
A1! Top-tier! Pinnacle of PRs!

hfrick:
😂

PR #1026, opened 2023-11-17
remove new_data argument from predict.censoring_model_reverse_km()

### LINKED ISSUE #965: Deprecate `new_data` arg for `reverse_km` censoring model
The `predict()` method for `censoring_model_reverse_km` models has a `new_data` argument which is passed on to the predict method for prodlim objects. The definition for this arg from prodlim:
> A data frame with the same variable names as those that appear on the right hand side of the 'prodlim' formula. 

However, the right hand side for  `censoring_model_reverse_km` models is always just `1`, no covariates:
https://github.com/tidymodels/parsnip/blob/2a8f92e1a12638c8b2dfe8e47fb56923c752828a/R/survival-censoring-model.R#L27

Because of that, the argument doesn't do anything, you can even feed it a totally unrelated dataset and it won't complain. Do we anticipate adding covariates to that reverse KM anytime soon/ever? If not, can we deprecate that arg to reduce complexity we are not even making use of?

``` r
library(censored)

mod_fit <-
  survival_reg() %>%
  fit(Surv(time, status) ~ age + sex, data = lung)

# can use different data (because reverse_km model uses ~ 1)
pred_newdata <- predict(mod_fit$censor_probs, time = (7:10) * 100,
                        new_data = mtcars)
```

<sup>Created on 2023-05-11 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

EmilHvitfeldt:
To close #965

I removed `new_data` from the function signature since it is swalloed up by `...` anyways. I also added a little warning, since using `new_data` for predict methods are quite a common action

``` r
library(censored)

mod_fit <-
  survival_reg() %>%
  fit(Surv(time, status) ~ age + sex, data = lung)

# can use different data (because reverse_km model uses ~ 1)
pred_newdata <- predict(mod_fit$censor_probs, time = (7:10) * 100,
                        new_data = mtcars)
#> Warning: `new_data` is ignored when predicting on `censoring_model_reverse_km()` models
#> as it doesn't affect the result.
```

<sup>Created on 2023-11-17 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,8 @@
 
 * Fixed documentation for `mlp(engine = "brulee")`: the default values for `learn_rate` and `epochs` were swapped (#1018).
 
+* The `new_data` argument for the `predict()` method of `censoring_model_reverse_km()` has been deprecated (#965).
+
 # parsnip 1.1.1
 
 * Fixed bug where prediction on rank deficient `lm()` models produced `.pred_res` instead of `.pred`. (#985)
--- a/R/survival-censoring-model.R
+++ b/R/survival-censoring-model.R
@@ -65,6 +65,13 @@ predict.censoring_model_reverse_km <- function(object, time, as_vector = FALSE, 
   rlang::check_installed("prodlim", version = "2022.10.13")
   rlang::check_installed("censored", version = "0.1.1.9002")
 
+  if ("new_data" %in% names(list(...))) {
+    cli::cli_warn(
+      "{.arg new_data} is ignored when predicting on \\
+      {.fn censoring_model_reverse_km} models as it doesn't affect the result."
+    )
+  }
+
   res <- rep(NA_real_, length(time))
   if (length(time) == 0) {
     return(res)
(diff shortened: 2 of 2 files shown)


hfrick (changes requested):
`predict.censoring_model_reverse_km()` already went to CRAN so typically we'd deprecate instead of directly remove. 

It's pretty obscure so I'd say a `deprecate_stop()` is okay instead of kicking things off with `deprecate_warn()`. 

If we're not using the dots, we should check that they are empty. But I don't mind if we make that another issue 🗒️

EmilHvitfeldt:
I switched to `deprecate_stop()` now

``` r
library(censored)

mod_fit <-
  survival_reg() %>%
  fit(Surv(time, status) ~ age + sex, data = lung)

pred_times <- (7:10) * 100
pred_df <- predict(mod_fit$censor_probs, time = pred_times, new_data = lung)
#> Error:
#> ! The `new_data` argument of `predict.censoring_model_reverse_km()` was
#>   deprecated in parsnip 1.2.0 and is now defunct.
```

PR #1034, opened 2023-12-08
move .filter_eval_time to standalone

topepo:
As discussed in tidymodels/tune#780

The error and warning both use cli now so we might see some extratest breakages. I will investigate that.


hfrick (approved):
All `.filter_eval_time()` related changes are 👍  

The change in `.check_cens_type()` looks like a linting change to me but it causes a snaphot test in extratest to break so I would suggest undoing that one.

PR #1035, opened 2023-12-08
Don't modify spark data in `form_form()`

hfrick:
#1033 added an encodings-based removal of the intercept column to `form_form()`, which broke the spark case-weights tests in extratests.

So this PR moves that step so that we don't try this for spark data.


hfrick:
https://github.com/tidymodels/extratests/pull/151 is running again with the changes in this PR

Issue #968, opened 2023-05-15
Move `.censoring_weights_graf.workflow()` to workflows


hfrick:
We try to keep methods in the package that defines the class, we should move `.censoring_weights_graf.workflow()` to workflows. Tests will still need to be in extratests since we need censored to fit a workflow with a censored regression mode.

PR #1038, opened 2023-12-12
Move .censoring weights graf.workflow

EmilHvitfeldt:
to close https://github.com/tidymodels/parsnip/issues/968

Paired {workflows} PR: https://github.com/tidymodels/workflows/pull/213

There are 4 possible scenarios:

## `.censoring_weights_graf.workflow()` is exported from only {parsnip}

This is fine!

Happens if `parsnip < 1.1.1.9007` and `workflows < 1.1.3.9001`, which is another way of saying that both packages are installed before Dec 12.

## `.censoring_weights_graf.workflow()` is exported from only {workflows}

This is fine!

Happens if `parsnip >= 1.1.1.9007` and `workflows >= 1.1.3.9001`, which is another way of saying that both packages are installed after Dec 12 (or whenever the PRs are merged in.

## `.censoring_weights_graf.workflow()` is exported from both {parsnip} and {workflows}

This is not good.

Would need `parsnip < 1.1.1.9007` and `workflows >= 1.1.3.9001`. Which can't happen as workflows `1.1.3.9001` requires `parsnip >= 1.1.1.9007`. So we are good here.

##`.censoring_weights_graf.workflow()` isn't exported

Would happen if `parsnip >= 1.1.1.9007` and `workflow < 1.1.3.9001`. Which could happen. I can't recall if there is a way for us to deal with this last point

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: parsnip
 Title: A Common API to Modeling and Analysis Functions
-Version: 1.1.1.9005
+Version: 1.1.1.9007
 Authors@R: c(
     person("Max", "Kuhn", , "max@posit.co", role = c("aut", "cre")),
     person("Davis", "Vaughan", , "davis@posit.co", role = "aut"),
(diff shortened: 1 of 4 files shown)


hfrick (approved):
Looks good! We have previously done a more complicated approach with conditionally registering the moved method but since this method has not gone to CRAN yet, I think the approach via requiring a fixed version of parsnip is sufficient here.

PR #1048, opened 2024-01-16
deprecate `rpart_train()`

### LINKED ISSUE #1044: Deprecate `rpart_train()`
I don't think we use that function anywhere (anymore): https://github.com/search?q=org%3Atidymodels%20rpart_train&type=code

Let's deprecate and remove.

simonpcouch:
Closes #1044.🌲🚃

My main question here is whether we should port some of this argument documentation to the `details_decision_tree_rpart` help-file. Much of these docs seem relevant to `rpart_train()` rather than rpart in general.


R/decision_tree.R
```diff
@@ -165,6 +168,12 @@ check_args.decision_tree <- function(object) {
 #' @export
 rpart_train <-
   function(formula, data, weights = NULL, cp = 0.01, minsplit = 20, maxdepth = 30, ...) {
+    lifecycle::deprecate_warn(
```
simonpcouch, line 171: Starting with a more aggressive deprecation as, from what I can tell, `rpart_train()` [was never used](https://github.com/tidymodels/parsnip/commit/4cb1494fccd80009e0336c2b1fa3cfffe60166e2#diff-467048fe7785c21fd18aa72a36b87e4b949a9afee603690cb1629f9dcfb1c4f7), even when it was first committed.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # parsnip (development version)
 
+* `rpart_train()` has been deprecated in favor of using `decision_tree()` with the `"rpart"` engine or `rpart::rpart()` directly (#1044).
```
simonpcouch, line 3: Some of these bullets have the parens before the periods, others not. We can fix before release.☃️

hfrick: The style guide says to put them [before the period](https://style.tidyverse.org/news.html#acknowledgement).

EmilHvitfeldt: I'm so opposed to that part of the style guide 😄

topepo (approved):

PR #1059, opened 2024-01-22
export `check_args()`

### LINKED ISSUE #1057: Export `check_args()` for developers of custom models
## Feature

`check_args()` is an internal generic function that is used by parsnip models to validate that parameters have reasonable values before a model is fit.

For developers of [custom parsnip models](https://www.tidymodels.org/learn/develop/models/), it would be nice if this function was exported so that appropriate S3 methods could be written to validate parameters on any new models.

simonpcouch:
Closes #1057.


hfrick (approved):
I agree with exporting this. It'll only help for new models but not new engines but that's reason enough for me 👍

PR #1062, opened 2024-02-02
handle attributes on outcome columns

### LINKED ISSUE #1060: type checks fail uninformatively when outcome has `"label"` attribute
```
library(tidyverse)
library(WDI)

#Dataset for methane emissions
df_methane <- 
  WDI(indicator = "EN.ATM.METH.KT.CE", 
      extra = TRUE) %>% 
  as_tibble() %>% 
  janitor::clean_names() %>% 
  drop_na() %>% 
  rename(methane = en_atm_meth_kt_ce) 

#Modeling
library(tidymodels)

#Dataset for modeling
df_mod <- 
  df_methane %>% 
  filter(region != "Aggregates",
         income != "Not classified") %>% 
  mutate(latitude = latitude %>% as.numeric(),
         longitude = longitude %>% as.numeric(),
         income = income %>% as_factor(),
         region = region %>% as_factor()) %>% 
  select(region, income, longitude,latitude, methane) 


#Splitting
set.seed(12345)
df_split <- initial_split(df_mod, 
                          prop = 0.8,
                          strata = "income")

df_train <- training(df_split)
df_test <- testing(df_split)

tidymodels/workflows#10-fold cross validation for tuning
set.seed(12345)
df_fold <- 
  vfold_cv(df_train,
           strata = income,
           repeats = 5)

#Linear regression models for different engines/packages
spec_lm <- 
  linear_reg() %>% 
  set_engine("lm") 


spec_glm <- 
  linear_reg() %>% 
  set_engine("glm")

spec_glmnet <- 
  linear_reg(penalty = tune(),
             mixture = tune()) %>% 
  set_engine("glmnet")

spec_keras <- 
  linear_reg(penalty = tune()) %>% 
  set_engine("keras")

spec_stan <- 
  linear_reg() %>% 
  set_engine("stan")

#Workflow set
basic_recipe <-
  recipe(methane ~ ., data = df_train) 

all_workflows <- 
  workflow_set(
    preproc = list(basic = basic_recipe), 
    models = list(LM = spec_lm, 
                  GLM = spec_glm,
                  GLMNET = spec_glmnet,
                  Keras = spec_keras,
                  Stan = spec_stan)
  )

#Tuning and evaluating the models
grid_ctrl <-
  control_grid(
    save_pred = TRUE,
    parallel_over = "everything",
    save_workflow = TRUE
  )

grid_results <-
  all_workflows %>%
  workflow_map(
    seed = 98765,
    resamples = df_fold,
    grid = 15,
    control = grid_ctrl
  )

#one of the error lines:  
# `y` should be one of the following classes: 'data.frame', 'matrix', 'factor', 'Surv'
```

### LINKED ISSUE #1061: subsetting of outcome when outcome has label
When I was trying to reproduce #1060, this was my first go at a minimal reprex. There's no error here (at least for this engine), but parsnip takes only the first observation (and I'm assuming it's recycled later on?)

``` r
library(parsnip)

fit_xy(
  linear_reg(),
  data.frame(x = 1:5),
  y = structure(2:6, label = "hi")
)
#> parsnip model object
#> 
#> 
#> Call:
#> stats::lm(formula = ..y ~ ., data = data)
#> 
#> Coefficients:
#> (Intercept)            x  
#>   2.000e+00    1.573e-16
```

<sup>Created on 2024-02-02 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

Notice intercept is 2 and slope is basically zero. This is because the outcome is always `2:6[[1]]`😬

Offending code:

https://github.com/tidymodels/parsnip/blob/8ccf1be367daab2cc21a343b222e8ce417ba9be7/R/fit.R#L270-L276

simonpcouch:
We use `is.atomic()` in `check_outcome()` rather than `is.vector()` as it more often aligns with what we mean when we're checking for outcome types. I've transitioned those type checks in a few more places. Closes #1060, closes #1061.

### DIFF
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -234,7 +234,7 @@
   if (is.matrix(y)) {
     y <- as.data.frame(y)
   } else {
-    if (is.vector(y) | is.factor(y)) {
+    if (is.atomic(y) | is.factor(y)) {
       y <- data.frame(y)
       names(y) <- y_name
     }
@@ -328,7 +328,7 @@ make_formula <- function(x, y, short = TRUE) {
 
 
 will_make_matrix <- function(y) {
-  if (is.matrix(y) | is.vector(y) | is.factor(y))
+  if (is.matrix(y) | is.atomic(y) | is.factor(y))
     return(FALSE)
   cls <- unique(unlist(lapply(y, class)))
   if (length(cls) > 1)
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -234,7 +234,7 @@
   if (is.matrix(y)) {
     y <- as.data.frame(y)
   } else {
-    if (is.atomic(y) | is.factor(y)) {
+    if (is.atomic(y)) {
       y <- data.frame(y)
       names(y) <- y_name
     }
@@ -328,7 +328,7 @@ make_formula <- function(x, y, short = TRUE) {
 
 
 will_make_matrix <- function(y) {
-  if (is.matrix(y) | is.atomic(y) | is.factor(y))
+  if (is.matrix(y) | is.atomic(y))
     return(FALSE)
   cls <- unique(unlist(lapply(y, class)))
   if (length(cls) > 1)
--- a/R/fit.R
+++ b/R/fit.R
@@ -411,10 +411,8 @@ check_xy_interface <- function(x, y, cl, model) {
     inher(x, c("data.frame", "matrix"), cl)
   }
 
-  # `y` can be a vector (which is not a class), or a factor or
-  # Surv object (which are not vectors)
   if (!is.null(y) && !is.atomic(y))
-    inher(y, c("data.frame", "matrix", "factor", "Surv"), cl)
+    inher(y, c("data.frame", "matrix"), cl)
 
   # rule out spark data sets that don't use the formula interface
   if (inherits(x, "tbl_spark") | inherits(y, "tbl_spark"))
--- a/R/fit.R
+++ b/R/fit.R
@@ -267,7 +267,7 @@ fit_xy.model_spec <-
     }
     y_var <- colnames(y)
 
-    if (object$engine != "spark" & NCOL(y) == 1 & !(is.vector(y) | is.factor(y))) {
+    if (object$engine != "spark" & NCOL(y) == 1 & !(is.atomic(y) | is.factor(y))) {
       if (is.matrix(y)) {
         y <- y[, 1]
       } else {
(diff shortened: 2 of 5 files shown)


R/convert_data.R
```diff
@@ -234,7 +234,7 @@
   if (is.matrix(y)) {
     y <- as.data.frame(y)
   } else {
-    if (is.vector(y) | is.factor(y)) {
+    if (is.atomic(y) | is.factor(y)) {
```
hfrick, line 237:
```suggestion
    if (is.atomic(y)) {
```

because
``` r
y_factor <- factor(1:5)

# old
is.vector(y_factor)
#> [1] FALSE

# new
is.atomic(y_factor)
#> [1] TRUE
```

<sup>Created on 2024-02-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

R/convert_data.R
```diff
@@ -328,7 +328,7 @@ make_formula <- function(x, y, short = TRUE) {
 
 
 will_make_matrix <- function(y) {
-  if (is.matrix(y) | is.vector(y) | is.factor(y))
+  if (is.matrix(y) | is.atomic(y) | is.factor(y))
```
hfrick, line 331:
```suggestion
  if (is.matrix(y) | is.atomic(y))
```
same reason as earlier

R/fit.R
```diff
@@ -267,7 +267,7 @@ fit_xy.model_spec <-
     }
     y_var <- colnames(y)
 
-    if (object$engine != "spark" & NCOL(y) == 1 & !(is.vector(y) | is.factor(y))) {
+    if (object$engine != "spark" & NCOL(y) == 1 & !(is.atomic(y) | is.factor(y))) {
```
hfrick, line 270:
```suggestion
    if (object$engine != "spark" & NCOL(y) == 1 & !(is.atomic(y))) {
```

R/fit.R
```diff
@@ -413,7 +413,7 @@ check_xy_interface <- function(x, y, cl, model) {
 
   # `y` can be a vector (which is not a class), or a factor or
   # Surv object (which are not vectors)
-  if (!is.null(y) && !is.vector(y))
+  if (!is.null(y) && !is.atomic(y))
     inher(y, c("data.frame", "matrix", "factor", "Surv"), cl)
```
hfrick, line 418: `inher()` now only needs to check for data frames and matrices, because in addition to factors also Surv objects are "atomic". Let's update or remove the comment as well.

``` r
y_surv <- survival::Surv(1:4, c(0, 1, 0, 1))

# old
is.vector(y_surv)
#> [1] FALSE

# new
is.atomic(y_surv)
#> [1] TRUE
```

<sup>Created on 2024-02-12 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

hfrick (approved):
On board! Will be good to not trip over labels. I've suggested some small changes because the switch from `is.vector()` to `is.atomic()` also allows us to remove some of the special-casing for factors (and Surv objects).

R/fit.R
simonpcouch: Thanks. :) [9ceac0d](https://github.com/tidymodels/parsnip/pull/1062/commits/9ceac0d7c8d9ac30cb86952790ef11e3dcb9e2ec)

PR #1066, opened 2024-02-14
changes for #1039

### LINKED ISSUE #1039: Remove forced injection of `eval_time` in `predict_survival()`
We add `eval_time` to _all_ prediction calls in `predict_survival.model_fit()`  and  `predict_hazard.model_fit()` which causes problems if the the engine's predict method does not have an `eval_time` argument and does not swallow it via the dots, e.g. by enforcing empty dots. 

The `"aorsf"` engine for `rand_forest()` in censored enforces empty dots which is what lead to https://github.com/tidymodels/censored/issues/236. The dev version of censored does not rely on this pattern anymore so once that is on CRAN we should remove those lines:

https://github.com/tidymodels/parsnip/blob/8f13c1c41ce603261f25af64694c253f618fa999/R/predict_survival.R#L37-L38

https://github.com/tidymodels/parsnip/blob/8f13c1c41ce603261f25af64694c253f618fa999/R/predict_hazard.R#L35-L36

topepo:
Closes #1039 

Tested on extratests via `devtools::test(filter = "(survival)|(censor)")`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # parsnip (development version)
 
+* We no longer add `eval_time` arguments to the prediction specification (#1039).
+
 * parsnip now lets the engines for [mlp()] check for acceptable values of the activation function (#1019)
 
 * Tightened logic for outcome checking. This resolves issues—some errors and some silent failures—when atomic outcome variables have an attribute (#1060, #1061).
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,6 @@
 # parsnip (development version)
 
-* We no longer add `eval_time` arguments to the prediction specification (#1039).
+* We no longer add `eval_time` arguments to the prediction specification for the engine (#1039).
 
 * parsnip now lets the engines for [mlp()] check for acceptable values of the activation function (#1019)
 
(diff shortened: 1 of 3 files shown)


R/predict_hazard.R
```diff
@@ -33,7 +33,6 @@ predict_hazard.model_fit <- function(object,
     new_data <- object$spec$method$pred$hazard$pre(new_data, object)
 
   # Pass some extra arguments to be used in post-processor
```
hfrick, line 35: This comment was for the line we remove here so let's get rid of this as well! 🧹

R/predict_survival.R
```diff
@@ -35,7 +35,6 @@ predict_survival.model_fit <- function(object,
     new_data <- object$spec$method$pred$survival$pre(new_data, object)
 
   # Pass some extra arguments to be used in post-processor
```
hfrick, line 37: same here

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # parsnip (development version)
 
+* We no longer add `eval_time` arguments to the prediction specification (#1039).
```
hfrick, line 3:
```suggestion
* We no longer add `eval_time` arguments to the prediction specification for the engine (#1039).
```

hfrick (approved):
Bye bye special case! 🎉

R/predict_hazard.R
hfrick, line 35:
```suggestion
```

R/predict_survival.R
hfrick, line 37:
```suggestion
```

PR #1122, opened 2024-05-23
Doc Fix: `mlp_brulee()`

kscott-1:
The default parameter for `penalty` in the docs is incorrect. This is a comically small PR, but it did trip me up this week thinking the default was 0 when it wasn't.

Here is some quick reprex proof via `parsnip::` and via `brulee::` itself (see 'weight decay' value):

``` r
data <- modeldata::two_class_dat
# - recipe
rec <-
    data |>
        recipes::recipe(Class ~ .)
# - default param brulee_mlp via parsnip
mod <-
    parsnip::mlp() |>
        parsnip::set_engine("brulee") |>
        parsnip::set_mode("classification")
# - workflow
wf <-
    workflows::workflow() |>
        workflows::add_recipe(rec) |>
        workflows::add_model(mod)
# - fit
wf |>
    parsnip::fit(data = data)
#> ══ Workflow [trained] ══════════════════════════════════════════════════════════
#> Preprocessor: Recipe
#> Model: mlp()
#> 
#> ── Preprocessor ────────────────────────────────────────────────────────────────
#> 0 Recipe Steps
#> 
#> ── Model ───────────────────────────────────────────────────────────────────────
#> Multilayer perceptron
#> 
#> relu activation
#> 3 hidden units,  17 model parameters
#> 791 samples, 2 features, 2 classes 
#> class weights Class1=1, Class2=1 
#> weight decay: 0.001 
#> dropout proportion: 0 
#> batch size: 712 
#> learn rate: 0.01 
#> validation loss after 4 epochs: 0.5
```

``` r

brulee::brulee_mlp(
    x = rec,
    data = data
)
#> Multilayer perceptron
#> 
#> relu activation
#> 3 hidden units,  17 model parameters
#> 791 samples, 2 features, 2 classes 
#> class weights Class1=1, Class2=1 
#> weight decay: 0.001 
#> dropout proportion: 0 
#> batch size: 712 
#> learn rate: 0.01 
#> validation loss after 30 epochs: 0.379
```

PR #1123, opened 2024-05-23
run `devtools::document()` for .Rd file

kscott-1:
Follow up to #1122 , PR didn't include the .Rd file needed for building the site. This should fix the issue.


simonpcouch:
You got it! FYI those changes will appear only on parsnip.tidymodels.org/dev rather than parsnip.tidymodels.org until a new release of the package makes it to CRAN.

kscott-1:
Sweet!

PR #1150, opened 2024-08-15
Document default ranges for tunable parameters

### LINKED ISSUE #954: Document default ranges for tuning parameters
when you run the neighbors() in R, it returns that the default range is [1,10]. Additionally, when you look up the documentation for neighbors(), it says the default is [1,10].

However, when you set neighbors = tune() in a model and try to use the default range, R selects a range of [1, 15] instead of [1, 10]. This can be seen when you run extract_set_parameter_dials() and look at the range.

it would be super helpful to have this discrepancy fixed, either by changing the documentation to be [1, 15] or changing the default that results when you set neighbors = tune()
Screenshot 2023-03-31 at 1 47 42 PM
Screenshot 2023-03-31 at 1 48 03 PM

rkb965:
Documents {parsnip}'s default ranges for tunable parameter via `set_model_arg()`. Also removes documentation for mars model earth engine that described non-implemented functionality for `num_terms`.

Fixes #954 

Note that this does not include tunable parameters where this is set by another means.

### DIFF
--- a/man/rmd/bart_dbarts.Rmd
+++ b/man/rmd/bart_dbarts.Rmd
@@ -25,6 +25,8 @@ This model has `r nrow(param)` tuning parameters:
 param$item
 ```
 
+Parsnip changes the default range for `trees` to `c(50, 500)`.
+
 ## Important engine-specific options
 
 Some relevant arguments that can be passed to `set_engine()`: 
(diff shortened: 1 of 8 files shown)


rkb965:
Oh and also note that I was unable to successfully install all engine dependencies, so only .Rmds are updated and `devtools::document()` was not run (sorry!)

topepo:
Thanks!

PR #1152, opened 2024-08-15
mass replace of rlang::abort with cli::cli_abort

RodDalBen:
mass replace of rlang::abort with cli::cli_abort, no user input or special glue::glue brackets were used, so a mass replace was made #1142

### DIFF
--- a/R/nearest_neighbor_data.R
+++ b/R/nearest_neighbor_data.R
@@ -93,7 +93,7 @@ set_pred(
     # model mode
     pre = function(x, object) {
       if (object$fit$response != "continuous") {
-        rlang::abort(
+        cli::cli_abort(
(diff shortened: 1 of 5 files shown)


simonpcouch (commented):
While transitioning directly from `rlang::abort()` to `cli::cli_abort()` with a Find and Replace doesn't introduce new issues, we tend to also revisit the message formatting using [inline markup](https://cli.r-lib.org/reference/inline-markup.html) when we do so. Would you be interested in revising this PR to remove calls to `glue::glue()` and format these messages with inline markup?

simonpcouch:
I'm going to go ahead and close. Thank you for the PR! If you'd like to pick this back up, please feel free to holler here and we'll reopen.

PR #1157, opened 2024-08-18
Engine and extension installs

### LINKED ISSUE #1156: make it easier to run engine documentation
There are non-standard package requirements to successfully run `knit_engine_docs()`. We know what they extras are so that function should check for them.

topepo:
Closes #1156 

Updates the existing (internal) `install_engine_packages()` and calls it when `knir_engine_docs()` is executed.

### DIFF
--- a/R/install_packages.R
+++ b/R/install_packages.R
@@ -3,9 +3,9 @@ install_engine_packages <- function(extension = TRUE, extras = TRUE,
                                     ignore_pkgs = c("stats", "liquidSVM",
                                                     "parsnip")) {
   bio_pkgs <- c()
-  extensions_packages <- extensions()
 
   if (extension) {
+    extensions_packages <- extensions()
     rlang::check_installed(extensions_packages)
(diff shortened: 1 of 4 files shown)


R/install_packages.R
```diff
@@ -1,16 +1,12 @@
+# Installs packed needed to run `knit_engine_docs()`.
```
simonpcouch, line 1:
```suggestion
# Installs packages needed to run `knit_engine_docs()`.
```

R/install_packages.R
```diff
@@ -1,16 +1,12 @@
...
 install_engine_packages <- function(extension = TRUE, extras = TRUE,
                                     ignore_pkgs = c("stats", "liquidSVM",
                                                     "parsnip")) {
   bio_pkgs <- c()
+  extensions_packages <- extensions()
```
simonpcouch, line 6: Why is this now outside of the `if (extension)` conditional? I don't believe it's newly used outside of it.

simonpcouch (commented):
🐝

PR #1164, opened 2024-08-29
refine parsnip overhead test

simonpcouch:
Maybe every three or four times we run the parsnip overhead unit tests on the macOS-release and ubuntu-oldrel-4 runners, we see a failure. We should figure out how to make that happen much, much less often.

Starting out by making the error message more informative to see whether we're "close" to the limit.

### DIFF
--- a/tests/testthat/test-fit_interfaces.R
+++ b/tests/testthat/test-fit_interfaces.R
@@ -169,7 +169,7 @@ test_that("overhead of parsnip interface is minimal (#1071)", {
   )
 
   expect_true(
-    bm$median[2] < 3,
+    bm$median[2] < 3.5,
     label = paste0("parsnip overhead factor (formula interface): ", bm$median[2])
   )
   expect_true(
--- a/tests/testthat/test-fit_interfaces.R
+++ b/tests/testthat/test-fit_interfaces.R
@@ -168,6 +168,12 @@ test_that("overhead of parsnip interface is minimal (#1071)", {
     check = FALSE
   )
 
-  expect_true(bm$median[2] < 3, label = "parsnip overhead is minimal (formula interface)")
-  expect_true(bm$median[3] < 3.5, , label = "parsnip overhead is minimal (xy interface)")
+  expect_true(
+    bm$median[2] < 3,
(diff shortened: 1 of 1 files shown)


simonpcouch:
The hard-coded limit for `fit_xy()` is a factor of 3.5x in the minimal example. We see in one failure:

> parsnip overhead factor (xy interface): 3.52

Sooo.. it _is_ a close call, which made me think we may have indeed that fit time slightly in recent changes. Profiling the dev version:

Screenshot 2024-08-29 at 12 36 34 PM


This is actually the case—calls to `proc.time()` from https://github.com/tidymodels/parsnip/commit/1e7dad2753f8bd0629f5c49bfaca854b4f4167b9 (and before that, https://github.com/tidymodels/parsnip/commit/40524aab2cd6b4605fa4b7f65c91c3b363fc16ea) are what's causing that inflation in fit time.

simonpcouch:
Here are the distributions of those factor slowdowns I see locally:

``` r
library(parsnip)
library(bench)

form <- numeric()
xy <- numeric()

for (i in 1:30) {
  bm <- bench::mark(
    time_engine = lm(mpg ~ ., mtcars),
    time_parsnip_form = fit(linear_reg(), mpg ~ ., mtcars),
    time_parsnip_xy = fit_xy(linear_reg(), mtcars[2:11],  mtcars[1]),
    relative = TRUE,
    check = FALSE
  )
  
  form <- c(form, bm$median[2])
  xy <- c(xy, bm$median[3])
}

hist(form)
```



The limit for the above image is currently 3.

``` r
hist(xy)
```



The limit for the above image is currently 3.5.

<sup>Created on 2024-08-29 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

On average, these `bench::mark()` runs are running 300 to 2000 iterations per expression.

At least on this PR, I've seen 3/6 runs failing with XY overhead factors of 3.52, 3.56, and 3.501. I've yet to see any failures with the formula interface.

Alternatives from bench and ps are a good bit slower, at least on macOS. I do think the timing functionality is worth having, _and_ it's made it to CRAN already. I'm leaning towards increasing that overhead limit for the XY interface—3.75 should give plenty of room.

Issue #1125, opened 2024-06-13
Roadmap: parsnip support for sparse tibbles


EmilHvitfeldt:
What we need:

- [x] `fit()` to take sparse tibbles as data https://github.com/tidymodels/parsnip/pull/1165
- [x] `fit()` to take {Matrix} sparse matrix as data https://github.com/tidymodels/parsnip/pull/1121
  - turn them into sparse tibbles early
- [x] `predict()` to take sparse tibbles as data https://github.com/tidymodels/parsnip/pull/1167
- [x] `predict()` to take {Matrix} sparse matrix as data https://github.com/tidymodels/parsnip/pull/1167
  - turn them into sparse tibbles early
- [x] look into if we document which engines are sparse friendly https://github.com/tidymodels/parsnip/pull/1171
- [x] special cases for some model types https://github.com/tidymodels/parsnip/pull/1173
  - {xgboost} with `xgboost::xgb.DMatrix()` and `xgb_train()` (only special case engine that allows sparse data)
- [x] Better error if sparse matrix is used with `fit()` https://github.com/tidymodels/parsnip/pull/1174
- [x] Make sure all error messages are piped through correctly https://github.com/tidymodels/parsnip/pull/1174

I think we could use a `option()` of some kind to unit test that the data passed is passed around in a way that keeps the sparsity.

Adding all of this will give us
- standalone usage of sparse matrices in {parsnip} 
- everything it needs to be able to work with the rest of {tidymodels} in regards to sparse tibbles

PR #1171, opened 2024-09-04
document sparse data usage in parsnip

EmilHvitfeldt:
Ref: https://github.com/tidymodels/parsnip/issues/1125

Tried to follow how we documented case weights usage across models/engines.

`?sparse_data` doesn't really say a lot, but it also doesn't really need to because parsnip should just work if the engine allows it


R/sparsevctrs.R
```diff
@@ -1,3 +1,21 @@
...
+#' calling `get_encoding("name of model")` and looking at the `allow_sparse_x`
+#' column.
+#' 
+#' Using sparse data for model fitting and prediction shouldn't require any 
+#' additional configurations. Just pass in a sparse matrix such as dgCMatrix 
+#' from the `Matrix` package or a sparse tibble from the `sparsevctrs` package 
+#' to the data argument of the respective [fit()], [fit_xy()], and [predict()].
+#' 
+#' Models that don't support sparse data will try to convert to non-sparse data 
+#' with warnings. An informative error will be thrown if conversion isn't
```
topepo, line 13: > If conversion isn’t possible, an informative error will be thrown.

Is a little better

R/sparsevctrs.R
```diff
@@ -1,3 +1,21 @@
+#' Using sparse data with parsnip
+#' 
+#' You can figure out whether a given model engine supports sparse data by 
+#' calling `get_encoding("name of model")` and looking at the `allow_sparse_x`
+#' column.
+#' 
+#' Using sparse data for model fitting and prediction shouldn't require any 
+#' additional configurations. Just pass in a sparse matrix such as dgCMatrix 
+#' from the `Matrix` package or a sparse tibble from the `sparsevctrs` package
```
simonpcouch, line 9:
```suggestion
#' from the `Matrix` package or a sparse tibble from the sparsevctrs package 
```

R/sparsevctrs.R
```diff
@@ -1,3 +1,21 @@
+#' Using sparse data with parsnip
+#' 
+#' You can figure out whether a given model engine supports sparse data by 
+#' calling `get_encoding("name of model")` and looking at the `allow_sparse_x`
+#' column.
+#' 
+#' Using sparse data for model fitting and prediction shouldn't require any 
+#' additional configurations. Just pass in a sparse matrix such as dgCMatrix 
+#' from the `Matrix` package or a sparse tibble from the `sparsevctrs` package 
+#' to the data argument of the respective [fit()], [fit_xy()], and [predict()].
```
simonpcouch, line 10:
```suggestion
#' to the data argument of [fit()], [fit_xy()], and [predict()].
```

Since we've now happy-pathed `fit()` with sparse matrices. Also, "respective" didn't have a 1-to-1 mapping here.

simonpcouch (approved):
Wonderful, this is a nice addition.

Note to self that I ought to make sure passing sparse data works in bonsai with LightGBM. Let me know if you have any notes to keep in mind there.

R/sparsevctrs.R
EmilHvitfeldt: done as https://github.com/tidymodels/parsnip/pull/1171/commits/9a74d3a9a24b22b8e5b6c16b30cd4dc07b1b1a46

PR #1182, opened 2024-09-09
use type checkers in `fit.R`

### LINKED ISSUE #1181: transition to rlang type checkers in `fit.R`

simonpcouch:
Closes #1181.🐝

### DIFF
--- a/R/fit.R
+++ b/R/fit.R
@@ -55,11 +55,11 @@
 #' a "reverse Kaplan-Meier" curve that models the probability of censoring. This
 #' may be used later to compute inverse probability censoring weights for
 #' performance measures.
-#' 
+#'
 #' Sparse data is supported, with the use of the `x` argument in `fit_xy()`. See
-#' `allow_sparse_x` column of [parsnip::get_encoding()] for sparse input 
+#' `allow_sparse_x` column of [parsnip::get_encoding()] for sparse input
 #' compatibility.
-#' 
+#'
 #' @examplesIf !parsnip:::is_cran_check()
 #' # Although `glm()` only has a formula interface, different
 #' # methods for specifying the model can be used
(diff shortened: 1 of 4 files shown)


R/misc.R
```diff
@@ -533,6 +533,21 @@ check_case_weights <- function(x, spec, call = rlang::caller_env()) {
   invisible(NULL)
 }
 
+# ------------------------------------------------------------------------------
+
+check_inherits <- function(x, cls, arg = caller_arg(x), call = caller_env()) {
```
simonpcouch, line 538: Decided to write out `inher()` in favor of this helper. I think we could probably get a good bit of mileage out of this based on the number of `if (inherits(` in this repo.

R/fit.R
```diff
@@ -373,59 +370,47 @@ eval_mod <- function(e, capture = FALSE, catch = FALSE, envir = NULL, ...) {
...
-  invisible(x)
-}
-
-# ------------------------------------------------------------------------------
-
-check_interface <- function(formula, data, cl, model) {
-  inher(formula, "formula", cl)
-  inher(data, c("data.frame", "dgCMatrix", "tbl_spark"), cl)
+check_interface <- function(formula, data, cl, model, call = caller_env()) {
+  check_inherits(formula, "formula", call = call)
```
EmilHvitfeldt, line 374: couldn't this be `check_formula()`?

R/misc.R
EmilHvitfeldt: 100%, and the name is much better too!

R/fit.R
```diff
@@ -444,7 +429,7 @@ check_xy_interface <- function(x, y, cl, model) {
 
   check_outcome(y, model)
 
-  cli::cli_abort("Error when checking the interface")
+  cli::cli_abort("Error when checking the interface", call = call)
```
EmilHvitfeldt, line 432:
```suggestion
  cli::cli_abort("Error when checking the interface.", call = call)
```

EmilHvitfeldt (approved):
so happy to see this being done!

R/fit.R
simonpcouch: omg totally

Issue #1096, opened 2024-04-02
expect_error() is being used too much


EmilHvitfeldt:
as an example the following test https://github.com/tidymodels/parsnip/blob/main/tests/testthat/test_decision_tree.R#L15-L18

```r
  expect_snapshot_error({
    bt <- decision_tree(cost_complexity = -1) %>% set_engine("rpart")
    fit(bt, class ~ ., hpc)
  })
```

is expected to test that `cost_complexity` isn’t good, but it instead tests that the mode isn’t set

```r
#> Error in `fit()`:
#> ! Please set the mode in the model specification.
```

PR #1188, opened 2024-09-09
`expect_error(regexp = NA)` -> `expect_no_condition()`

simonpcouch:
Related to #1096 but does not close—these are just the cases where `regexp = NA`, i.e. we're testing that there's no error. For all but one case, we can actually test the more stricter version, that there is no condition.

### DIFF
--- a/tests/testthat/test-logistic_reg.R
+++ b/tests/testthat/test-logistic_reg.R
@@ -37,15 +37,14 @@ test_that('glm execution', {
 
 
   # passes interactively but not on R CMD check
-  # expect_error(
+  # expect_no_condition(
   #   res <- fit(
   #     lc_basic,
   #     lc_form,
   #     data = lending_club,
   #     control = ctrl,
   #     engine = "glm"
-  #   ),
-  #   regexp = NA
+  #   )
   # )
   expect_no_condition(
     res <- fit_xy(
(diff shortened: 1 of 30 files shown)


tests/testthat/test-re_registration.R
```diff
@@ -22,17 +21,16 @@ test_that('re-registration of engine', {
 
 test_that('re-registration of package dependencies', {
   old_val <- get_from_env("bart_pkgs")
-  expect_error(
-    set_dependency("bart", "dbarts", "dbarts"),
-    regexp = NA
-  )
+  expect_no_error(expect_no_warning(
+    set_dependency("bart", "dbarts", "dbarts")
```
simonpcouch, line 25: There's a `<dplyr_regroup>` condition with no message here; it's not visible at all as a user, so just opted to ignore it.

tests/testthat/test-logistic_reg.R
```diff
@@ -47,14 +47,13 @@ test_that('glm execution', {
   #   ),
   #   regexp = NA
   # )
```
EmilHvitfeldt, line 49: i would translate this one too. Even though it is commented out

EmilHvitfeldt (approved):
You missed a couple in

- `test-descriptors.R`: 202, 208
- `test-extract.R`: 93
- `test-logistic_reg.R`: 180
- `test-misc.R`: 76, 80, 164
- `test-svm_linear.R`: 51

PR #1190, opened 2024-09-09
transition to snapshot testing

### LINKED ISSUE #1096: expect_error() is being used too much

simonpcouch:
Closes #1096.

This PR surfaces quite a few places where bugs had triggered unintended changes to messages (bad glue snytax, no object found, etc.) over time and others where we've yet to set an informative `call`. I'm opting not to address them here in favor of spinning up separate issues; if we feel that those changes would live more happily in this PR I'm happy to put some time in there.

### DIFF
--- a/tests/testthat/_snaps/adds.md
+++ b/tests/testthat/_snaps/adds.md
@@ -1,0 +1,8 @@
+# adding row indicies
+
+    Code
+      as.matrix(mtcars) %>% add_rowindex()
+    Condition
+      Error in `add_rowindex()`:
+      ! `x` should be a data frame.
+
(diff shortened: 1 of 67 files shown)


EmilHvitfeldt (approved):
looking fire. you missed a `expect_message()` so i added that

PR #1191, opened 2024-09-09
Quantile predictions output constructor

dajmcdon:
@topepo 

1. This is just the vctrs constructor. 
2. It only allows a matrix input (with columns corresponding to `quantile_levels`).
3. It works with both types of `rq()` predictions (1 or more `tau`/`quantile_levels`).

I added a test for the constructor and adjusted your tests for the `quantreg` engine.

Don't hesitate to ask for more changes! Or we can revisit if you'd really prefer the tibble list-col version.


R/aaa_quantiles.R
```diff
@@ -21,23 +21,100 @@ check_quantile_level <- function(x, object, call) {
...
+# -------------------------------------------------------------------------
+# A column vector of quantiles with an attribute
+
+#' @export
+vec_ptype_abbr.vctrs_quantiles <- function(x, ...) "qntls"
+
+#' @export
+vec_ptype_full.vctrs_quantiles <- function(x, ...) "quantiles"
+
+#' @importFrom rlang is_double !!!
```
simonpcouch, line 34:
```suggestion
```

redundant with: https://github.com/tidymodels/parsnip/blob/cee2bb8fc4766b1067faec266cd3036d7fca5f3c/R/parsnip-package.R#L11.

R/aaa_quantiles.R
```diff
@@ -21,23 +21,100 @@ check_quantile_level <- function(x, object, call) {
...
   }
-  n <- nrow(x)
+  values <- lapply(vctrs::vec_chop(values), drop)
+  new_vec_quantiles(values, quantile_levels)
+}
 
+check_vec_quantiles_inputs <- function(values, levels) {
+  if (!is.matrix(values)) {
+    cls <- class(values)[1]
+    cli::cli_abort("{.arg values} must be a {.cls matrix} not a {.cls {cls}}.")
```
simonpcouch, line 76:
```suggestion
    cli::cli_abort("{.arg values} must be a {.cls matrix}, not {.obj_type_friendly {values}}.")
```

tests/testthat/test-vec_quantiles.R
simonpcouch, line 1: Let's rename this file to `test-aaa_quantiles.R` for `usethis::use_test()` happiness :)

tests/testthat/test-vec_quantiles.R
```diff
@@ -0,0 +1,25 @@
+test_that("vec_quantiles error types", {
+  expect_error(vec_quantiles(1:10, 1:4 / 5), "matrix")
+  expect_error(
+    vec_quantiles(matrix(1:20, 5), -1:4 / 5),
+    "`quantile_levels` must be a number between 0 and 1"
+  )
```
simonpcouch, line 6:
```suggestion
  expect_snapshot(
    error = TRUE,
    vec_quantiles(matrix(1:20, 5), -1:4 / 5)
  )
```

Will just do one of these, but we now prefer snapshot tests for errors over `expect_error()`!

R/aaa_quantiles.R
```diff
@@ -21,23 +21,100 @@ check_quantile_level <- function(x, object, call) {
...
+      "The number of columns in {.arg values} must be equal to the length of
+        {.arg quantile_levels}."
+    )
   }
-  n <- nrow(x)
+  values <- lapply(vctrs::vec_chop(values), drop)
+  new_vec_quantiles(values, quantile_levels)
+}
 
+check_vec_quantiles_inputs <- function(values, levels) {
```
simonpcouch, line 73:
```suggestion
check_vec_quantiles_inputs <- function(values, levels, call = caller_env()) {
```

Then, in each of the calls to `cli::cli_abort()` in this function, pass `call = call` so that users see "Error in `user_facing_fn()`" rather than "Error in `unexported_helper_fn()`"

R/aaa_quantiles.R
```diff
@@ -21,23 +21,100 @@ check_quantile_level <- function(x, object, call) {
...
+vctrs::obj_print_footer
+
+#' @export
+obj_print_footer.vctrs_quantiles <- function(x, ...) {
+  lvls <- attr(x, "quantile_levels")
+  cat("# Quantile levels: ", format(lvls, digits = 3), "\n", sep = " ")
+}
+
+restructure_rq_pred <- function(x, object) {
+  if (!is.matrix(x)) x <- as.matrix(x)
```
simonpcouch, line 113:
```suggestion
  if (!is.matrix(x)) {
    x <- as.matrix(x)
  }
```

R/aaa_quantiles.R
simonpcouch: Note that this doesn't apply to `vec_quantiles()`, as the default `caller_env()` refers to the correct function in that case!

simonpcouch (commented):
Stoked that this is coming together. Will defer to others on feedback related to the design/interface, but some engineering and code-style edits!

dajmcdon:
@simonpcouch Thanks!

R/aaa_quantiles.R
```diff
@@ -21,23 +21,115 @@ check_quantile_level <- function(x, object, call) {
...
+    cli::cli_abort(
+      "{.arg quantile_levels} must be sorted in increasing order.",
+      call = call
+    )
+  }
+  invisible(NULL)
+}
+
+#' @export
+format.vctrs_quantiles <- function(x, ...) {
```
EmilHvitfeldt, line 101: since you are defining a `format()` method, you should snapshot test it with `expect_snapshot()` to make sure it prints what you exepct

R/aaa_quantiles.R
```diff
@@ -21,23 +21,115 @@ check_quantile_level <- function(x, object, call) {
...
+#' @export
+vec_ptype_abbr.vctrs_quantiles <- function(x, ...) "qntls"
+
+#' @export
+vec_ptype_full.vctrs_quantiles <- function(x, ...) "quantiles"
+
+new_vec_quantiles <- function(values = list(), quantile_levels = double()) {
+  quantile_levels <- vctrs::vec_cast(quantile_levels, double())
+  vctrs::new_vctr(
+    values, quantile_levels = quantile_levels, class = "vctrs_quantiles"
```
EmilHvitfeldt, line 46: This is more global comment. When writing {vctrs} classes, it is customary to prefix the name of the class with the package where it lives. So it would be `parsnip_quantiles` instead of `vctrs_quantiles`.

EmilHvitfeldt (commented):
looking good so far. left a couple of suggestions

R/aaa_quantiles.R
```diff
@@ -21,23 +21,115 @@ check_quantile_level <- function(x, object, call) {
...
+
+#' @export
+format.parsnip_quantiles <- function(x, ...) {
+  quantile_levels <- attr(x, "quantile_levels")
+  if (length(quantile_levels) == 1L) {
+    x <- unlist(x)
+    out <- round(x, 3L)
+    out[is.na(x)] <- NA
+  } else {
+    rng <- sapply(x, range)
```
topepo, line 108:
```suggestion
    rng <- sapply(x, range, na.rm = TRUE)
```

because

``` r
library(parsnip)
pred <- matrix(rnorm(20), 5)
pred[1,1] <- NA
v <- parsnip_quantiles(pred, c(.2, .4, .6, .8))
v
#> <quantiles[5]>
#> [1] <NA>            [-0.254, 1.053] [-0.102, 1.405] [-1.758, 0.441]
#> [5] [-1.612, 1.074]
#> # Quantile levels:  0.2 0.4 0.6 0.8
unclass(v)
#> [[1]]
#> [1]         NA  0.1452904 -0.3481203  0.4976215
#> 
#> [[2]]
#> [1]  0.4449808  1.0531736  0.5797585 -0.2539239
#> 
#> [[3]]
#> [1]  1.4052520  0.6012013 -0.1020287  0.5743619
#> 
#> [[4]]
#> [1] -1.7581333  0.4408413 -0.4214030  0.2257039
#> 
#> [[5]]
#> [1] -0.1161026 -1.6124717  1.0739467 -1.1894932
#> 
#> attr(,"quantile_levels")
#> [1] 0.2 0.4 0.6 0.8
```

<sup>Created on 2024-09-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

topepo:
Thanks for making this

> Or we can revisit if you'd really prefer the tibble list-col version.

Nope, I'm good with using this class!

topepo:
One other small thing... let's remove row and column names from `values` in `parsnip_quantiles()`. They carry over into the vctrs object (although you can't see them unless you do something like `unclass()`)

R/aaa_quantiles.R
```diff
@@ -21,23 +21,115 @@ check_quantile_level <- function(x, object, call) {
...
+  }
+  out
+}
+
+#' @importFrom vctrs obj_print_footer
+#' @export
+vctrs::obj_print_footer
+
+#' @export
+obj_print_footer.parsnip_quantiles <- function(x, ...) {
```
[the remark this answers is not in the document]
topepo: For the testing example: 

```r
unclass(ten_quant_pred$.pred_quantile) %>% 
  map_chr(~ cli::format_inline("[{length(.x)} qntl{?s}]"))
```

or similar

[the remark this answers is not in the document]
dajmcdon: How about something like this:

``` r
library(parsnip)
set.seed(12345)
b <- quantile_pred(matrix(rnorm(20), ncol = 4), 1:4 / 5) # no explicit 0.5
b
#> <quantiles[5]>
#> [1] [-0.967] [1.224]  [0.047]  [0.118]  [-0.835]
#> # Quantile levels:  0.2 0.4 0.6 0.8
tibble::tibble(q = b)
#> # A tibble: 5 × 1
#>           q
#>   <qtls(4)>
#> 1  [-0.967]
#> 2   [1.224]
#> 3   [0.047]
#> 4   [0.118]
#> 5  [-0.835]
```

<sup>Created on 2024-09-10 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

Now the "short" version in the header of the tibble tells me how many there are, and the median gets shown (rather than the range), with `[` to indicate that it's something other than the raw numbers.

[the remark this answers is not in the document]
topepo: That's good.

R/aaa_quantiles.R
```diff
@@ -9,35 +9,214 @@ check_quantile_level <- function(x, object, call) {
...
+  map_dbl(x, ~ stats::approx(lvls, .x, xout = 0.5)$y)
+}
 
+restructure_rq_pred <- function(x, object) {
+  if (!is.matrix(x)) {
+    x <- as.matrix(x)
+  }
+  rownames(x) <- NULL
+  n_pred_quantiles <- ncol(x)
+  # TODO check p = length(quantile_level)
```
hfrick, line 189: is this a TODO that should be address in this PR?

R/aaa_quantiles.R
```diff
@@ -9,35 +9,214 @@ check_quantile_level <- function(x, object, call) {
...
-  res <-
-    tibble::tibble(
-    .pred_quantile = as.vector(x),
-    .quantile_level = rep(quantile_level, each = n),
-    .row = rep(1:n, num_quantiles))
-  res <- vctrs::vec_split(x = res[,1:2], by = res[, ".row"])
-  res <- vctrs::vec_cbind(res$key, tibble::new_tibble(list(.pred_quantile = res$val)))
-  res$.row <- NULL
-  res
+  tibble::tibble(.pred_quantile = quantile_pred(x, quantile_level))
```
hfrick, line 191: If we are doing input checking already (related to the TODO above), we could switch to `tibble::new_tibble()` which is faster because it does fewer checks. This is particularly helpful in places that called often.

tests/testthat/test-linear_reg_quantreg.R
```diff
@@ -24,32 +17,43 @@ test_that('linear quantile regression via quantreg - single quantile', {
   expect_true(nrow(one_quant_pred) == nrow(sac_test))
   expect_named(one_quant_pred, ".pred_quantile")
   expect_true(is.list(one_quant_pred[[1]]))
-  expect_s3_class(one_quant_pred$.pred_quantile[[1]], c("tbl_df", "tbl", "data.frame"))
-  expect_named(one_quant_pred$.pred_quantile[[1]], c(".pred_quantile", ".quantile_level"))
-  expect_true(nrow(one_quant_pred$.pred_quantile[[1]]) == 1L)
+  expect_s3_class(
```
hfrick, line 20: testthat has [`expect_vector()`](https://testthat.r-lib.org/reference/expect_vector.html) specifically for testing vctrs vectors, that seems relevant here.

R/aaa_quantiles.R
```diff
@@ -9,35 +9,214 @@ check_quantile_level <- function(x, object, call) {
       {.arg quantile_level} must be specified for quantile regression models.")
     }
   }
+  if ( any(is.na(x)) ) {
```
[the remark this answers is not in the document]
topepo: I prefer the existing format; easier to read

tests/testthat/test-linear_reg_quantreg.R
topepo: The output is a vector but the prototype for `expect_vector()` has matrix and vector inputs whose sizes must match and cannot be zero. It's not very straightforward to do that; It's a quare peg in a round hole. I think that the complexity there is not really worth it.

R/aaa_quantiles.R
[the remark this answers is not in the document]
simonpcouch: Hannah's suggestion would adhere to the tidyverse style guide, which I'd say we generally ought to aim for.

simonpcouch (approved):
Woot woot!

R/predict_quantile.R
```diff
@@ -18,12 +23,11 @@ predict_quantile.model_fit <- function(object, new_data, ...) {
   new_data <- prepare_data(object, new_data)
 
   # preprocess data
-  if (!is.null(object$spec$method$pred$quantile$pre)) {
+  if (!is.null(object$spec$method$pred$quantile$pre))
     new_data <- object$spec$method$pred$quantile$pre(new_data, object)
-  }
```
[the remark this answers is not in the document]
topepo: That's the original (https://github.com/tidymodels/parsnip/blob/main/R/predict_quantile.R#L26). I changed it earlier but reverted the change to the results you see. I'll fix it.

tests/testthat/test-aaa_quantiles.R
```diff
@@ -0,0 +1,51 @@
...
+  v <- quantile_pred(matrix(1:20, 5), 1:4 / 5)
+  expect_s3_class(v, "quantile_pred")
+  expect_identical(attr(v, "quantile_levels"), 1:4 / 5)
+  expect_identical(
+    vctrs::vec_data(v),
+    lapply(vctrs::vec_chop(matrix(1:20, 5)), drop)
+  )
+})
+
+test_that("quantile_pred formatting", {
```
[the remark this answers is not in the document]
topepo: We do test that in `test-linear_reg_quantreg.R()` since we made objects there. I'll add some here

topepo:
@dajmcdon Merge time! 

One note... after looking into performance metrics, we realized that `quantile_pred()` and its methods will also be needed in yardstick. We want to avoid adding parsnip as a dependency there. 

For these cases, we move infrastructure to hardhat (which everyone depends on at some level). After I merge this, I’ll PR into hardhat to put the functions there, and then PR to parsnip to remove them.

PR #1204, opened 2024-09-17
Stricter sparsevctrs tests

EmilHvitfeldt:
All sparsvctrs tests now have option set such that unintentional materialization results in an error.

In addition, we now error on formula interface for sparse data, due to the `model.matrix()` issue we are running into

### DIFF
--- a/R/convert_data.R
+++ b/R/convert_data.R
@@ -48,6 +48,13 @@
     )
   }
 
+  if (is_sparse_tibble(data)) {
+    cli::cli_abort(
+      "Sparse data cannot be used with formula interface. Please use 
+      {.fn fit_xy} instead."
+    )
+  }
+
   if (remove_intercept) {
     data <- data[, colnames(data) != "(Intercept)", drop = FALSE]
   }
--- a/tests/testthat/test-sparsevctrs.R
+++ b/tests/testthat/test-sparsevctrs.R
@@ -1,4 +1,4 @@
-test_that("sparse tibble can be passed to `fit()", {
+test_that("sparse tibble can be passed to `fit() - supported", {
   skip_if_not_installed("xgboost")
   withr::local_options("sparsevctrs.verbose_materialize" = 3)
 
@@ -12,19 +12,21 @@ test_that("sparse tibble can be passed to `fit()", {
     error = TRUE,
     lm_fit <- fit(spec, avg_price_per_room ~ ., data = hotel_data)
   )
+})
+
+test_that("sparse tibble can be passed to `fit() - unsupported", {
+  hotel_data <- sparse_hotel_rates(tibble = TRUE)
 
   spec <- linear_reg() %>%
     set_mode("regression") %>%
     set_engine("lm")
-
-  withr::local_options("sparsevctrs.verbose_materialize" = NULL)
 
   expect_snapshot(
     lm_fit <- fit(spec, avg_price_per_room ~ ., data = hotel_data[1:100, ])
   )
 })
 
-test_that("sparse matrix can be passed to `fit()", {
+test_that("sparse matrix can be passed to `fit() - supported", {
   skip_if_not_installed("xgboost")
   withr::local_options("sparsevctrs.verbose_materialize" = 3)
   
@@ -39,7 +41,10 @@ test_that("sparse matrix can be passed to `fit()", {
     lm_fit <- fit(spec, avg_price_per_room ~ ., data = hotel_data)
   )
 
-  withr::local_options("sparsevctrs.verbose_materialize" = NULL)
+})
+
+test_that("sparse matrix can be passed to `fit() - unsupported", {
+  hotel_data <- sparse_hotel_rates()
 
   spec <- linear_reg() %>%
     set_mode("regression") %>%
@@ -50,12 +55,11 @@ test_that("sparse matrix can be passed to `fit()", {
   )
 })
 
-test_that("sparse tibble can be passed to `fit_xy()", {
+test_that("sparse tibble can be passed to `fit_xy() - supported", {
   skip_if_not_installed("xgboost")
+  withr::local_options("sparsevctrs.verbose_materialize" = 3)
   
   hotel_data <- sparse_hotel_rates(tibble = TRUE)
-  
-  withr::local_options("sparsevctrs.verbose_materialize" = 3)
 
   spec <- boost_tree() %>%
     set_mode("regression") %>%
@@ -64,8 +68,10 @@ test_that("sparse tibble can be passed to `fit_xy()", {
   expect_no_error(
     lm_fit <- fit_xy(spec, x = hotel_data[, -1], y = hotel_data[, 1])
   )
+})
 
-  withr::local_options("sparsevctrs.verbose_materialize" = NULL)
+test_that("sparse tibble can be passed to `fit_xy() - unsupported", {
+  hotel_data <- sparse_hotel_rates(tibble = TRUE)
 
   spec <- linear_reg() %>%
     set_mode("regression") %>%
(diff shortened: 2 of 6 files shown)


R/convert_data.R
```diff
@@ -48,6 +48,13 @@
     )
   }
 
+  if (is_sparse_tibble(data)) {
+    cli::cli_abort(
+      "Sparse data cannot be used with formula interface. Please use 
+      {.fn fit_xy} instead."
+    )
```
simonpcouch, line 55:
```suggestion
    cli::cli_abort(c(
      "Sparse data cannot be used with formula interface by default.",
      "i" = "Use {.fn fit_xy} instead to preserve sparsity or set the option
             {.help [{.code sparsevctrs.verbose_materialize = NULL}](TODO::LINK_HERE)}
             to allow for materializing sparse vectors."
    ))
```

I think we ought to let people know they can bypass this restriction in the error message. Is there a help-page documenting this option that we can link to?

tests/testthat/test-sparsevctrs.R
```diff
@@ -1,39 +1,47 @@
 test_that("sparse tibble can be passed to `fit()", {
   skip_if_not_installed("xgboost")
+  withr::local_options("sparsevctrs.verbose_materialize" = 3)
```
simonpcouch, line 3: Could we add a comment referencing documentation on this option once per test block? My first questions when I saw this were 1) what is the default and 2) what do different values of this option mean, and I wasn't able to find that information without looking at the source code.

tests/testthat/test-sparsevctrs.R
```diff
@@ -94,6 +110,11 @@ test_that("sparse tibble can be passed to `predict()", {
 
   hotel_data <- sparse_hotel_rates()
   hotel_data <- sparsevctrs::coerce_to_sparse_tibble(hotel_data)
+    
+  # materialize outcome
```
simonpcouch, line 114:
```suggestion
  withr::local_options("sparsevctrs.verbose_materialize" = NULL)
  # materialize outcome
```

This test currently depends on the option not having been set.

tests/testthat/test-sparsevctrs.R
```diff
@@ -1,39 +1,47 @@
...
 
   spec <- boost_tree() %>%
     set_mode("regression") %>%
     set_engine("xgboost")
-
-  expect_no_error(
+  
+  expect_snapshot(
+    error = TRUE,
     lm_fit <- fit(spec, avg_price_per_room ~ ., data = hotel_data)
```
simonpcouch, line 14:
```suggestion
    xgb_fit <- fit(spec, avg_price_per_room ~ ., data = hotel_data)
```

tests/testthat/test-sparsevctrs.R
```diff
@@ -1,39 +1,47 @@
...
   )
 
   spec <- linear_reg() %>%
     set_mode("regression") %>%
     set_engine("lm")
 
+  withr::local_options("sparsevctrs.verbose_materialize" = NULL)
```
simonpcouch, line 21:
```suggestion
  # the default value:
  withr::local_options("sparsevctrs.verbose_materialize" = NULL)
```

simonpcouch: Would be nice to drop this in where it's used.

tests/testthat/test-sparsevctrs.R
simonpcouch: These tests now test:

1) Set option to 3, try engine that supports sparsity
2) Set option to NULL (its default), try engine that doesn't support sparsity

The thing that I'm mostly interested in, and that we're not currently testing, is what happens when a user "does nothing" re: the option but supplies sparse data to an engine that supports it? Would also be nice to test option equal to 3 with engine that supports sparsity.

simonpcouch (commented):
Related to https://github.com/tidymodels/workflows/issues/246. 

I agree that this is the direction we want to be taking! 

Leaving review as Comment due to the comment beginning with "These tests now test..."—I think we ought to be testing each of the 4 options x engine_supports_sparsity combinations, but are only testing 2.

R/convert_data.R
EmilHvitfeldt: I'm setting `sparsevctrs.verbose_materialize` in the tests to make sure things work correctly. But the option is generally not meant to be used by the user. Sparse vectors will materialize silently when needs be. So you don't have to set any options.

I can be less opinionated and switch to a warning here instead 😄

tests/testthat/test-sparsevctrs.R
EmilHvitfeldt: > Could we add a comment referencing documentation on this option once per test block? My first questions when I saw this were 1) what is the default and 2) what do different values of this option mean, and I wasn't able to find that information without looking at the source code.

Yes, i can add that. https://github.com/r-lib/sparsevctrs/issues/74

tests/testthat/test-sparsevctrs.R
EmilHvitfeldt: I think this and some of the other comments above could be aliviated by splitting up some of these `test_that()`.
So instead of doing

```r
test_that("sparse tibble can be passed to `fit()", {
  withr::local_options("sparsevctrs.verbose_materialize" = 3)

  # test with model that support sparsity

  withr::local_options("sparsevctrs.verbose_materialize" = NULL)

  # test with model that doesn't support sparsity
})
```

we rewrite it as

```r
test_that("sparse tibble can be passed to `fit() - supported", {
  withr::local_options("sparsevctrs.verbose_materialize" = 3)

  # test with model that support sparsity
})

test_that("sparse tibble can be passed to `fit() - unsupported", {
  # test with model that doesn't support sparsity
})
```

tests/testthat/test-sparsevctrs.R
EmilHvitfeldt: 2 times 2 choices. I would argue that we only need to test 2 of them.

### Models that support sparsity

Setting `"sparsevctrs.verbose_materialize" = NULL` (default) would mean that the model fits as expected, regardless of if the sparse data materializes or not.

Setting `"sparsevctrs.verbose_materialize" = 3` (throw errors on materialization) would give us indications that we did something wrong, as the sparse data should not be materialized for models that support sparsity.

Any test that passes `"sparsevctrs.verbose_materialize" = 3` would by definition also pass with `"sparsevctrs.verbose_materialize" = NULL`.

### Models that doesn't support sparsity

Setting `"sparsevctrs.verbose_materialize" = NULL` (default) would throw a warning letting the user know that the model doesn't work and that it is converted to non-sparse.

```
`data` is a sparse tibble, but `linear_reg()` with engine "lm" doesn't accept that. 
Converting to non-sparse. 
```

Setting `"sparsevctrs.verbose_materialize" = 3` (throw errors on materialization) would give the same warning as above, followed by an error when the sparse data is materialized.

So I think the default `NULL` makes more sense here.

simonpcouch: Heard, got it. I'm on board for just testing those two options, then—as long as https://github.com/tidymodels/parsnip/pull/1204#discussion_r1765714029 incorporates some commenting that addresses these points above then I'm with it!

R/convert_data.R
simonpcouch: Okay, two thoughts on revisiting:

1) I agree that the default behavior should be to error here. No need to convert to warning. :)
2) Before, I thought it'd be nice if users could get around this error and allow for materializing sparse data in `fit()`, but now that I'm looking at this again I'm realizing that users won't have the analogous option for workflows, so I'm fine with ignoring this comment and keeping that option developer-facing only.

PR #1207, opened 2024-09-23
quantile regression updates for new hardhat model

topepo:
Depends on hardhat >= 1.4.0.9002. 

The main update is that the hardhat code is standardized on the plural `quantile_levels`.


topepo:
Closed and reopened to trigger CI for `branch != main`

tests/testthat/test-linear_reg_quantreg.R
```diff
@@ -1,12 +1,15 @@
 test_that('linear quantile regression via quantreg - single quantile', {
   skip_if_not_installed("quantreg")
+  skip_if_not_installed("hardhat", minimum_version = "1.4.0.9002")
+
+  library(hardhat)
```
hfrick, line 5: Can we avoid that `library()` call here?

tests/testthat/_snaps/quantile-reg-specs.md
hfrick, line 1: These snapshots also look like they are meant to test functionality of `hardhat::check_quantile_levels()` so as with the tests in `test-augment.R`, I think we only need one of them.

hfrick (approved):
Looks good!

Issue #1203, opened 2024-09-12
update `predict.modelfit(type = "quantile")`


topepo:
We are adding a mode for quantile regression but have one engine that already enables such prediction (using the censored regression mode). 

We should allow that but make some adjustments to harmonize both approaches.

topepo:
Some notes... 

The problem is that we have a `predict()` method that takes `type = "quantile"`. 

With the new quantile regression mode, we specify the quantile levels with `set_mode()`. The current `predict()` method has a `quantile` argument, which is problematic. 

A few models have quantile prediction methods. Two survival engines for parametric models (`flexsurv` and `survival`) have methods. Also, the bayesian package has this prediction type of regression and classification models. 

Proposed changes: 

 - [ ] Throw a warning if a model uses `type = "quantile"` when the mode is regression. Other modes use the existing interface.
 - [ ] Currrently we do not produce any interval estimate for quantile predictions, removed the `interval` and `level` arguments. 
 - [ ] We will want to transition the quantile prediction type to be specifically reserved for regression models built to predict quantiles. Some ordinary regression models can compute quantiles of the prediction distribution, but these are not optimized for accuracy.  We can grandfather the existing censored regression models as-is since they cannot have a censored regression mode _and_ a quantile regression mode. 
 - [ ] We will want to enforce a specific tidy format that mirrors our new format using a `pred_quantile` vctrs class. This will require a breaking change to the existing engines.

topepo:
Regarding the bayesian package... it will be a breaking change. However, the package doesn't really follow any of our guidelines for naming arguments/prediction columns and using tidy data formats. 

``` r
library(tidymodels)
library(bayesian)
#> Loading required package: brms
#> Loading required package: Rcpp
#> 
#> Attaching package: 'Rcpp'
#> The following object is masked from 'package:rsample':
#> 
#>     populate
#> Loading 'brms' package (version 2.21.0). Useful instructions
#> can be found by typing help('brms'). A more detailed introduction
#> to the package is available through vignette('brms_overview').
#> 
#> Attaching package: 'brms'
#> The following object is masked from 'package:dials':
#> 
#>     mixture
#> The following object is masked from 'package:stats':
#> 
#>     ar
```

``` r
# regression example

bayesian_fit <-
  bayesian() %>%
  set_mode("regression") %>%
  set_engine("brms") %>%
  fit(
    rating ~ treat + period + carry + (1 | subject),
    data = inhaler
  )
[...]

Issue #1206, opened 2024-09-20
`deprecate_stop()` for `surv_reg()`


hfrick:
How long do we want to have `surv_reg()` sit in the warning stage of the deprecation? It's been warning on CRAN since *Apr 12, 2023*, so over a year now. Do we typically give people longer than that?

History:
- `deprecate_warn()` since Mar 30, 2023 on dev, released in parsnip 1.1.0 on *Apr 12, 2023*
- `deprecate_soft()` since Mar 23, 2021 on dev

simonpcouch:
I support transitioning to `deprecate_stop`!👍

PR #1208, opened 2024-09-25
Change to `quantile` argument to `quantile levels`

topepo:
For #1203

After this, we will 
- merge the `quantile-mode` branch to main,
- update GHA for the extra branch, and 
- will create a PR for censored with updates to those engines.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -14,6 +14,7 @@
 ## Breaking Change
 
 * For quantile prediction, the `predict()` argument has been changed from `quantile` to `quantile_levels` for consistency. This does not affect models with mode `"quantile regression"`. 
+* The quantile regression prediction type was disabled for the deprecated `surv_reg()` model. 
 
 # parsnip 1.2.1
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,9 @@
 
 * New `extract_fit_time()` method has been added that returns the time it took to train the model (#853).
 
+## Breaking Change
+
+* For quantile prediction, the `predict()` argument has been changed from `quantile` to `quantile_levels` for consistency. This does not affect models with mode `"quantile regression"`. 
 
 # parsnip 1.2.1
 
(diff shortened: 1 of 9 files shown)


tests/testthat/test-surv_reg_survreg.R
```diff
@@ -46,7 +44,7 @@ test_that('survival execution', {
 })
 
 test_that('survival prediction', {
-  skip_on_travis()
+  skip_if_not_installed("censored", minimum_version = "0.3.2.9001")
```
hfrick, line 47: Why do we need censored for this here? These are tests for the (deprecated) `surv_reg()` for which the engines are in parsnip itself. There is a `survival::Surv()` call in the model formula, so I'm thinking this might actually be a `skip_if_not_installed("survival")`.

NEWS.md
```diff
@@ -11,6 +11,9 @@
 
 * New `extract_fit_time()` method has been added that returns the time it took to train the model (#853).
 
+## Breaking Change
+
+* For quantile prediction, the `predict()` argument has been changed from `quantile` to `quantile_levels` for consistency. This does not affect models with mode `"quantile regression"`.
```
hfrick, line 16: Since the quantile regression mode itself is new, I would drop that second sentence.
```suggestion
* For quantile prediction, the `predict()` argument has been changed from `quantile` to `quantile_levels` for consistency. 
```

hfrick (approved):

tests/testthat/test-surv_reg_survreg.R
topepo: `skip_if_not_installed("survival")` would help, but some of the post-processing code that this (deprecated) model will require will be censored.  

Obviously, I don't plan on having parsnip import censored. I think that we should disable this type of prediction for `surv_reg()`. Any objections to that?

NEWS.md
topepo: Disagree. If someone is just reading the news, this would be informative (and otherwise there's no harm in leaving it in).

tests/testthat/test-surv_reg_survreg.R
hfrick: No objections 👍 `surv_reg()` is not going to be around much longer anyway: https://github.com/tidymodels/parsnip/issues/1206

PR #1209, opened 2024-09-26
Add quantile regression mode

topepo:
Culmination of #1191, #1207, and #1208.

cc @dajmcdon

- Adds the new mode.
- Allows users to set quantile levels in `set_mode()`.
- Changes the argument to `predict_quantile()` from `quantile` to `quantile_levels`
- Removes quantile prediction methods for deprecated `surv_reg()` model. 
- Infrastructure to use the new vctrs object for quantile predictions.
- Adds a "quantreg" engine to `linear_reg()`.


dajmcdon:
Thanks for all your work on this @topepo and others!

If I were to make additional engines, would you like them in PRs here, or just elsewhere for now?

topepo:
> If I were to make additional engines, would you like them in PRs here, or just elsewhere for now?

If we can avoid wrapper and more imports, then go ahead and put them here. Were you thinking of making another package to house new methods?

hfrick (commented):
This looks good so far. I'm having second thoughts on the name swap for the quantiles/quantile_levels argument: I'd be in favor of deprecating that argument rather than changing it without warning.

Since you are not only adding a new mode but also a new engine with it, could you go over the checklists in https://github.com/tidymodels/tidymodels/pull/97/files? I think the PR is still missing the updates to the model registry, docs, examples, etc.

R/aaa_models.R
```diff
@@ -194,9 +194,9 @@ stop_missing_engine <- function(cls, call) {
   )
 }
 
-check_mode_for_new_engine <- function(cls, eng, mode, call = caller_env()) {
-  all_modes <- get_from_env(paste0(cls, "_modes"))
-  if (!(mode %in% all_modes)) {
+check_mode_for_new_engine <- function(cls, eng, mode) {
+  model_modes <- get_from_env(paste0(cls, "_modes"))
+  if (!(mode %in% model_modes)) {
```
EmilHvitfeldt, line 199: I would revert this change. Largely because it breaks the pattern used other places in the same file

R/arguments.R
```diff
@@ -111,11 +118,21 @@ set_mode.model_spec <- function(object, mode) {
 
   object$mode <- mode
   object$user_specified_mode <- TRUE
+  if (mode == "quantile regression") {
+      hardhat::check_quantile_levels(quantile_levels)
+  } else {
+    if (!is.null(quantile_levels)) {
+      cli::cli_abort("{.arg quantile_levels} is only used when the mode is
```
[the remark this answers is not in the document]
EmilHvitfeldt: i would suggest warning here as well. since the presence of the argument doesn't matter when you don't have the mode `"quantile regression"`. 

It will just be a reminder "hey! you left this argument here, but I can continue without it"

R/parsnip-package.R
```diff
@@ -21,7 +21,7 @@
 #' @importFrom stats .checkMFClasses .getXlevels as.formula binomial coef
 #' @importFrom stats delete.response model.frame model.matrix model.offset
 #' @importFrom stats model.response model.weights na.omit na.pass predict qnorm
-#' @importFrom stats qt quantile setNames terms update
+#' @importFrom stats qt quantile setNames terms update median
```
EmilHvitfeldt, line 24: Why are we importing `median()`? I don't see it used in this PR

R/linear_reg_data.R
```diff
@@ -582,3 +583,67 @@ set_pred(
...
+  value = list(
+    pre = NULL,
+    post = NULL,
+    func = c(fun = "predict"),
+    args =
+      list(
+        object = expr(object$fit),
+        newdata = expr(new_data),
+        type = "response",
+        rankdeficient = "simple"
```
EmilHvitfeldt, line 629: This new method is missing model-engine documentation such as https://github.com/tidymodels/parsnip/blob/main/man/rmd/linear_reg_glmnet.Rmd.

Adding it here, since it would be the place where we would say that we default to `rankdeficient = "simple"` and why

R/predict_quantile.R
```diff
@@ -1,48 +1,63 @@
 #' @keywords internal
 #' @rdname other_predict
-#' @param quantile A vector of numbers between 0 and 1 for the quantile being
-#'  predicted.
+#' @param quantile_levels  A vector of values between zero and one for the
```
EmilHvitfeldt, line 3:
```suggestion
#' @param quantile_levels  A vector of values between 0 and 1 for the
```

R/predict_quantile.R
```diff
@@ -1,48 +1,63 @@
...
 #' @method predict_quantile model_fit
 #' @export predict_quantile.model_fit
 #' @export
 predict_quantile.model_fit <- function(object,
                                        new_data,
-                                       quantile = (1:9)/10,
+                                       quantile_levels = NULL,
                                        interval = "none",
                                        level = 0.95,
                                        ...) {
```
EmilHvitfeldt, line 16:
```suggestion
check_dots_empty()
```

R/predict_quantile.R
```diff
@@ -1,48 +1,63 @@
...
 #' @inheritParams predict.model_fit
 #' @method predict_quantile model_fit
 #' @export predict_quantile.model_fit
 #' @export
 predict_quantile.model_fit <- function(object,
                                        new_data,
-                                       quantile = (1:9)/10,
+                                       quantile_levels = NULL,
                                        interval = "none",
                                        level = 0.95,
```
EmilHvitfeldt, line 14: Do we still need the `level` argument here?

R/print.R
```diff
@@ -23,6 +23,10 @@ print_model_spec <- function(x, cls = class(x)[1], desc = get_model_desc(cls), .
     print(show_call(x))
   }
 
+  if ( x$mode == "quantile regression" ) {
```
EmilHvitfeldt, line 26:
```suggestion
  if (x$mode == "quantile regression") {
```

EmilHvitfeldt (commented):
looking good so far! left some suggested changes and questions

R/aaa_models.R
topepo: We have a global variable named `all_modes` and this change avoids confusion regarding which data are being used.

R/parsnip-package.R
topepo: I'll remove. That calculation was moved into hardhat

R/predict_quantile.R
topepo: Yes. Hannah needs it for some censored engines

R/aaa-import-standalone-types-check.R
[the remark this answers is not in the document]
topepo: The [current version](https://raw.githubusercontent.com/r-lib/rlang/refs/heads/main/R/standalone-types-check.R) in rlang (as of 2024-08-15). 

It is confusing since it has 

```
# last-updated: 2023-03-13
```

as well as

```
# 2024-08-15:
# - `check_character()` gains an `allow_na` argument (@martaalcalde, #1724)
```

PR #1210, opened 2024-09-30
don't turn sparse matrix into dense matrix for glmnet prediction

EmilHvitfeldt:
What the title says:

there isn't really an easy way to test for this. This doesn't show up in our current tests as this is outside the scope of {sparsevctrs}


simonpcouch (approved):
Awesome!

I do think we ought find a way to test this before merging, either with a `verbose_materialize` and/or just passing some sparse data and making sure output looks like something we'd expect.

EmilHvitfeldt:
i got it done with some mocking 🎉 

https://github.com/tidymodels/parsnip/pull/1210/commits/ba56e2a68fe30c4002a8711259d777a80e4c6c42

PR #1216, opened 2024-10-24
workshop protected arguments warning

simonpcouch:
I came across this warning while spring cleaning bonsai and felt that it was reasonable that we tested for it but not that we screenshotted it. This PR adds a warning subclass and corrects pluralization in the message.

### DIFF
--- a/R/arguments.R
+++ b/R/arguments.R
@@ -20,11 +20,12 @@ check_eng_args <- function(args, obj, core_args) {
   protected_args <- unique(c(obj$protect, core_args))
   common_args <- intersect(protected_args, names(args))
   if (length(common_args) > 0) {
-    args <- args[!(names(args) %in% common_args)]
-    common_args <- paste0(common_args, collapse = ", ")
     cli::cli_warn(
-      "The argument{?s} {.arg {common_args}} cannot be manually
-       modified and {?was/were} removed."
+      c(
+        "The argument{?s} {.arg {common_args}} cannot be manually modified
+         and {?was/were} removed."
+      ),
+      class = "parsnip_protected_arg_warning"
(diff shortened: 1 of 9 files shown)


tests/testthat/test-arguments.R
simonpcouch, line 1: Note that this is a totally new test file. This warning is tested in a couple engine-specific files already—I just test for the subclass now.

tests/testthat/_snaps/mlp.md
```diff
@@ -32,25 +32,6 @@
       x Engine "wat?" is not supported for `mlp()`
       i See `show_engines("mlp")`.
 
----
```
simonpcouch, line 35: Each of these removed snapshots both tested duplicates of the print method and this warning message.

simonpcouch:
I imagine this will cause snap changes in many places in extratests and parsnip extensions. Those should be able to just test the warning class now.😜

R/arguments.R
```diff
@@ -20,11 +20,12 @@ check_eng_args <- function(args, obj, core_args) {
   protected_args <- unique(c(obj$protect, core_args))
   common_args <- intersect(protected_args, names(args))
   if (length(common_args) > 0) {
-    args <- args[!(names(args) %in% common_args)]
```
hfrick, line 23: You do not want to remove this line :)

hfrick (changes requested):
Generally totally on board with this but hitting "request changes" to make that part obvious. Happy for you to merge after that's fixed 👍

R/arguments.R
simonpcouch: Ah, thank you!

Issue #953, opened 2023-04-27
Ordinal mode (between classification and regression)


Tripartio:
I think that a major missing gap in parsnip is explicit support for ordinal models, by which I mean **models where the response variable is an ordered factor**.

My proposal here is a follow up to [closed issue 35](https://github.com/tidymodels/parsnip/issues/35), specifically focusing on **ordinal models**. @topepo wrote:

> The main issue that I had as about how to organize the functions. In parsnip we try to have the main model functions describe the structural aspects of the model (e.g. linear_reg(), rand_forest(), etc). For the ordinal models based on generalized linear model (e.g. cumulative logits, adjacent categories etc), my thinking was to have:
> 
> ordinal_cumulative(link = "logit", odds = "proportional")
> ordinal_adjacent(link = "logit", odds = "proportional")
> and so on. My thinking is that people would probably want to look at the parallel assumption (assuming they have the right design for that) and tuning over the odd argument would be helpful.
> 
> How would you like to see these types of model organized?
> 

I am not a statistical expert on ordinal models. but I have used them in my research code and I have studied to learn about different kinds. Based on my understanding, the examples above are the kind of detail that parsnip is supposed to abstract away. I would think that users care mainly about three things:
* The function call to the model versus the mode of the model
* The broom tidier functions for model outputs (tidy, glance, augment)
* The prediction results

## Function call versus mode

In his initial thoughts, @topepo seemed to frame ordinal models as a parsnip model type with names like `ordinal_cumulative` or `ordinal_adjacent`. But fundamentally, in parsnip terms, **I do not think of an ordinal model as a model type; I think of it as a mode**, just like classification and regression are modes. An ordinal model is simply one where the response variable type is neither an unordered class nor a real number but rather something in between, an ordered category. This is borne out by his mention of various types of ordinal (glmnetcr, ordinalNet, ordinalForest, party models for trees, and brms models). Each of these would correspond to their various underlying model types, be it `logistic_reg` or `rand_forest`, with their corresponding engines. `ordinal` would simply be the mode.

[...]

PR #1217, opened 2024-10-25
predict ordinal factors from ordinal regression models

topepo:
Currently, we accidentally strip off the `ordered` element of factors when we make class predictions. 

Relevant to #953


hfrick (approved):
Looks good!

PR #1218, opened 2024-11-02
fix likely typo 'linear_reg' in 'logistic_reg' documentation

corybrunson:
This example illustrates a non-default model family in an apparent wrong model type, which does not produce an error so would not have been caught during checks.

I apologize if the use of `linear_reg()` was intended, but in that case it should probably be explained.


simonpcouch:
Thank you, @corybrunson!

PR #1219, opened 2024-11-13
Divert linear regressions with poisson family to poisson_reg()

topepo:
Motivated by https://github.com/tidymodels/tune/issues/956

There are _so_ many ways people can pass the family to `glm()` and `glmnet()`. This checks for all four methods that I'm aware of.


tests/testthat/test-linear_reg.R
```diff
@@ -358,3 +358,33 @@ test_that("check_args() works", {
...
+      set_engine("glmnet", family = stats::poisson()) %>%
+      translate(),
+    error = TRUE
+  )
+  expect_snapshot(
+    linear_reg(penalty = 1) %>%
+      set_engine("glmnet", family = "poisson") %>%
+      translate(),
+    error = TRUE
+  )
```
hfrick, line 387: Love the clarity of these tests

hfrick (approved):
Nice one! For your consideration but not a request: Would it make sense to move those checks to `check_args.linear_reg()` or would checks on engine args in addition to main args mess with anything else?

PR #1224, opened 2024-12-04
function to get prediction columns

topepo:
In service of tidymodels/workflows#273

For a fitted model or workflow, return the column names that will contain predictions for the main types given by `augment()`.

Tests for censored regression models and workflows will be in extratests


R/misc.R
```diff
@@ -575,3 +575,70 @@ is_cran_check <- function() {
...
+    if (any(predict_types == "prob")) {
+      res$probabilities <- paste0(".pred_", x$lvl)
+    }
+  } else if (model_mode == "censored regression") {
+    res$estimate <- ".pred_time"
+    if (any(predict_types %in% c("survival"))) {
+      res$probabilities <- ".pred"
+    }
+  } else {
+    cli::cli_abort("Unsupported model mode {model_mode}.")
```
topepo, line 637: I don't know that this branch of the if/then will ever be encountered, given the error check above. However, we might hit is when we have a mode for quantile regression data so I'd err on the side of leaving it in, untested.

EmilHvitfeldt: i think that is reasonable

R/misc.R
```diff
@@ -575,3 +575,74 @@ is_cran_check <- function() {
...
```
EmilHvitfeldt, line 641: I like to do this

```suggestion
    # Should be unreachable
    cli::cli_abort("Unsupported model mode {model_mode}.")
```

R/misc.R
```diff
@@ -575,3 +575,74 @@ is_cran_check <- function() {
 }
 # nocov end
 
+# ------------------------------------------------------------------------------
+
+#' Obtain names of prediction columns for a fitted model or workflow
+#'
+#' [.get_prediction_column_names()] returns a list that has the names of the
+#' columns for the primary prediction types for a model.
+#' @param x A fitted model (class `"model_fit"`) or a fitted workflow.
```
EmilHvitfeldt, line 584:
```suggestion
#' @param x A fitted parsnip model (class `"model_fit"`) or a fitted workflow.
```

R/misc.R
```diff
@@ -575,3 +575,74 @@ is_cran_check <- function() {
...
+#' @param syms Should the column names be converted to symbols?
```
EmilHvitfeldt, line 585:
```suggestion
#' @param syms Should the column names be converted to symbols? Defaults to `FALSE`.
```

EmilHvitfeldt (approved):
looking good. very clean

R/misc.R
```diff
@@ -575,3 +575,75 @@ is_cran_check <- function() {
...
+#' library(dplyr)
+#' library(modeldata)
+#' data("two_class_dat")
+#'
+#' levels(two_class_dat$Class)
+#' lr_fit <- logistic_reg() %>% fit(Class ~ ., data = two_class_dat)
+#'
+#' .get_prediction_column_names(lr_fit)
+#' .get_prediction_column_names(lr_fit, syms = TRUE)
+#' @export
```
hfrick, line 597: Could we make this function `@keyword internal`?

R/misc.R
```diff
@@ -575,3 +575,75 @@ is_cran_check <- function() {
...
+.get_prediction_column_names <- function(x, syms = FALSE) {
+  if (!inherits(x, c("model_fit", "workflow"))) {
+    cli::cli_abort("{.arg x} should be an object with class {.cls model_fit} or
+                    {.cls workflow}, not {.obj_type_friendly {x}}.")
+  }
+
+  if (inherits(x, "workflow")) {
+    x <- x %>% extract_fit_parsnip(x)
+  }
```
hfrick, line 606: With an eye toward the release cascade and trying to keep things as modular as possible, I think it would be good to make this function only for `model_spec` objects. This call to `extract_fit_parsnip()` seems to be all that's necessary for workflow objects. I'd move that to workflows or wherever this is going to be needed.

R/misc.R
```diff
@@ -575,3 +575,75 @@ is_cran_check <- function() {
...
+  inst_res <- purrr::map(required_pkgs(x), rlang::check_installed)
+  predict_types <-
+    get_from_env(paste0(model_type, "_predict")) %>%
+    dplyr::filter(engine == model_engine & mode == model_mode) %>%
+    purrr::pluck("type")
+
+  if (length(predict_types) == 0) {
+    cli::cli_abort("Prediction information could not be found for this
+                   {.fn {model_type}} with engine {.val {model_engine}} and mode
+                   {.val {model_mode}}. Does a parsnip extension package need to
```
hfrick, line 622: That last question would be very nice in an `i` bullet rather than the error message itself.

tests/testthat/test-misc.R
```diff
@@ -249,3 +249,53 @@ test_that('check_outcome works as expected', {
...
+  # in extratests
+
+  ### bad input
+  expect_snapshot(
+    .get_prediction_column_names(1),
+    error = TRUE
+  )
+
+  unk_fit <- ols_fit
+  unk_fit$spec$mode <- "Depeche"
```
hfrick, line 295: 😆

PR #1239, opened 2025-01-28
Switch to hardhat's version of `contr_one_hot()`

### LINKED ISSUE #1215: `contr_one_hot()` is also in hardhat but not exported
`contr_one_hot()` is also in parsnip (where it is exported). 

There are tests there that are a bit more extensive: https://github.com/tidymodels/parsnip/blob/main/tests/testthat/test-contr_one_hot.R

You can bring those over here and I'll open a parsnip issue about the duplication. hardhat would have to export this function for parsnip to use it.

_Originally posted by @topepo in https://github.com/tidymodels/hardhat/pull/268#discussion_r1812633876_

hfrick:
closes #1215, as a pair with https://github.com/tidymodels/hardhat/pull/278


hfrick:
That failed check is one on the parsnip overhead but since that test passes on all other platforms, I think it's not related to this change.

topepo:
Although I do see an additional warning for 

> Found the following significant warnings:
> Warning: replacing previous import ‘hardhat::contr_one_hot’ by 
> ‘parsnip::contr_one_hot’ when loading ‘viralmodels’

I'll add an issues to their repo.

PR #1248, opened 2025-02-01
Fail on NULL engine

### LINKED ISSUE #1242: NULL engines
The cli refactor led to the regression (picked up in revdeps) that allowed NULL for the engine: 

```r
set_engine(nearest_neighbor_adaptive(), NULL)
```

topepo:
Closes #1242

During our cli upgrades, we now fail (not gracefully) when `NULL` is passed as an engine. 

This PR fails with a nice error message.


R/engines.R
```diff
@@ -111,7 +111,7 @@ set_engine <- function(object, engine, ...) {
 set_engine.model_spec <- function(object, engine, ...) {
   mod_type <- class(object)[1]
 
-  if (rlang::is_missing(engine)) {
+  if (rlang::is_missing(engine) | is.null(engine)) {
```
simonpcouch, line 114:
```suggestion
  if (rlang::is_missing(engine) || is.null(engine)) {
```

I believe this change will stop evaluating the expression once `rlang::is_missing(engine)` returns TRUE to avoid the `argument "engine" is missing, with no default` error from `is.null()` that's introduced in the snapshots below.

tests/testthat/_snaps/args_and_modes.md
```diff
@@ -108,16 +108,16 @@
     Code
       linear_reg() %>% set_engine()
     Condition
-      Error in `set_engine()`:
-      ! Missing engine. Possible mode/engine combinations are: quantile regression {quantreg} and regression {lm, glm, glmnet, stan, spark, keras, brulee}.
+      Error in `set_engine.model_spec()`:
+      ! argument "engine" is missing, with no default
```
simonpcouch, line 112: This feels like a regression to me—I believe the proposed change above should resolve.

simonpcouch (approved):
🤌

PR #1299, opened 2025-10-27
generalized random forests

topepo:
Adds engines for classification, regression, and quantile regression. 

Tests in tidymodels/extratests#281

### DIFF
--- a/R/rand_forest_data.R
+++ b/R/rand_forest_data.R
@@ -148,7 +148,7 @@ grf_conf_int <- function(
       res <- vctrs::vec_cbind(res, std_err)
     }
   } else {
-    rlang::abort(
+    cli::cli_abort(
       "No confidence interval implementation for objects with class(es)
       {.cls {class(object$fit)[1]}}"
     )
--- a/R/rand_forest_grf.R
+++ b/R/rand_forest_grf.R
@@ -1,7 +1,7 @@
-#' Random forests via grf
+#' Generalized random forests via grf
 #'
 #' The \pkg{grf} fits models that create a large number of decision
-#' trees, each independent of the others. The final prediction uses all
+#' trees, each independent of the others. The final prediction uses
 #' predictions from the individual trees and combines them.
 #'
 #' @includeRmd man/rmd/rand_forest_grf.md details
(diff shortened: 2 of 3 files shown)


R/rand_forest_data.R
```diff
@@ -76,11 +92,83 @@ ranger_confint <- function(object, new_data, ...) {
...
+
+    res <- vctrs::vec_cbind(lowers, uppers)
+
+    if (object$spec$method$pred$conf_int$extras$std_error) {
+      std_err <- tibble::as_tibble(std_err, .name_repair = "minimal")
+      names(std_err) <- paste0(".std_error_", names(std_err))
+      res <- vctrs::vec_cbind(res, std_err)
+    }
+  } else {
+    rlang::abort(
```
hfrick: this woud be nice as `cli_abort()` with pluralization

R/rand_forest_data.R
```diff
@@ -603,8 +680,221 @@ set_pred(
...
-    args =
-      list(x = quote(object$fit),
-           dataset = quote(new_data))
+    args = list(x = quote(object$fit), dataset = quote(new_data))
+  )
+)
+
+
+# ------------------------------------------------------------------------------
+# grf components
```
hfrick, line 689: Nit: It'd be nice if the ordering here was consistent. Either each "thing" for all modes or everything for one mode, then the next mode.

R/rand_forest_grf.R
```diff
@@ -0,0 +1,13 @@
+#' Random forests via grf
+#'
+#' The \pkg{grf} fits models that create a large number of decision
+#' trees, each independent of the others. The final prediction uses all
+#' predictions from the individual trees and combines them.
```
hfrick, line 5: This reads very general, and usually we try to be a bit more specific in this description block. Is there anything else we could say here? I do realize that this is about generalized random forests so there might not be!

R/rand_forest_data.R
topepo: If I understand correctly, the pattern that is in parnsip is to run all of the `set_model_engine()` lines for each mode then all of the `set_dependency()` calls for each mode. 

I don't think that we are ordering anything different for grf than the others in this file or in other files.

R/rand_forest_grf.R
topepo: I made a very small tweak, but I don't really think that there is anything additional to say. It's a generalization of random forests (and not different from the other details files). We do have a reference, so people can look there for more details. 

The main reason we have that description block is to avoid saying "random forest" since some people may not know what it is. It's not meant to be very detailed.

PR #1320, opened 2026-01-14
Update "details" documents regarding case weights

topepo:
There were a number of models that support these but did not note it in the details section. Additional notes were added for some tree ensemble methods where the case weights are actually sampling weights. 

Also, updated the model information for `mlp()`'s `nnet` engine since _it does_ support case weights.

### DIFF
--- a/man/details_bart_dbarts.Rd
+++ b/man/details_bart_dbarts.Rd
@@ -73,9 +73,9 @@ times number of observations.
 ## Computational engine: dbarts 
 ## 
 ## Model fit template:
-## dbarts::bart(x = missing_arg(), y = missing_arg(), ntree = integer(1), 
-##     base = double(1), power = double(1), k = double(1), verbose = FALSE, 
-##     keeptrees = TRUE, keepcall = FALSE)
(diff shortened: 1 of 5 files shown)


tests/testthat/_snaps/registration.md
```diff
@@ -424,9 +424,10 @@
        modes: unknown, classification, regression 
       
        engines: 
-         classification: brulee, brulee_two_layer, keras, nnet
-         regression:     brulee, brulee_two_layer, keras, nnet
+         classification: brulee, brulee_two_layer, keras, nnet1
+         regression:     brulee, brulee_two_layer, keras, nnet1
```
EmilHvitfeldt, line 428: this doesn't seem right?

tests/testthat/_snaps/registration.md
```diff
@@ -424,9 +424,10 @@
...
       
+      1The model can use case weights.
```
EmilHvitfeldt, line 430: same here

tests/testthat/_snaps/registration.md
topepo: It is; ~~we~~ I didn't specify that we can use case weights. I enabled that so we get the footnote.

PR #1326, opened 2026-01-29
Keep quantile levels when modifying a model spec

### LINKED ISSUE #1304: set_engine() removes quantile levels
``` r
library(tidymodels)

qnt_lvls <- (1:3) / 4

linear_reg() |> 
  set_engine("quantreg") |> 
  set_mode("quantile regression", quantile_levels = qnt_lvls)
#> Linear Regression Model Specification (quantile regression)
#> 
#> Computational engine: quantreg
#> Quantile levels: 0.25, 0.5, and 0.75.

linear_reg() |> 
  set_mode("quantile regression", quantile_levels = qnt_lvls) |> 
  set_engine("quantreg")
#> Linear Regression Model Specification (quantile regression)
#> 
#> Computational engine: quantreg
#> Quantile levels: .
```

<sup>Created on 2025-11-14 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

We should create a constructor function for model specifications to avoid this issue in the future.

topepo:
Closes #1304

Updated the model spec constructor to include the quantile levels (if any) when modifying the spec.


tests/testthat/test-model_basics.R
```diff
@@ -98,3 +98,36 @@ test_that('basic object classes and print methods', {
...
+
+  rf_spec_2me <-
+    rf_spec_2m |>
+    set_engine("grf")
+
+  expect_snapshot(rf_spec_2me)
+
+  rf_spec_2em <-
+    rf_spec_2e |>
+    set_mode("quantile regression", quantile_levels = (1:3) / 4)
```
hfrick: I would set a different value for the quantile levels here so that we can see directly that the overwrite works.

PR #1328, opened 2026-02-04
linear predictor prediction type documentation

### LINKED ISSUE #1314: Document `type = "linear_pred"` better
`type = "linear_pred"` is not listed in the options for `type` but should be. It's mentioned in the section on censored regression but should also be mentioned in the Value/return section.

https://github.com/tidymodels/parsnip/blob/77e08226814f7f51849450671ca1600b00e630aa/R/predict.R#L9-L11

topepo:
Closes #1314


hfrick (approved):
Nice!

PR #1341, opened 2026-03-02
Better error message for misspelling of engine packages, for engines only in extension packages

### LINKED ISSUE #1110: unhelpful error if engine is miss-specified for extension packages
``` r
library(poissonreg)

spec <- poisson_reg() %>% 
  set_engine("gml") %>%
  set_mode("regression")

fit(spec, gear ~ ., data = mtcars)
#> Error in `fit()`:
#> ! formula_ is unknown.
```

<sup>Created on 2024-04-09 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

EmilHvitfeldt:
to close #1110

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: parsnip
 Title: A Common API to Modeling and Analysis Functions
-Version: 1.4.1.9003
+Version: 1.4.1.9004
 Authors@R: c(
     person("Max", "Kuhn", , "max@posit.co", role = c("cre", "aut"),
            comment = c(ORCID = "0000-0003-2402-136X")),


tests/testthat/test-poissonreg.R
```diff
@@ -0,0 +1,11 @@
+test_that("misspelled engine names give informative error for extension packages", {
```
EmilHvitfeldt: while this might have been better in a different test file, we need it here before poissonreg is loaded

topepo: Should this go in extratests?

topepo: Also, I would write a few tests that define the model in different orders, like: 

```r
poisson_reg() |>
    set_engine("gml") |>
    set_mode("regression")

poisson_reg() |>
    set_engine("gml") 

# etc
```

EmilHvitfeldt: both handled in https://github.com/tidymodels/extratests/pull/298

PR #1343, opened 2026-03-02
add `format_predictions()`

### LINKED ISSUE #927: Improve `format_<prediction type>()`
The formatting functions `format_num()`, `format_class()`, etc are fairly similar. Let's check if a general `format_parsnip()` makes sense.

EmilHvitfeldt:
To close #927

In addition in [`d98ec8a` (this PR)](https://github.com/tidymodels/parsnip/pull/1343/commits/d98ec8a96f35f5443e53d740a97950c8c2ee350c) I deprecated the old functions. We use some of them in aqua, and maybe some other people use it too, but it should be a fairly easy transition


R/predict.R
```diff
@@ -297,32 +285,79 @@ check_pred_type <- function(object, type, ..., call = rlang::caller_env()) {
...
+
+  switch(
+    type,
+    numeric = ensure_parsnip_format(x, ".pred", overwrite = FALSE),
+    class = ensure_parsnip_format(x, ".pred_class"),
+    prob = format_classprobs_impl(x),
+    time = ensure_parsnip_format(x, ".pred_time", overwrite = FALSE),
+    survival = ensure_parsnip_format(x, ".pred"),
+    linear_pred = ensure_parsnip_format(x, ".pred_linear_pred"),
+    hazard = ensure_parsnip_format(x, ".pred"),
```
topepo, line 314: We should probably check types `"conf_int"`, `"pred_int"`, and `"quantile"` too.

EmilHvitfeldt: done in [`5d7a0a2` (this PR)](https://github.com/tidymodels/parsnip/pull/1343/commits/5d7a0a2a602e4b8eb29acc2efd4bc0959bb1498a).

Note that we are explicitly handling `raw` as well. It would have been flowing through anyway but now we have parity with `pred_types` which is nice

Issue #548, opened 2021-08-29
Issue with `fit_xy()` in presence of missing values


p-lemercier:
I get an error message when trying to fit a model using `fit_xy()` in presence of missing values but not with `fit()`.

Here is just one example with `rand_forest() %>% set_engine("ranger")`. The same problem occurs with `rand_forest() %>% set_engine("randomForest")` or `linear_reg(penalty = 1, mixture = 1) %>% set_engine("glmnet")`.

``` r
library(tidymodels)
#> Registered S3 method overwritten by 'tune':
#>   method                   from   
#>   required_pkgs.model_spec parsnip

data(airquality)
summary(airquality)
#>      Ozone           Solar.R           Wind             Temp      
#>  Min.   :  1.00   Min.   :  7.0   Min.   : 1.700   Min.   :56.00  
#>  1st Qu.: 18.00   1st Qu.:115.8   1st Qu.: 7.400   1st Qu.:72.00  
#>  Median : 31.50   Median :205.0   Median : 9.700   Median :79.00  
#>  Mean   : 42.13   Mean   :185.9   Mean   : 9.958   Mean   :77.88  
#>  3rd Qu.: 63.25   3rd Qu.:258.8   3rd Qu.:11.500   3rd Qu.:85.00  
#>  Max.   :168.00   Max.   :334.0   Max.   :20.700   Max.   :97.00  
#>  NA's   :37       NA's   :7                                       
#>      Month            Day      
#>  Min.   :5.000   Min.   : 1.0  
#>  1st Qu.:6.000   1st Qu.: 8.0  
#>  Median :7.000   Median :16.0  
#>  Mean   :6.993   Mean   :15.8  
#>  3rd Qu.:8.000   3rd Qu.:23.0  
#>  Max.   :9.000   Max.   :31.0  
#> 

mod1 <- rand_forest() %>% 
  set_engine("ranger") %>% 
  set_mode("regression") %>% 
  fit(Ozone ~ Solar.R + Temp + Wind, data = airquality)

mod2 <- rand_forest() %>% 
  set_engine("ranger") %>% 
  set_mode("regression") %>% 
  fit_xy(x = airquality[,c("Solar.R", "Temp", "Wind")], y = airquality$Ozone)
#> Error: Missing data in columns: Solar.R.
#> Chronométrage arrêté à : 0 0 0.001
```

<sup>Created on 2021-08-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

[...]

PR #1344, opened 2026-03-02
properly catch `na.action`

EmilHvitfeldt:
Hoping to close https://github.com/tidymodels/parsnip/issues/548

I think Julia was right in her assessment in the original issue. But we should think a little more if this is right thing to do


R/convert_data.R
```diff
@@ -66,7 +66,7 @@
 
   ## Assemble model.frame call from call arguments
   mf_call <- quote(model.frame(formula, data))
-  mf_call$na.action <- match.call()$na.action # TODO this should work better
+  mf_call$na.action <- substitute(na.action)
```
topepo, line 75: This looks good but please add some comments here for future us to understand.

EmilHvitfeldt: done in [`373c735` (this PR)](https://github.com/tidymodels/parsnip/pull/1344/commits/373c73536dfdb8752471aad8c629590adfcc4aaa)

Issue #1298, opened 2025-10-16
add new model ordinal_reg() for ordinal regression


corybrunson:
This was discussed in two previous issues, #35 and #953. The new model is being added in [this fork branch](https://github.com/corybrunson/parsnip/tree/ordered), and an extension with engines is underway [here](https://github.com/corybrunson/ordered).

First see [this article](https://www.tidymodels.org/learn/develop/models/) for guidance on adding a new model.

### Checklist for adding a new model

Copied from [this article](https://parsnip.tidymodels.org/articles/checklists.html):

* [x] Create the main modeling function (e.g. `nearest_neighbors()`). When choosing argument names, check for existing ones across other model types which could be reused, e.g., `trees` for the number of trees is used across different tree-based models like `rand_forest()` and `boost_tree()`.
  - [here](https://github.com/corybrunson/parsnip/blob/59fe7023c77474127b9c1fab1512a3fd3f57ce79/R/ordinal_reg.R#L49)
  - Model parameters currently include `penalty`, `mixture`, `ordinal_link`, and `odds_link`.

* [x] Register the model and a mode in the parsnip model “database” with `parsnip::set_new_model()` and `parsnip::set_model_mode()`. [[example]](https://github.com/tidymodels/parsnip/blob/c54f07b7e1f7ce164aab8f95bc7b1356b68558c8/R/proportional_hazards_data.R)
  - [specification only](https://github.com/corybrunson/parsnip/blob/ordered/R/ordinal_reg_data.R)

* [x] Create an `update()` method.
  - [here](https://github.com/corybrunson/parsnip/blob/59fe7023c77474127b9c1fab1512a3fd3f57ce79/R/ordinal_reg.R#L84)

* [x] Add parameter objects for main parameters to *dials*, if needed. See [How to create a tuning parameter function](https://www.tidymodels.org/learn/develop/parameters/) for more details.
  - [here](https://github.com/corybrunson/dials/blob/ordered/R/param_ordinal_link.R)
  - This is done in [a coordinated fork branch](https://github.com/corybrunson/dials/tree/ordered).

* [x] Write unit tests for the basic modeling function (independent of the engines). 

* [x] Add a pkgdown entry.

* [x] Add model description to the internal `model_descs` tibble.

* [ ] Run `purrr::map(parsnip:::extensions(), ~ library(.x, character.only = TRUE))`.

* [ ] Run `parsnip:::update_model_info_file()`.

[...]

PR #1346, opened 2026-03-04
updates related to 'ordered' extenision package

corybrunson:
This PR introduces the `ordinal_reg()` model and six new engines for ordinal prediction: 4 for ordinal regression models, 1 for decision tree models, and 1 for random forest models. This prepares for a limited first release of the [{ordered} extension](https://github.com/corybrunson/ordered). Here are details copied from 'NEWS.md'—note that these are contained in a subsection there that might not be organized appropriately:

> The changes in this section are discussed in #1298.
> 
> * A new model `ordinal_reg()` is introduced for ordinal regression (#953):
>   * Its sole mode is `"classification"` and its default engine is `"polr"` for `MASS::polr()`.
>   * Additional engines `"ordinalNet"` for `ordinalNet::ordinalNet()` and `"vglm"` for `VGAM::vglm()` are documented.
>   * In addition to `penalty` and `mixture` common to GLM models, two new model parameters are introduced: `ordinal_link` and `odds_link`.
>   * Methods are written for `translate()`, `update()`, `check_args()`, and `tunable()`.
>   * Existing pan-model tests are extended and new model-specific tests are added.
> 
> * New engines for ordinal prediction are documented:
>   * `"vgam"` for `VGAM::vgam()` under `gen_additive_mod()`
>   * `"rpartScore"` for `rpartScore::rpartScore()` under `decision_tree()`
>   * `"ordinalForest"` for `ordinalForest::ordfor()` under `rand_forest()`
> 
> * Engine arguments for the new ordinal prediction engines are documented.
> 
> * Ordinal prediction tools are coordinated with new parameter tuners in dials and engines in ordered.
> 
> * `model_info_table` is updated with the above 6 engines in anticipation of submission of ordered to CRAN.

One outstanding task is to complete the engine documentation files. Minimal versions are provided, and i'd be glad to complete them after discussion with or illustration by a maintainer.

Another outstanding task is to run `purrr::map(parsnip:::extensions(), ~ library(.x, character.only = TRUE))` as stipulated on the checklist. This caused many deletions from 'inst/models.tsv' on my end, presumably because many dependencies are not installed.

A companion PR to {dials} is in preparation.


R/ordinal_reg.R
```diff
@@ -0,0 +1,257 @@
...
+        call = rlang::caller_env()
+      )
+    }
+  }
+
+  # adapted from `.check_glmnet_penalty_fit()`
+  if (engine == "ordinalNet") {
+
+    pen <- rlang::eval_tidy(x$args$penalty)
+    if (length(pen) >= 1L) {
```
topepo: I'm pretty sure that this should be `length(pen) != 1L` like it is here: 

https://github.com/tidymodels/parsnip/blob/9e009cf21c9979193fc7652b23b50e0c8113e907/R/glmnet-engines.R#L392

The rationale is that these models can fit multiple penalties simultaneously, but we want to specify a single penalty when defining the model. I'll make the change but flag me down if I'm incorrect

R/ordinal_reg.R
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    # adapted from `set_glmnet_penalty_path()`
+    if (any(names(x$eng_args) == "path_values")) {
+      x$method$fit$args$lambdaVals <- x$eng_args$path_values
+      x$eng_args$path_values <- NULL
+      x$method$fit$args$path_values <- NULL
+    } else {
+      # } else if (! rlang::is_call(x$method$fit$args$lambdaVals)) {
+      # REVIEW: `ordinalNet` models won't use values of `lambdaVals` at
```
topepo: This is the best approach for this model. The "submodel" approach can enable super-fast tuning, but this model only supports values along the path (rather than interpolating between them, as glmnet does).

topepo: _However_, I see that ordered sets `has_submodel = TRUE` in the model definition. I think that should be set to `FALSE`.

R/ordinal_reg.R
```diff
@@ -0,0 +1,251 @@
...
+        }
+      x$method$fit$args$lambdaMinRatio <- min_lambda
+      x$method$fit$args$includeLambda0 <- TRUE
+      x$method$fit$args$lambdaVals <- NULL
+    }
+    # Since the `fit` information is gone for the penalty, we need to have an
+    # evaluated value for the parameter.
+    x$args$penalty <- rlang::eval_tidy(x$args$penalty)
+
+    # REVIEW: Below is an alternative solution to
```
topepo: If this can eliminate the need for the wrapper, I'd say let's try to make this part work. This sort of interception is basically what the `translate()` methods are designed for.

R/ordinal_reg.R
```diff
@@ -0,0 +1,251 @@
...
+
+# ------------------------------------------------------------------------------
+
+#' @export
+translate.ordinal_reg <- function(x, engine = x$engine, ...) {
+  dots <- list(...)
+
+  x <- translate.default(x, engine, ...)
+
+  # REVIEW: What's the preferred way to flag when a legitimate model parameter
```
topepo, line 148: If you think that this will be a common issue, this method is fine. Otherwise, we would rely on the users reading the engine documentation to know that this is not an argument to polr.

R/ordinal_reg_vgam.R
```diff
@@ -0,0 +1,12 @@
+#' Ordinal regression via vector GAMs
```
topepo: It seems like this should be an engine for `gen_additive_mod()`, not `ordinal_reg()`.

topepo: If that's the case, just move it over to that model.

man/rmd/ordinal_reg_vglm.Rmd
```diff
@@ -0,0 +1,41 @@
...
+
+```{r vglm-oreg}
+library(ordered)
+
+ordinal_reg() %>% 
+  set_engine("vglm") %>% 
+  set_mode("classification") %>% 
+  translate()
+```
+
```
topepo, line 40: If you move this to be an engine to `gen_add_mod()`, take a look at the extra sections in the mgcv docs to help the users out: 

https://github.com/tidymodels/parsnip/blob/main/man/rmd/gen_additive_mod_mgcv.Rmd#L54-L110

man/rmd/ordinal_reg_ordinalNet.Rmd
```diff
@@ -0,0 +1,75 @@
...
+```{r}
+#| child: template-same-scale.Rmd
+```
+
+By default, [ordinalNet::ordinalNet()] uses the argument `standardize = TRUE` to center and scale the data. 
+
+## Case weights
+
+```{r}
+#| child: template-no-case-weights.Rmd
```
topepo: There isn't a `weights` argument to `ordinalNet::ordinalNet()`. ordered's model definition uses

```r
protect = c("x", "y", "weights")
```

and should be 

```r
protect = c("x", "y")
```

man/rmd/ordinal_reg_vglm.Rmd
```diff
@@ -0,0 +1,70 @@
...
+`r descr_models("ordinal_reg", "vglm")`
+
+## Tuning Parameters
+
+```{r vglm-param-info, echo=FALSE}
+defaults <- 
+  tibble::tibble(
+    parsnip = c("ordinal_link", "odds_link"),
```
topepo, line 11: Can you verify that these are the defaults?

topepo (changes requested):
This looks great. There's _a lot_ here, and you did a great job on working through it. Afterwards, maybe start an issue in parsnip and give us feedback and suggestions about what could make it easier (e.g., update the checklist). 

A few things that should be changed in ordered: 

 - no case weights for ordinalNet
 - no submodels for ordinalNet

For this PR, tell me what you think about: 

 - using `translate()` to avoid a wrapper function
 - move `vgam` to `gen_add_model()`

Other than that, I think that we are in good shape. 

One question unrelated to this PR: can you list any other parameters for these models that users might want to optimize? For those, we can add more dials parameter functions to make that process smoother. 

parsnip depends on devel dials, so we would probably want to make those dials parameters (if any) pretty soon.

R/ordinal_reg.R
corybrunson: This was intentional, but provisional. The issue is that a single penalty value is ignored regardless, and i think this change only prevents the user from being alerted to that. Yet, if `penalty` is not passed then `xy_xy()` complains. So my intention was to allow the user to silence both by passing `penalty = NULL`.

I wrote a few scripts to experiment with the engines but haven't been tracking them. I'll start tracking them in an R-ignored folder in the `ordered` repo. Here's an abridged version of the one for `ordinalNet()`:

```r
library(tidymodels); library(ordered)

# disaggregated data & partition
house_data <-
  MASS::housing[rep(seq(nrow(MASS::housing)), MASS::housing$Freq), -5]
house_split <- initial_split(house_data, prop = .8)
house_train <- training(house_split)
house_test <- testing(house_split)

# tunable model & analysis specification
# NB: The recipe must include categorical variable encoding because the tuners
# rely on `fit_xy()` rather than on `fit()` and therefore do not call
# `convert_form_to_xy_fit()`.
# house_rec <- recipe(Sat ~ Infl + Type + Cont, data = house_train)
house_rec <- recipe(Sat ~ Infl + Type + Cont, data = house_train) |>
  step_dummy(all_factor_predictors(), one_hot = TRUE)
house_spec <- ordinal_reg() |>
  set_engine("ordinalNet") |>
  set_args(ordinal_link = tune(), penalty = tune(), mixture = tune())

# prepped recipe
house_prep <- prep(house_rec)
# baked training data
house_train_baked <- bake(house_prep, new_data = house_train)

# fitted model
house_spec |>
  set_args(ordinal_link = "logistic", penalty = .001, mixture = .5) |>
  # fit(house_prep, data = house_train) |>
  # fit(formula(house_prep), data = house_train) |>
  fit(formula(house_prep), data = house_train_baked) |>
  print() -> house_fit
# prediction
predict(house_fit, new_data = head(house_train_baked), type = "class")
predict(house_fit, new_data = head(house_train_baked), type = "prob")
# prediction with specified penalty
house_fit |>
  predict(new_data = head(house_train_baked), type = "prob", penalty = .001)
house_fit |>
  predict(new_data = head(house_train_baked), type = "class", penalty = .01)
# prediction outside specified range
house_fit$fit$lambdaVals |> range()
house_fit |>
  predict(new_data = head(house_train_baked), type = "prob", penalty = 0)
house_fit |>
  predict(new_data = head(house_train_baked), type = "class", penalty = 1)

# multiple prediction
house_fit |>
  multi_predict(new_data = head(house_train_baked), type = "prob",
                penalty = c(.1, .01, .001))
house_fit |>
  multi_predict(new_data = head(house_train_baked), type = "class",
                penalty = c(.1, .01, .001))

# multiple model fitting
house_spec |>
  set_args(
    ordinal_link = "logistic",
    path_values = c(.01, .001),
    penalty = 1,
    mixture = .5
  ) |>
  # translate()
  fit(formula(house_prep), data = house_train_baked) |>
  print() -> house_multi_fit
# WARNING: `penalty` must be specified even though not used.
house_spec |>
  set_args(
    ordinal_link = "logistic",
    path_values = c(.01, .001),
    penalty = NULL,
    mixture = .5
  ) |>
  # translate()
  fit(formula(house_prep), data = house_train_baked) |>
  print() -> house_multi_fit
```

R/ordinal_reg.R
corybrunson: Based in part on my own early frustrations learning tidymodels, i think it is important to help the user identify and fix this for any model-level (versus engine-level) or dialable parameter, since the user may throw several engines into a workflow that tunes it.

R/ordinal_reg.R
corybrunson: Oh, i need to double-check that the current code is consistently taking one of the two approaches described here; i might have let them diverge.

corybrunson: `has_submodel = TRUE` was deliberate in order to enable `multi_predict()` for multiple penalty values, as demoed in [a tinkering script](https://github.com/corybrunson/ordered/blob/66f0549227bc2a97f1c505b3b0988906b79b3f23/ignore/tinker-ordinalNet.R#L104-L110). Am i right that `TRUE` is necessary to enable this? or am i wrong that this accomplishes something other than iterating single-fit over the penalty values?

R/ordinal_reg.R
corybrunson: OK, i'll try this in a subbranch and see if anything breaks.

man/rmd/ordinal_reg_ordinalNet.Rmd
corybrunson: There isn't a `weights` argument, but `ordinalNet()` can handle weights when `y` is provided as a matrix. Since this isn't a {parsnip} option, [the wrapper does this](https://github.com/corybrunson/ordered/blob/66f0549227bc2a97f1c505b3b0988906b79b3f23/R/ordinal_reg-ordinalNet.R#L93-L99) before evaluation. Is this an acceptable trick?

R/ordinal_reg_vgam.R
corybrunson: This makes sense, but i would need to figure out how to translate `select_features` and `adjust_deg_free` for `VGAM::vgam()`, as i'm not fluent in these models. Is it clear to you from `?VGAM::vgam`? The first paragraph of "Details" links to the smoothers.

R/ordinal_reg.R
corybrunson: Actually, i'd like to pause this experiment until [the issue of weights](https://github.com/tidymodels/parsnip/pull/1346#discussion_r2900277367) is resolved, just to prevent an unnecessary time sink.

man/rmd/ordinal_reg_vglm.Rmd
corybrunson: They are the defaults [in {ordered}](https://github.com/corybrunson/ordered/blob/main/R/ordinal_reg-VGAM.R) and [in {dials}](https://github.com/tidymodels/dials/blob/main/R/param_ordinal_link.R) (up to translation), but `VGAM::vglm()` requires the user to specify these options explicitly, with the default `family = stop("argument 'family' needs to be assigned")`.

R/ordinal_reg.R
topepo: see `multi_predict()` notes below

R/ordinal_reg.R
topepo: Yes setting this means that you are allowed to make multiple predictions (across submodels) at once, but that's about all that it does. 

I think that we can get parsnip to work well with submodels but I don't know that it would be create for tuning. We'd have to make sure that the values in `path_values` aligns with the grid points and that would be pretty fragile. 

One thing that we could do is to implement our own interpolation  functions for predictions (and coefficients) so that we don't have to worry about it. ordinalNet isn't on CRAN was was updated last month. I/we could contact the authors to see if they were open to a change and, if not, we can implement it and put it ordered or parsnip. 

How does that sound?

man/rmd/ordinal_reg_ordinalNet.Rmd
topepo: Looks good. Nice!

R/ordinal_reg_vgam.R
topepo: I don't see a ~~`vglm()`~~ `vgam()` option analogous to `select_features` (aka `select` in mgcv) or `adjust_deg_free ` (aka `gamma`). That's ok.

R/ordinal_reg.R
topepo: I'm sorry, I didn't see that you already to this in ordered. I'll take a more substantial review tomorrow.

R/ordinal_reg_vgam.R
corybrunson: To be sure, did you check `vgam()` or `vglm()`?

R/ordinal_reg.R
corybrunson: Copy! I left a lot of `REVIEW:` tags there and i'll try to respond to feedback swiftly.

R/ordinal_reg_vgam.R
topepo: Sorry, that was a typo.

R/ordinal_reg.R
topepo: I've tested it a lot and think that were are good with `multi_predict()`.

R/ordinal_reg.R
corybrunson: To be sure, is the plan now to keep the wrapper rather than attempt to build everything into `translate()`?

R/tunable.R
```diff
@@ -333,4 +401,18 @@ tunable.survival_reg <- function(x, ...) {
   res
 }
 
+#' @export
+tunable.ordinal_reg <- function(x, ...) {
```
corybrunson: @topepo i verified these lines a while back but before recent changes. Do they look basically sound?

topepo: We made a change in how this is handled a week or so ago in tidymodels/parsnip#1350. You can now manage that in ordered and it is much more simple and direct. 

~~I'll leave as-is for a bit while I make an ordered PR for these values. Once that is merged, I'll undo that is here.~~

They need to come out of parnsip, then I can add them to ordered.

topepo: Changes for `tunable()` in corybrunson/ordered#17

corybrunson: I believe [PR 17 at {ordered}](https://github.com/corybrunson/ordered/pull/17/) has complete source code now, though rendering may be incomplete and i'm not proficient enough with CI to diagnose the workflow errors. Is this convo resolved?

man/rmd/ordinal_reg_vglm.Rmd
corybrunson: I gave this a try and have been unable to render the details documents locally. (So the new template i added has not been tested; it's a suggestion.)

I also encounter an unexpected error in the `vgam` details document. See [this {parsnip} branch](https://github.com/corybrunson/parsnip/tree/ordered-vgam) and [this {ordered} branch](https://github.com/corybrunson/ordered/tree/main-vgam). Searching the error led me to [this issue](https://github.com/tidymodels/workflows/issues/34) but i did not find a solution there.

corybrunson: Opened [this `parsnip` PR](https://github.com/corybrunson/parsnip/pull/1) with more details and [this companion `ordered` PR](https://github.com/corybrunson/ordered/pull/22) to ease comparison.

topepo:
🎉

PR #1347, opened 2026-03-05
more tunable tests

topepo:
Before refactoring how we augment and update `tunable()`, we should make additional tests with more explicit expectations.

### DIFF
--- a/tests/testthat/test-bart-tunable.R
+++ b/tests/testthat/test-bart-tunable.R
@@ -1,7 +1,7 @@
 test_that("tunable parameters for bart + dbarts", {
-  expect_snapshot({
+  expect_snapshot(
     display_tunable_call_info(
       bart() |> set_engine("dbarts")
     )
-  })
+  )
 })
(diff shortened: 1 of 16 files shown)


tests/testthat/helper-tunable.R
```diff
@@ -0,0 +1,49 @@
+# Helper function for displaying tunable call_info in snapshot tests
+# This expands and displays all call_info elements in a format suitable for
+# snapshot testing, capturing pkg, fun, range, and values.
+
+display_tunable_call_info <- function(model_spec) {
```
EmilHvitfeldt, line 5: Have you looked into trying to use `glue()` instead of `paste()` here?

tests/testthat/test-bart-tunable.R
```diff
@@ -0,0 +1,7 @@
+test_that("tunable parameters for bart + dbarts", {
+  expect_snapshot({
```
EmilHvitfeldt: I'm not quite sure that we need the `{` here since it is only a single expression inside

EmilHvitfeldt (approved):
Looking nice!

what is your reasoning for adding new test files `test-model_name-tunable.R` instead of adding these tests to `test-model_name-engine.R`?

PR #1372, opened 2026-05-13
Optionally add censoring weights to `predict()` output for `type = "survival"`

hfrick:
closes #1371

The CI failures are all https://github.com/tidymodels/parsnip/issues/1375 so I'm going to move this from Draft to For Review.

The tests for this are running in https://github.com/tidymodels/censored/pull/369


R/fit.R
```diff
@@ -281,7 +281,11 @@ fit_xy.model_spec <-
         cli::cli_warn("Engine set to {.val {object$engine}}.")
       }
     }
-    y_var <- colnames(y)
+    if (inherits(y, "Surv")) {
+      y_var = character(0)
```
topepo:
```suggestion
      y_var <- character(0)
```

R/predict.R
```diff
@@ -553,6 +563,15 @@ check_pred_type_dots <- function(
     )
   }
 
+  # `add_censoring_weights` only applies to survival predictions
+  if (any(nms == "add_censoring_weights") & type != "survival") {
+    cli::cli_abort(
+      "{.arg add_censoring_weights} should only be passed to {.fn predict} \\
```
topepo:
```suggestion
      "{.arg add_censoring_weights} should only be passed to {.fn predict}
```

I don't think that we need these, unless they are required for call threading: 

``` r
cli::cli_abort(
  "{.arg add_censoring_weights} should only be passed to {.fn predict} \\
       when {.arg type} is {.val survival}."
)
#> Error:
#> ! `add_censoring_weights` should only be passed to `predict()` when
#>   `type` is "survival".

cli::cli_abort(
  "{.arg add_censoring_weights} should only be passed to {.fn predict}
       when {.arg type} is {.val survival}."
)
#> Error:
#> ! `add_censoring_weights` should only be passed to `predict()` when
#>   `type` is "survival".
```

<sup>Created on 2026-05-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

``` r
options(width = 120)

cli::cli_abort(
  "{.arg add_censoring_weights} should only be passed to {.fn predict} \\
       when {.arg type} is {.val survival}."
)
#> Error:
#> ! `add_censoring_weights` should only be passed to `predict()` when `type` is "survival".

cli::cli_abort(
  "{.arg add_censoring_weights} should only be passed to {.fn predict}
       when {.arg type} is {.val survival}."
)
#> Error:
#> ! `add_censoring_weights` should only be passed to `predict()` when `type` is "survival".
```

<sup>Created on 2026-05-21 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

R/survival-censoring-weights.R
```diff
@@ -268,3 +268,44 @@ add_graf_weights_vec <- function(
...
+    hint <- "Add a {.cls Surv} column, or include the
+     {cli::qty(y_var)}variable{?s} used in the model formula ({.val {y_var}})."
+  } else {
+    hint <- "Add a {.cls Surv} column to {.arg new_data}."
+  }
+  cli::cli_abort(
+    c(
+      "{.code add_censoring_weights = TRUE} requires the survival outcome in 
+      {.arg new_data}.",
+      i = hint
```
topepo, line 307: nice!

R/predict.R
hfrick: Happy to see it go, that was just pattern matching 🧹

PR #1377, opened 2026-05-22
Fix `y_var` coming from `fit_xy()`

### LINKED ISSUE #1375: failing xgboost prediction with sparse data
First seen in #1372 and reproduced locally

```
══ Failed tests ════════════════════════════════════════════════════════════════
── Failure ('test-sparsevctrs.R:240:3'): sparse data work with xgboost engine ──
Expected `predict(xgb_fit, hotel_data)` not to throw any errors.
Actually got a <simpleError> with message:
  [15:41:15] src/learner.cc:1289: Check failed: learner_model_param_.num_feature >= p_fmat->Info().num_col_ (509 vs. 510) : Number of columns does not match number of features in booster.
  Stack trace:
    [bt] (0) 1   xgboost.so                          0x000000017cc29e3c dmlc::LogMessageFatal::~LogMessageFatal() + 124
    [bt] (1) 2   xgboost.so                          0x000000017ceb9040 xgboost::LearnerImpl::ValidateDMatrix(xgboost::DMatrix*, bool) const + 448
    [bt] (2) 3   xgboost.so                          0x000000017ceb9290 xgboost::LearnerImpl::PredictRaw(xgboost::DMatrix*, xgboost::PredictionCacheEntry*, bool, unsigned int, unsigned int) const + 80
    [bt] (3) 4   xgboost.so                          0x000000017cea9e38 xgboost::LearnerImpl::Predict(std::__1::shared_ptr<xgboost::DMatrix>, bool, xgboost::HostDeviceVector<float>*, int, int, bool, bool, bool, bool, bool) + 632
    [bt] (4) 5   xgboost.so                          0x000000017cf89984 XGBoosterPredictFromDMatrix + 836
    [bt] 

[ FAIL 1 | WARN 0 | SKIP 22 | PASS 1558 ]
```

hfrick:
closes #1375 -- it was due to a change in #1372

The first commit has the breaking change from #1372 in it and we see tests breaking in same way and captured in #1375.

The second commit fixes it. Once this is on `main` I'll update #1372 with the change.

Long story short of what happened:
- I noticed that when we captured `y_var` from the matrix interface in `fit_xy()` in https://github.com/tidymodels/parsnip/pull/961, we only passed it through `xy_form()` but not `xy_xy()` and I fixed that when working on the survival PR. 
- Some engines with a matrix interface are strict in that they only want the predictors given to them for prediction - and xgboost happens to be one of them. That got fixed in https://github.com/tidymodels/parsnip/pull/1168 but it relied on the missing `y_var` in the xy-xy path as the path signal.
- So when I evened out the path, it broke the path signal. 
- I now let it rely on the `x_names` which were added for this purpose.
- I do think we should capture the path between interfaces in a flag instead of relying on the absence or presence of other things but as this whole story illustrates yet again, that should be its separate PR.

### DIFF
--- a/R/fit_helpers.R
+++ b/R/fit_helpers.R
@@ -119,10 +119,14 @@ xy_xy <- function(
   )
   elapsed <- proc.time() - time
 
-  if (is.atomic(env$y)) {
-    y_name <- character(0)
+  if (!is.null(env$y_var)) {
+    y_name <- env$y_var
   } else {
-    y_name <- colnames(env$y)
+    if (is.atomic(env$y)) {
+      y_name <- character(0)
+    } else {
+      y_name <- colnames(env$y)
+    }
   }
   res$preproc <- list(y_var = y_name, x_names = colnames(env$x))
   res$elapsed <- list(elapsed = elapsed, print = control$verbosity > 1L)
--- a/R/predict.R
+++ b/R/predict.R
@@ -569,8 +569,7 @@ prepare_data <- function(object, new_data) {
   translate_from_xy_to_formula <- any(preproc_names == "x_var", na.rm = TRUE)
   # For backwards compatibility, only do this if `y_var` is missing and
   # `x_names` is present
-  translate_from_xy_to_xy <- any(preproc_names == "x_names", na.rm = TRUE) &&
-    identical(object$preproc$y_var, character(0))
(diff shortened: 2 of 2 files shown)


EmilHvitfeldt (approved):
Great find!

Issue #25, opened 2018-06-22
test abnormal class values


topepo:
These become class names eventually so test back-ticks and maybe standardize on tibbles

topepo:
This works (esp for factor levels "0" and "1" 🙄). Back-ticks happen automatically.

Issue #21, opened 2018-06-22
data descriptors for spark data objects


topepo:
This will need to be different code for spark objects that should emulate `get_descr_form` (since we've constrained `spark` objects to the formula method).

PR #1384, opened 2026-06-22
prepare engine list and translation steps for new ordinal regression engines in ordered

corybrunson:
This PR follows #1346 and continues #1298.

As an experiment with OpenCode, i've been preparing source code, documentation, and unit tests in {ordered} for several new engines; see [this companion branch](https://github.com/corybrunson/ordered/tree/eng-opencode). (I've reviewed all generated code and modified most of it, but there may still be artifacts neither i nor the checks noticed. Commits prefixed `w/opencode-` resulted from a few iterations of generation plus manual modifications; commits without this prefix comprise only manual changes.)

This PR begins to prepare parsnip for these new engines. It is not complete; remaining tasks are mentioned in the sections.

## model directory

All new engines are listed in 'R/aaa_archive.R'

## `translate.ordinal_reg()`

Two new engines require special handling of arguments, which i've shunted to `translate.ordinal_reg()` in order to avoid the burden of additional wrappers:

1. `ordinal::clm()`, like `VGAM::vglm()` and `ordinalNet::ordinalNet()`, recognizes `"logit"` rather than `"logistic"` as an ordinal link option. Whereas those two require wrappers for other reasons, this one can get by with just the translation step.
2. `glmnetcr::glmnetcr()`, like `ordinalNet::ordinalNet()`, accepts (and prefers) a penalty path rather than a single penalty value. With the exception of the option to include `0`, it behaves analogously to `ordinalNet::ordinalNet()`. The translation steps are therefore adapted from the existing steps for that engine.

There are two notable differences between the current {ordered} prediction methods for `ordinalNet` and `glmnetcr` models:
* If the user provides a penalty value outside the path range, then `predict_glmnetcr_wrapper()` uses the nearer endpoint of the path rather than throwing an error as `predict_ordinalNet_wrapper()` does.
* `multi_predict._glmnetcr()` aborts when passed `type = "raw"`, whereas `multi_predict._ordinalNet()` handles this case.

Some additional tests of these behaviors are added to 'test-translate.R'.

Any desired changes might have implications for the `translate()` method, so they should be completed before this PR is merged.

## files awaiting changes

'inst/models.tsv' is not yet updated because i don't have all extensions installed locally. Engine documentation files ('R/ordinal_reg_*.R', 'man/rmd/ordinal_reg_*.Rmd') and 'NEWS.md' items are not yet written because i want to be sure that the PR is otherwise complete before writing them.

## next steps

Following this PR, i'll introduce the `clm`, `lrm`, `orm`, `glmnetcr`, and `orf` engines. The new engines set parsnip and ordered up for two additional coordinated enhancements:

* Enable the `"linear_pred"` prediction type for `ordinal_reg` and `gen_additive_mod` models in parsnip and register this prediction type for the `ordinalNet`, `vglm`, `clm`, `lrm`, `orm`, and `vgam` engines (see [`ordered` issue #21](https://github.com/corybrunson/ordered/issues/21).
* Either introduce a new `ordinal_reg` model argument in parsnip called `parallel_reg` to control the parallel regression assumption in `ordinalNet`, `clm`, and `vglm` engines or expose the engine arguments for these features directly in ordered (see [`ordered` issue #25](https://github.com/corybrunson/ordered/issues/25).

Because they must be coordinated, i only expect to resubmit to CRAN once all three are completed (and, of course, following resubmission of parsnip). If preferable, both can be absorbed into this PR.

Thank you all!


topepo:
PR looks good. 

> Following this PR, i'll introduce the `clm`, `lrm`, `orm`, `glmnetcr`, and `orf` engines. The new engines set parsnip and ordered up for two additional coordinated enhancements:
> 
> * Enable the `"linear_pred"` prediction type for `ordinal_reg` and `gen_additive_mod` models in parsnip and register this prediction type for the `ordinalNet`, `vglm`, `clm`, `lrm`, `orm`, and `vgam` engines (see [`ordered` issue #21](https://github.com/corybrunson/ordered/issues/21).
> * Either introduce a new `ordinal_reg` model argument in parsnip called `parallel_reg` to control the parallel regression assumption in `ordinalNet`, `clm`, and `vglm` engines or expose the engine arguments for these features directly in ordered (see [`ordered` issue #25](https://github.com/corybrunson/ordered/issues/25).

☝️ A parameter for parallel/non-parallel is a great idea. 

Just to make sure: the changes above are going into ordered and not parsnip? They should be in ordered IMO. 

> Because they must be coordinated, i only expect to resubmit to CRAN once all three are completed (and, of course, following resubmission of parsnip). If preferable, both can be absorbed into this PR.

I'd do more PRs. You cna send in more than one at a time if that helps. 

We are going to do a parsnip release in about two weeks, mostly to add some new model types (for a new package of ours not yet on CRAN). It would be good to add any new engine docs before that. We are *not* on a tight schedule, and we might miss the CRAN closure (Aug 5-19), but that's not a huge problem (for us, at least). 

So submit PRs as much as you like. I'll be out next week but I'll make them a priority when I get back.

corybrunson:
@topepo thank you!

> Just to make sure: the changes above are going into ordered and not parsnip? They should be in ordered IMO.

I believe both next steps require (or at least call for) small changes to parsnip, despite the bulk being done in ordered:

* `"linear_pred"` is enabled in censored by adding it to the `pred_types` object copied from parsnip. With two extensions using it, it makes sense to me to add it to `pred_types` in parsnip. (Though i just looked and it's already there—was it only added recently? So this wouldn't involve any change to parsnip.) EDIT: Sorry, the needed change is actually to `check_pred_type()`; see [these lines](https://github.com/tidymodels/parsnip/blob/262699756ad4d95ba824e3133aafe0d681e478fb/R/predict.R#L279-L284).
* If it makes sense for all three of these `ordinal_reg` engines to use the same `parallel_reg` parameter, then my thought was that it should become a model argument rather than engine arguments.. But i'm not insistent on that.

> It would be good to add any new engine docs before that.

I want to spend some time understanding all the parallel regression options provided by the three packages. I'll try to put in that time the week after next. But if you think they should be engine arguments then parsnip won't have to wait for any more PRs.

PR #1386, opened 2026-07-02
tabby models

topepo:
Model definitions, tests, and engine documentation for models in [tabby](https://github.com/tidymodels/tabby). Once this is merged, I will PR there to remove the bits added here.


R/tabular_auto_int.R
```diff
@@ -0,0 +1,279 @@
...
+    mode = "unknown",
+    engine = "brulee",
+    epochs = NULL,
+    num_embedding = NULL,
+    hidden_units = NULL,
+    hidden_activations = NULL,
+    num_attn_feat = NULL,
+    num_attn_heads = NULL,
+    num_attn_blocks = NULL,
+    activation = NULL,
```
hfrick, line 62: Since this model has `activation` and `hidden_activations`:

`activation` gets its documentation via `@inheritParams` and starts with
> A single character string denoting the type of relationship between the original predictors and the hidden unit layer.

For `tabular_resnet(), that first sentence has been overwritten to read
> A vector character strings denoting the type of relationship between the layers.

Should either of them get changed?

R/tabular_auto_int.R
```diff
@@ -0,0 +1,279 @@
...
+    dropout_attn = NULL,
+    dropout_embedding = NULL,
+    penalty = NULL,
+    mixture = NULL,
+    learn_rate = NULL,
+    rate_schedule = NULL,
+    momentum = NULL,
+    batch_size = NULL,
+    class_weights = NULL,
+    stop_iter = NULL
```
hfrick, line 73: This and other models with brulee engines get a whole lot more main arguments than `mlp()` which only has them as engine arguments. I assume that's very intentional but wanted to make sure?

R/tabular_auto_int.R
```diff
@@ -0,0 +1,279 @@
...
+    is.numeric(args$penalty) &&
+      is.numeric(args$dropout) &&
+      args$dropout > 0 &&
+      args$penalty > 0
+  ) {
+    cli::cli_abort(
+      "Both weight decay and dropout should not be specified.",
+      call = call
+    )
+  }
```
hfrick: This check is repeated a couple of times and could/should be extracted into a custom check function.

It comes up for `tabular_resnet()`, `tabular_auto_int()`, `tabular_saint()`, and `mlp()`.

I would also change the wording to "Weight decay and dropout should not be both specified" because the current sentence reads to me as if neither should be specified.

R/tabular_rln.R
```diff
@@ -0,0 +1,180 @@
...
+#' @export
+check_args.tabular_rln <- function(object, call = rlang::caller_env()) {
+  args <- lapply(object$args, rlang::eval_tidy)
+
+  if (!is.null(args$penalty_type) && !args$penalty_type %in% c("L1", "L2")) {
+    cli::cli_abort(
+      "{.arg penalty_type} must be {.val L1} or {.val L2}, not {.val {args$penalty_type}}.",
+      call = call
+    )
+  }
```
hfrick: I would use `rlang::arg_match()` here for the standard "must be one of … did you mean …?" message.

R/tabular_saint.R
```diff
@@ -0,0 +1,306 @@
...
+
+  if (
+    !is.null(args$attention_type) &&
+      !args$attention_type %in% c("column", "row", "both")
+  ) {
+    cli::cli_abort(
+      "{.arg attention_type} must be one of {.val column}, {.val row}, or {.val both}.",
+      call = call
+    )
+  }
```
hfrick: same comment about `rlang::arg_match()`

R/tabular_icl.R
```diff
@@ -0,0 +1,133 @@
...
+  check_number_decimal(
+    args$softmax_temperature,
+    min = 0,
+    allow_null = TRUE,
+    call = call,
+    arg = "softmax_temperature"
+  )
+  check_number_whole(
+    args$num_estimators,
+    min = 0,
```
hfrick: `min` in inclusive, so this allows an ensemble size of 0. Is this correct?

R/tabular_pfn.R
```diff
@@ -0,0 +1,165 @@
...
+  check_number_decimal(
+    args$softmax_temperature,
+    min = 0,
+    allow_null = TRUE,
+    call = call,
+    arg = "softmax_temperature"
+  )
+  check_number_whole(
+    args$num_estimators,
+    min = 0,
```
hfrick: same question about the required minimum size of the ensemble

R/tabular_resnet.R
```diff
@@ -0,0 +1,223 @@
...
+      user_specified_mode = !missing(mode),
+      method = NULL,
+      engine = engine,
+      user_specified_engine = !missing(engine)
+    )
+  }
+
+# ------------------------------------------------------------------------------
+
+#' Updating a model specification
```
hfrick:
```suggestion
```

we can rely on `@rdname parsnip_update`

hfrick (approved):
Looks good! Just some smaller things in the comments

R/tabular_auto_int.R
topepo: It is, and in hindsight, I should have expanded the main arguments for `mlp()`. All of these new ones only have one engine (unlike `mlp()`) so that's a pretty easy decision to make.

R/tabular_icl.R
topepo: Nope!

R/tabular_auto_int.R
topepo: It's a good point but the arguments are not the same across these 6 models. It would increase complexity to have conditionals for each configuration, so I'll leave as-is for now. 

I'm going to remove that check for both penalization and dropout; I've found a case or two where they might both help.

PR #1390, opened 2026-07-16
changes to decrease inefficiency between parsnip and engine function

topepo:
See #1071 (overhead test), follows up on #1369, #1073, #1075

Mostly on GHA, we've been seeing these unit test failureS: 

```
── Failure ('test-fit_interfaces.R:192:3'): overhead of parsnip interface is minimal (#1071) ──
Expected parsnip overhead factor (xy interface): 3.8247 to be TRUE.
Differences:
`actual`:   FALSE
`expected`: TRUE 
```

Assuming that we've made changes to increase inefficiency, this PR is trying to mitigate the issues. 

### Registry and helper functions

- `get_model_env()` returns the registry environment directly instead of calling `utils::getFromNamespace("parsnip", ns = "parsnip")` on every access (`R/aaa_models.R`).
- `check_spec_mode_engine_val()` fast-paths the success case by counting matching rows instead of materializing a `vctrs::vec_slice()` of the model table. All diagnostic/error branches are unchanged (`R/aaa_models.R`).
- `get_encoding()` looks the table up directly and only runs `check_model_exists()` plus the legacy dplyr fallback when the lookup returns `NULL`, instead of wrapping every call in `try()` (`R/aaa_models.R`).
- `get_args()` builds its result column-wise with `tibble::new_tibble()` instead of `vec_slice()` on the tibble followed by `args$engine <- NULL` (which dispatches `$<-.tbl_df`) (`R/translate.R`).
- `translate.default()` skips `check_spec_mode_engine_val()` when `x$method` is already populated, since `add_methods()` has just validated the same combination during `fit()`. Direct user calls on a bare spec still validate. Also `purrr::map_lgl()` → `vapply()` (`R/translate.R`).
- `possible_engines()` uses `[[` instead of `rlang::env_get()` (`R/engines.R`).
- `is_installed()` caches successful `requireNamespace()` results in a package-local environment (`pkg_install_cache`). Failures are not cached so installing a missing package mid-session takes effect without restarting R. `check_installs()` uses `vapply()` (`R/engines.R`).
- `is_sparse_matrix()` short-circuits with `isS4()` before `methods::is()`; all Matrix sparse classes are S4, so behavior is identical and the common data-frame case is ~100x faster (`R/sparsevctrs.R`).
- `condense_control()` returns early when the input already has the reference's class and names. `fit()`/`fit_xy()` pass a prebuilt `default_parsnip_control` constant as the reference instead of constructing `control_parsnip()` per fit (`R/condense_control.R`, `R/control_parsnip.R`, `R/fit.R`).

### Structural cleanups

- Removed `cl <- match.call(expand.dots = TRUE)` from `fit.model_spec()` and `fit_xy.model_spec()` and dropped the unused `cl` parameter from `check_interface()` and `check_xy_interface()`; updated the two tester functions in `tests/testthat/test-fit_interfaces.R` accordingly (`R/fit.R`).
- Removed the dead `dots <- quos(...)` in `fit_xy.model_spec()`; the one in `fit.model_spec()` is replaced with `...names()` (R >= 4.1) (`R/fit.R`).
- In all four fit wrappers (`form_form()`, `xy_xy()`, `form_xy()`, `xy_form()`), the encoding lookup uses direct column indexing with an `is_spec_encoding` logical selector instead of `dplyr::filter()` or `vec_slice()` on the tibble (`R/fit_helpers.R`).

### Other

- `NEWS.md` bullet under the development version.
- Incidental: testthat pruned a stale snapshot from `tests/testthat/_snaps/mlp.md` ("Both weight decay and dropout should not be specified.") whose generating test no longer exists; it only survived on machines where keras is not installed and the test skips.

## Compatibility notes

- Exported helpers (`get_model_env()`, `get_from_env()`, `get_encoding()`, `condense_control()`) keep their signatures and return types, so extension packages (censored, bonsai, etc.) are unaffected.
- `check_interface()`/`check_xy_interface()` are internal; the dropped `cl` argument was never used.
- The test thresholds (3.5/3.75) are unchanged.

## Verification

- `devtools::test()`: 1772 pass, 0 fail, 6 pre-existing environment skips (spark/keras).
- Coefficients from `fit(linear_reg(), mpg ~ ., mtcars)` and `fit_xy()` match `lm()` exactly; ranger fits exercised the `check_installs()` cache and the xy/formula conversion paths.
- Overhead benchmark re-run three times with stable results (see table above).
- `air format .` applied; `jarl check` reports no new findings in the modified files (19 pre-existing findings verified identical against the pre-change baseline).
- `R CMD check` (not as CRAN): 0 errors, 0 warnings, 0 notes.

## Deliberately not included (possible follow-ups)

- `.convert_xy_to_form_fit()` still costs ~63µs on the xy-to-formula path; it was optimized in #1369 and further changes are riskier.
- `make_form_call()` (~17µs) and memoizing `translate()` output across tuning iterations are larger design changes and were not needed to restore margin.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -24,3 +24,4 @@ derby.log
 ^\.vscode$
 ^[.]?air[.]toml$
 ^\.claude$
+^plans$
--- a/R/engines.R
+++ b/R/engines.R
@@ -5,8 +5,8 @@ specific_model <- function(x) {
 
 possible_engines <- function(object, ...) {
   m_env <- get_model_env()
-  engs <- rlang::env_get(m_env, specific_model(object))
-  unique(engs$engine)
+  engines <- m_env[[specific_model(object)]]
+  unique(engines$engine)
 }
 
 # ------------------------------------------------------------------------------
@@ -15,14 +15,26 @@ shhhh <- function(x) {
   suppressPackageStartupMessages(requireNamespace(x, quietly = TRUE))
 }
 
+# An installed package stays installed for the rest of the session, so
+# remember successful checks to avoid repeated `requireNamespace()` calls on
+# every fit. Failures are not cached so that installing a missing package
+# takes effect without restarting R.
+pkg_install_cache <- new.env(parent = emptyenv())
+
 is_installed <- function(pkg) {
-  res <- try(shhhh(pkg), silent = TRUE)
-  res
+  if (isTRUE(pkg_install_cache[[pkg]])) {
+    return(TRUE)
+  }
+  installed <- isTRUE(try(shhhh(pkg), silent = TRUE))
+  if (installed) {
+    pkg_install_cache[[pkg]] <- TRUE
+  }
+  installed
 }
 
 check_installs <- function(x, call = rlang::caller_env()) {
   if (length(x$method$libs) > 0) {
-    is_inst <- map_lgl(x$method$libs, is_installed)
+    is_inst <- vapply(x$method$libs, is_installed, logical(1))
     if (!all(is_inst)) {
       missing_pkg <- x$method$libs[!is_inst]
       missing_pkg <- paste0(missing_pkg, collapse = ", ")
--- a/R/fit_helpers.R
+++ b/R/fit_helpers.R
@@ -9,14 +9,11 @@ form_form <-
       check_outcome(eval_tidy(rlang::f_lhs(env$formula), env$data), object)
 
       encoding_info <- get_encoding(class(object)[1])
-      encoding_info <-
-        vctrs::vec_slice(
-          encoding_info,
-          encoding_info$mode == object$mode &
-            encoding_info$engine == object$engine
-        )
+      is_spec_encoding <-
+        encoding_info$mode == object$mode &
+        encoding_info$engine == object$engine
 
-      remove_intercept <- encoding_info$remove_intercept
+      remove_intercept <- encoding_info$remove_intercept[is_spec_encoding]
       if (remove_intercept) {
         env$data <- env$data[,
           colnames(env$data) != "(Intercept)",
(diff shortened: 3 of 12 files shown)


R/engines.R
```diff
@@ -15,14 +15,26 @@ shhhh <- function(x) {
...
+  if (installed) {
+    pkg_install_cache[[pkg]] <- TRUE
+  }
+  installed
 }
 
 check_installs <- function(x, call = rlang::caller_env()) {
   if (length(x$method$libs) > 0) {
-    is_inst <- map_lgl(x$method$libs, is_installed)
+    is_inst <- vapply(x$method$libs, is_installed, logical(1))
```
EmilHvitfeldt, line 37: I'm surpricsed this was worth the change

R/fit_helpers.R
```diff
@@ -84,11 +81,12 @@ xy_xy <- function(
...
-  encoding_info <-
-    get_encoding(class(object)[1]) |>
-    dplyr::filter(mode == object$mode, engine == object$engine)
+  encoding_info <- get_encoding(class(object)[1])
+  is_spec_encoding <-
+    encoding_info$mode == object$mode &
+    encoding_info$engine == object$engine
 
-  remove_intercept <- encoding_info$remove_intercept
+  remove_intercept <- encoding_info$remove_intercept[is_spec_encoding]
```
hfrick, line 91: Given that this undoes an instance of "dplyr::filter() creeping back in", I'd consider adding a code comment that this way of filtering is intentionally written in the base-R way for speed.

hfrick: Same for other occurences of this pattern

.Rbuildignore
```diff
@@ -24,3 +24,4 @@ derby.log
 ^\.vscode$
 ^[.]?air[.]toml$
 ^\.claude$
+^plans$
```
hfrick, line 27: have conventions from usethis changed? I thought we are sticking plans into `.claude/plans`?

hfrick (approved):
looks good 👍

EmilHvitfeldt (approved):

R/engines.R
topepo: Claude did not report a timing difference here.

topepo:
You can't be too paranoid... 

we checked 100 reverse dependencies (99 from CRAN + 1 from Bioconductor), comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 0 new problems

PR #1391, opened 2026-07-17
allow linear_pred predictions for ordinal regression and generalized additive models

corybrunson:
Currently `parsnip:::check_pred_type()` allows the `"linear_pred"` option only for the censored regression _mode_. This PR additionally allows it for ordinal regression and generalized additive _models_. The change deviates from the pattern of only considering the mode, so caution is warranted, though i haven't encountered a problem locally.

I am also not sure if it is appropriate, and it seems unnecessary, as several `predict._<engine>()` methods in ordered serve to bypass this check. However, the pattern itself may be reason for making the change in parsnip rather than duplication across several engines.

The PR also implements this prediction option for the `mgcv` engine, following the same conventions. This also may or may not be appropriate, but i included it as a proof of concept and companion to the implementation for `vgam`.

'NEWS.md' and other documentation have not been updated yet.

This PR coordinates with [this PR](https://github.com/corybrunson/ordered/pull/28) from the `linear-pred` branch of `ordered`. As mentioned in #1384, this is the second of two or three PRs coordinated with changes to ordered.


topepo (approved):
Looks good. We decided to remove the check for linear predictors

corybrunson:
Cool! Since the new ordered engines won't be accessible until the user installs the parsnip upgrade anyway, i'll go ahead and remove the `linear_pred` workarounds as they won't be needed with the upgrade.

Issue #35, opened 2018-07-30
next set of models


topepo:
- [x] knn via `kknn` package
 - [x] `decision_tree` via `rpart`, `C5.0`, `spark` (others?)
 - [x] SVM models: linear, RBF, polynomial as separate functions (`kernlab`)
 - [x] multinomial regression via `glment` and `spark`
 - [x] mars via `earth` package
 - [x] null model wrapper as well as fit and pred functions
 - [x] naive Bayes (`klaR`, ~~spark~~ ?)
 - [x] cubist
 - [x] discriminant analysis (of various types)
 - [x] PLS (sparse, DA)
 - [x] FDA models with different basis functions (MARS, poly)
 - [x] bagged trees (helped by potential new `rpart` version and side package)
 - [x] bagged MARS (based on side package)
 - [x] Poisson regression (perhaps including ZIP models; otherwise a clone of `linear_reg`)
 - [x] [ARIMA and other time series models](https://github.com/business-science/modeltime)
 - [x] generalized additive models
 - [x] multilevel model extension engines for linear, logistic, multinomial, and Poisson regression (in the [`multilevelmod` package](https://github.com/tidymodels/multilevelmod/))
 - [x] more models for censored data (in the [`censored` package](https://github.com/tidymodels/censored))

👆Already in `parsnip` or [adjacent package](https://www.tidymodels.org/find/parsnip/)
👇working on or thinking about

 - [ ] ordinal regression
 - [ ] rotation forests

billdenney:
I think that this is the right place to ask:

What about:

* nonlinear models?
* linear mixed effect models? (I think that this is sufficiently separate from `linear_regression()`)
* nonlinear mixed effects models?

As a lower priority, what about more complex cases that are combinations of other model types:
* ordinal mixed-effect nonlinear regression models?  (This isn't just making something up-- it's a project that I'm currently working on in Stan.)
* multiple-outcome models that mix modeling types:  last year I was working on a nonlinear mixed-effects regression combined with a nonlinear time to event model.

alexpghayes:
So far `parsnip` mainly supports `fit()`ing and `predict()`ing, but doesn't really provide any infrastructure for inference or inspecting model parameters. I suspect `parsnip` will eventually support multi-level models, but probably at the same `fit()/predict()` level. Perhaps this will cover your use case, but typically I think people want more inference from these models. `lme*` and `brms`, etc, etc, provide a lot of the infrastructure to make this inference easier for users.

[...]

PR #1393, opened 2026-07-18
model arguments for threshold structure and parallel regression assumption

corybrunson:
This PR introduces two model arguments for `ordinal_reg`, proposed and outlined in [this issue at ordered](https://github.com/corybrunson/ordered/issues/25) but deviating from that outline:
* `threshold_structure` is character-valued and applies to the `clm` and `vglm` engines. It controls whether and what constraint is placed on the ordinal thresholds.
* `parallel_reg` can be used by `clm`, `vglm`, and `ordinalNet`. It takes a logical, a formula, or a list of two of these; this was an attempt to give the user full control over each engine's facilities while hewing to their syntaces. General propriety is checked in `check_args()` while engine-specific quirks are handled in `translate()` and in the engine wrappers in ordered.

I'm very open to critical feedback on the many decisions that went into this PR, including whether these arguments should belong to the model rather than to the engines.

One open question is whether and how these arguments should be made available to `vgam`; i don't know if they apply to `mgcv`. Currently they are supported by the `vgam` wrapper.

These model arguments were designed with a brms engine additionally in mind, but their utility/prospects seem uncertain in light of issue #35 and [bayesian](https://github.com/hsbadr/bayesian).

This is the last of the three PRs described in #1384. Coordinated PRs to dials and ordered are ~forthcoming~ [here](https://github.com/tidymodels/dials/pull/462) and [here](https://github.com/corybrunson/ordered/pull/29).


corybrunson:
The most recent binary available for VGAM is 1.1-9; my office computer, for example, can't install more recent versions than that. Unfortunately, this is also the last (and first) version to use the `threshold` parameter, which was replaced in 1.1-10 with `Thresh` (see the [NEWS](https://github.com/cran/VGAM/blob/master/NEWS) there). Maybe it's fine to specify version 1.1-10 in the description; but is there a preferred way to handle model arguments for an engine whose API has changed? I think this could be done using a wrapper that checks the installed version before passing a call, but ordered is already wrapper-heavy.

topepo:
Looking at this, I think it is trying to do too much with the `parallel_reg` argument. 

I (meaning Claude) spent some time trying to understand the underlying APIs that are driving these design decisions. You probably know all of this already, but it is good to document it here (and there are more details at the bottom). 

It looks like the engine-package syntax is very nonuniform in form and function: 

| Engine | Package | Non-parallel? | Specification style |
|---|---|---|---|
| `polr` | MASS | No | — |
| `vglm` | VGAM | **Yes** | Omnibus logical *or* covariate-specific formula |
| `ordinalNet` | ordinalNet | **Yes** | Omnibus flags; semi-parallel model is data-driven per covariate |
| `lrm` | rms | No | — |
| `orm` | rms | No | — |
| `glmnetcr` | glmnetcr | No | — |
| `clm` | ordinal | **Yes** | Covariate-specific only (separate `nominal` formula) |

Some asides

- `ordinalNet` is really keeping to its glmnet roots with bad API design, having arguments `parallelTerms` and `nonparallelTerms`, both of which could have values of `TRUE`. 
- VGAM has an interface of `parallel = TRUE ~ Type + Cont - 1` which is _a choice_, especially if you must remove the intercept 🙄 

Anyway, I think that the main question is "How much functionality in the engines do you want to support in the tidymodels interface?" and how should that happen? In no particular order...

## Option 1: leverage the formula

One idea is to handle it via a formula method. If you constrain these models to only work with the formula method (like we do for GAMs and multilevel models), you can enable them to do things like

```r
# Everything is parallel
y ~ a + b + c
y ~ .

# or make b nonparallel
y ~ a + nonparallel(b) + c

# all non-parallel
y ~ nonparallel(a) + nonparallel(b) + nonparallel(c)
```

and so on. This is handled by a fake `nonparallel()` function, a "special" in old-school S/R terminology. From `?formula`: 

> Some model-fitting functions recognize other “specials” like `strata` and `cluster` (typically by setting the specials argument of [terms.formula](https://stat.ethz.ch/R-manual/R-devel/library/stats/help/terms.formula.html)). Because specials are nothing more than syntax, disambiguation using [::](https://stat.ethz.ch/R-manual/R-devel/library/base/html/ns-dblcolon.html) or similar is nonsense (even if packages happen to define so-named functions) and the result is often a model that the user does not intend. 

We do this for our interface to the [gee](https://multilevelmod.tidymodels.org/articles/multilevelmod.html#generalized-estimator-equations-gee) interface in the multilevelmod package. 

It's expressive but adds more behind-the-scenes work and isn't great for models where the covariates are either all parallel or all non-parallel. 

To tune this, the user would use workflow sets with different formulas. 

## Option 2: all or nothing is parallel

Here, the tuning parameter is a simple boolean and you set all of the covariates to their formats in your wrapper. This is the easiest option. 

Supporting this _and_ the engine-specific API in this PR is up to you, but I would really avoid it 

## Option 3: Enable it but don't make it a formal tuning parameter

This is very similar to the first option. 

You could create a unified interface for your wrappers that takes a formula, letting users decide what should be non-parallel. You could do that with a special (as in the other option) or just a basic formula. Using the example above: 

```r
# everything is parallel (default)
nonparallel = ~ 0

# b is nonparallel
nonparallel = ~ b

# everything is nonparallel (and you remove the intercept)
nonparallel = ~ .
```

In this instance, you can't tune this in the traditional way (i.e., grid search or iterative search). We might be able to enable this in the future by relaxing the constraint that tuning parameter values must be a scalar. 

However, for now, users can tune the model this way by using a workflow set that specifies this interface with `nonparallel` as the engine argument. 

<hr> 

I would probably go with option 2 (but that's me). 

Also, for posterity, here is the rest of claude's summary of these APIs

<details>

Notes on the parallel-only engines:

- `MASS::polr()` is strictly proportional odds with no relaxation option.
- `rms::lrm()` and `rms::orm()` fit proportional odds / cumulative probability models with parallel slopes by construction; partial proportional odds in the rms ecosystem lives in `rmsb::blrm()`, which is not an engine here.
- `glmnetcr::glmnetcr()` fits the *constrained* (i.e., parallel) continuation ratio model; that constraint is the point of the package, and its restructured-data interface has no way to add covariate-by-cutpoint interactions.

## Current wrapper restriction for `vglm`

VGAM itself accepts either a logical or a constraint formula for `parallel`, but `VGAM_vglm_wrapper()` currently calls `check_logical(parallel)` with a "for now, require `parallel` to be logical" comment (`R/ordinal_reg-VGAM.R:62-63`), and the engine default is `parallel = TRUE` (`R/ordinal_reg-data.R:148`).

So through parsnip today only the omnibus form `set_engine("vglm", parallel = FALSE)` is available; formula-based partial parallelism would require relaxing that check.

## Common example

All syntax below was run and verified against the expanded housing data, with a 3-level ordered outcome (`Sat`: Low < Medium < High, so 2 thresholds) and predictors contributing 6 dummy columns.
The covariate freed in the partial (covariate-specific) models is `Infl`.

```r
house_data <- MASS::housing[rep(seq(nrow(MASS::housing)), MASS::housing$Freq), -5]
# model: Sat ~ Infl + Type + Cont
```

Coefficient counts for calibration: the parallel model has 8 coefficients (2 thresholds + 6 slopes), the fully non-parallel model has 14 (2 + 6 × 2), and freeing only `Infl` gives 10 (2 + 2 × 2 + 4).

## Engine-specific syntax (underlying functions, no wrappers)

### `VGAM::vglm` — logical or formula on `parallel`

```r
library(VGAM)

# parallel (proportional odds)                                  # 8 coefs
vglm(Sat ~ Infl + Type + Cont, data = house_data,
     family = cumulative(link = "logitlink", parallel = TRUE))

# fully non-parallel                                            # 14 coefs
vglm(Sat ~ Infl + Type + Cont, data = house_data,
     family = cumulative(link = "logitlink", parallel = FALSE))

# covariate-specific: free only Infl                            # 10 coefs
vglm(Sat ~ Infl + Type + Cont, data = house_data,
     family = cumulative(link = "logitlink", parallel = FALSE ~ Infl))

# same model stated from the other side: constrain Type and Cont
vglm(Sat ~ Infl + Type + Cont, data = house_data,
     family = cumulative(link = "logitlink", parallel = TRUE ~ Type + Cont - 1))
```

The left side of the constraint formula says whether the right-side terms are the parallel ones (`TRUE ~ ...`) or the freed ones (`FALSE ~ ...`); the `- 1` in the `TRUE` form keeps the intercepts (thresholds) out of the constraint.

Both formula forms produce identical 10-coefficient fits, with the freed terms named `InflMedium:1`, `InflMedium:2`, etc.

The same `parallel` argument works in `acat()`, `cratio()`, and `sratio()`.

### `ordinal::clm` — separate `nominal` (and `scale`) formulas

```r
# parallel (proportional odds)                                  # 8 coefs
ordinal::clm(Sat ~ Infl + Type + Cont, data = house_data, link = "logit")

# covariate-specific: free only Infl — move it to `nominal`     # 10 coefs
ordinal::clm(Sat ~ Type + Cont, nominal = ~ Infl, data = house_data)

# fully non-parallel: everything in `nominal`                   # 14 coefs
ordinal::clm(Sat ~ 1, nominal = ~ Infl + Type + Cont, data = house_data)

# location-scale alternative (multiplicative non-proportionality)
ordinal::clm(Sat ~ Infl + Type + Cont, scale = ~ Infl, data = house_data)
```

There is no omnibus switch; non-parallelism is declared covariate-by-covariate by moving predictors into the `nominal` formula.

Freed coefficients come back as `Low|Medium.InflMedium`, `Medium|High.InflMedium`, etc.
Caveat: `clm()` does **not** error when a variable appears in both `formula` and `nominal` — it fits silently and returns the aliased location coefficients as `NA` (verified: `InflMedium` and `InflHigh` come back `NA`).

This quiet aliasing matters in a parsnip context, where a user setting `set_engine("clm", nominal = ~ Infl)` must also keep `Infl` out of the formula or recipe passed to `fit()`, and where `NA` coefficients could confuse downstream code.

The `scale` formula is a different, multiplicative route to non-proportionality (location-scale model), and the `threshold` argument (exposed in this package as the `threshold_structure()` dial) constrains the intercepts rather than the slopes.

### `ordinalNet::ordinalNet` — two omnibus logical flags

Matrix interface, so build the design matrix first:

```r
house_x <- model.matrix(Sat ~ Infl + Type + Cont, data = house_data)[, -1]
house_y <- house_data$Sat

# parallel (the default)                                        # 8 coefs
ordinalNet::ordinalNet(house_x, house_y, family = "cumulative", link = "logit",
                       parallelTerms = TRUE, nonparallelTerms = FALSE)

# fully non-parallel                                            # 14 coefs
ordinalNet::ordinalNet(house_x, house_y, family = "cumulative", link = "logit",
                       parallelTerms = FALSE, nonparallelTerms = TRUE)

# semi-parallel: parallel base + penalized per-threshold deviations   # 20 coefs
ordinalNet::ordinalNet(house_x, house_y, family = "cumulative", link = "logit",
                       parallelTerms = TRUE, nonparallelTerms = TRUE,
                       parallelPenaltyFactor = 0)   # optionally leave the parallel part unpenalized
```

There is no per-covariate declaration — the semi-parallel model is over-parameterized on purpose (8 parallel + 12 deviation coefficients here, with names like `InflMedium` plus `InflMedium:1`, `InflMedium:2`), and the elastic-net penalty decides covariate-by-covariate which deviations survive.

So the declaration itself is omnibus, but the semi-parallel + penalty formulation acts as a soft, data-driven version of covariate-specific non-parallelism.

Both `nonparallelTerms = TRUE` variants warn that out-of-sample cumulative probabilities may not be monotone with the `cumulative` family; this is inherent to unconstrained non-parallel cumulative models, not a bug.

</details>

corybrunson:
tl;dr: I also feel that the argument is convoluted at this point. I'd be game to implement Option 2 in this upgrade, then assess user appetite for more flexibility for a future upgrade. How does that sound?

Option 1: I thought of doing something like this, borrowing `cs()` from {brms} rather than creating a new "special". A complication is that {VGAM} can admit both smoothers `s()` and category-specificity `cs()` for the same predictor while {brms} cannot. By introducing that syntax i'd feel obligated to allow users the full functionality of both packages through it, but that would require some careful, possibly fragile translation of the input formula.

Option 2: Fine with me, and consistent with the dial.

Option 3: I feel that this syntax is too far removed from the engine packages. They're all over the place themselves, but i'm concerned about falling into the standard proliferation trap. (It also removes the possibility of including both parallel and non-parallel terms for the same predictor, which some engines allow and is currently achieved via the two-element list input option.)

It would be possible to introduce a wrapper-enabled engine parameter, as in Option 1 or 3, while also introducing the `parallel_reg` model parameter, which could be handled as in the current PR or as a simple logical (Option 2). That too could wait until a future upgrade.

I have an elaborate OpenCode-driven RMarkdown doc that informed the design choices; i'll try to streamline it and then commit it in the 'ignore' folder.

hfrick:
Since I have the related dials PR waiting for me, I'm gonna add some thoughts on tuning.

I think we'll only get tuning via dials though
- all or nothing is parallel, i.e., a boolean `dials::parallel_reg()`
- data-driven relaxation of the parallel assumption via shrinkage in `ordinalNet`, i.e., we link their `parallelPenaltyFactor` argument to `dials::penalty()`

Any other combination of parallel and non-parallel assumptions depends on the predictors available and must thus be hand-crafted. Tuning that is only possible by making bespoke workflows and evaluating them as a workflowset.

If we want a straightforward relationship between tuning parameter and argument to the model spec via `ordinal_reg()`, it's Max's option 2 with the additional option of enabling the tuning of/via `ordinalNet`s `parallelPenaltyFactor`. In parsnip, we'd protect the engine's argument(s) for parallel regression and only translate the all and nothing cases.

If we want to support a more detailed specification of which predictors carry a parallel assumption and which don't,  we a) need a unifying syntax b) need to translate that and c) can only tune via workflow sets. I'll leave how much to translate to you two but out of the two syntax options, I'd prefer Option 3 because that allows you to specify most concisely. I'd add in `nonparallel ~ . - a` for "everything non-parallel except for a which is parallel" to round it out. Then you can say "all", "nothing", and both ways of "default this with the following exceptions" without ending up typing the whole predictor list.

The absolute maximum of complexity I would pack into a `parallel_reg` argument to `ordinal_reg` is the combination of the two options above, allowing it to either be a boolean or a formula but not a combination of the two.

hfrick:
Option 2 to start and more elaborate things later sounds really good to me 👍

corybrunson:
OK! I'll switch to Option 2 within a few days and confirm when done—unless instructed otherwise. I don't think it will incur any changes to the dials PR.

corybrunson:
@topepo @hfrick This revision changes `parallel_reg` to a logical parameter. It also:

* moves argument handling for several engines from {ordered} wrappers to `parsnip::translate.ordinal_reg()`;
* exposes `vgam` engine arguments corresponding to `ordinal_reg()` model arguments and assigns the model argument dials to them.

`ordinalNet::ordinalNet()` warns when `parallelPenaltyFactor` is passed without parallel terms; i made this an error in the wrapper function, which seemed more consistent with tidymodels behavior. But that doesn't address @hfrick 's suggestion that it be handled together with `penalty`—in case we still want to implement it.

Remaining is the issue of {VGAM}'s `Thresh` argument being different from `threshold` as it was in version 1.1-9. Should that be addressed? or should {ordered} just require version 1.1-10? (I, for example, still use a computer that can't install that version.)

I still need to clean up the document that illustrates how the parallel regression assumption is handled by the different packages; it's next on my list.

corybrunson:
I've tracked the document [here](https://github.com/corybrunson/ordered/pull/29/changes/742f16e7ad83a107d0d241815ca11eec8cc5d995).

R/ordinal_reg.R
```diff
@@ -140,146 +162,250 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+  if (engine == "glmnetcr") {
+    x <- translate_ordinal_reg_glmnetcr(x, call = call)
+  }
+
+  x
+}
+
+check_ordinal_reg_parallel <- function(x, engine, call = rlang::caller_env()) {
+  # reject `parallel_reg` for engines that don't support assumption violations
+  if (!engine %in% c("clm", "vglm", "ordinalNet", "brms")) {
```
hfrick: Why is `"brms"` in that list? I'm not seeing it in the list of engines in {ordered} and nowhere else in parsnip.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # parsnip (development version)
 
+* `ordinal_reg()` gains `threshold_structure` and `parallel_reg` arguments for controlling threshold constraints and the parallel regression assumption. `threshold_structure` is character-valued and usable by the `clm` and `vglm` engines. `parallel_reg` is logical-valued and usable by the `clm`, `vglm`, and `ordinalNet` engines.
+
+* `gen_additive_mod()`'s `vgam` translator handles as engine parameters the four arguments corresponding to `ordinal_reg()` model parameters.
```
hfrick: This isn't all that easy to read. Let's rephrase that in terms of what that means to the user.

R/ordinal_reg_vglm.R
```diff
@@ -10,3 +10,108 @@
...
+  dials::values_ordinal_link,
+  c("foldsqrt", "logc", "gord", "pord", "nbord")
+)
+
+values_threshold_structure_vglm <- c(
+  dials::values_threshold_structure,
+  "qnorm"
+)
+
+translate_ordinal_vgam_args <- function(args) {
```
hfrick: The NEWS entry says `vglm` uses `parallel_reg` but the translation for that uses this function which does not touch `parallel_reg`.

hfrick: Is that done in the fit wrapper in {ordered}?

R/ordinal_reg_vglm.R
```diff
@@ -10,3 +10,108 @@
...
+  }
+
+  # fmt: skip
+  if (
+    !link %in% c(
+      "logitlink", "probitlink", "logloglink", "clogloglink", "cauchitlink",
+      "foldsqrtlink", "logclink", "gordlink", "pordlink", "nbordlink"
+    )
+  ) {
+    link <- match.arg(link, values_ordinal_link_vglm)
```
hfrick: Let's use `rlang::arg_match0()` instead of `match.arg()` with a vector of values. There are a few instances of this, not just here. The rlang version avoids partial matching.

R/ordinal_reg_vglm.R
```diff
@@ -10,3 +10,108 @@
...
+# by `translate.ordinal_reg()` and `translate.gen_additive_mod()`.
+values_ordinal_link_vglm <- c(
+  dials::values_ordinal_link,
+  c("foldsqrt", "logc", "gord", "pord", "nbord")
+)
+
+values_threshold_structure_vglm <- c(
+  dials::values_threshold_structure,
+  "qnorm"
+)
```
hfrick, line 35: This pulls the values from dials at parsnip's install time, from whatever version of dials was installed then. If dials changes, it doesn't affect this. Let's instead inline them. I assume those values are pretty stable and then it's a) easier to debug if we need to and b) we can drop dials from the `Imports` in the DESCRIPTION.

R/ordinal_reg.R
```diff
@@ -140,146 +162,250 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+  call = rlang::caller_env()
+) {
   x <- translate.default(x, engine, ...)
 
-  # REVIEW: What's the preferred way to flag when a legitimate model parameter
-  # is passed a value that the engine doesn't accept?
-  if (engine == "polr") {
-    oddslink <- rlang::eval_tidy(x$args$odds_link)
+  check_ordinal_reg_parallel(x, engine, call = call)
+  warn_ordinal_reg_odds_link(x, engine, call = call)
```
hfrick: This gives two different answers to "what's the preferred way to flag a value the engine doesn't accept?" with an error for `parallel_reg` and a warning for the `odds_link`. Can we make that consistent? I'd error rather than warn because those are main arguments and the feedback should be clear when a user tries to set them to something impossible.

Also, since these only need the `engine` for the check to be carried out, they should move into `check_args()`.

R/ordinal_reg_ordinalNet.R
```diff
@@ -12,3 +12,28 @@
...
+match_ordinal_link_ordinalNet <- function(link) {
+  if (!is.character(link)) {
+    return(link)
+  }
+  link <- match.arg(link, dials::values_ordinal_link)
+  if (link == "logistic") {
+    link <- "logit"
+  }
+  if (link == "loglog") {
+    cli::cli_abort(
```
hfrick, line 37: needs a `call` argument

R/ordinal_reg_vglm.R
hfrick: And this should pass the `call` on -- edit: done

R/ordinal_reg_vglm.R
```diff
@@ -10,3 +10,108 @@
...
+}
+
+check_ordinal_link_family_vglm <- function(family, link) {
+  if (
+    is.character(family) &&
+      is.character(link) &&
+      family == "acat" &&
+      link %in% c("logitlink", "probitlink", "clogloglink")
+  ) {
+    cli::cli_abort(
```
hfrick, line 135: Should get a `call`

R/ordinal_reg.R
```diff
@@ -140,146 +162,250 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+# adapted from `.check_glmnet_penalty_fit()`
+check_ordinal_reg_penalty <- function(
+  penalty,
+  engine,
+  call = rlang::caller_env()
+) {
+  pen <- rlang::eval_tidy(penalty)
+  if (length(pen) != 1L) {
+    msg <- c(
+      "x" = "The {engine} engine ignores {.arg penalty} in favor of a
```
hfrick:
```suggestion
      "x" = "The {.val {engine}} engine ignores {.arg penalty} in favor of a
```

R/ordinal_reg.R
```diff
@@ -140,146 +162,250 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+        path that enables prediction at interpolated penalty values.",
+      "!" = "{.arg penalty} was passed {length(pen)} value{?s}.",
+      "i" = "Use `path_values` to override the default path."
```
hfrick:
```suggestion
      "i" = "Use {.arg path_values} to override the default path."
```

tests/testthat/test-ordinal_reg.R
```diff
@@ -10,3 +12,116 @@ test_that("odds_link", {
   tidy_spec <- ordinal_reg(engine = "polr", odds_link = "adjacent_categories")
   expect_warning(translate(tidy_spec), "odds_link")
 })
+
+test_that("parallel_reg is validated", {
+  expect_no_error(check_args(ordinal_reg(parallel_reg = NULL)))
+  expect_no_error(check_args(ordinal_reg(parallel_reg = TRUE)))
+  expect_no_error(check_args(ordinal_reg(parallel_reg = FALSE)))
+
+  expect_error(
```
hfrick: These `expect_error()` should be snapshot tests via `expect_snapshot(error = TRUE, ...)`.

tests/testthat/test-ordinal_reg.R
```diff
@@ -10,3 +12,116 @@ test_that("odds_link", {
...
+  )
+
+  result <- translate_ordinal_reg_vglm(x)
+
+  expect_equal(result$method$fit$args$link, "logitlink")
+  expect_equal(result$method$fit$args$family, "cratio")
+  expect_equal(result$method$fit$args$Thresh, "equid")
+  expect_equal(result$method$fit$args$untouched, 1)
+
+  expect_error(
```
hfrick: like above, these should be snapshot tests

tests/testthat/test-ordinal_reg.R
```diff
@@ -10,3 +12,116 @@ test_that("odds_link", {
...
+  expect_null(result$method$fit$args$parallel_reg)
+  expect_equal(result$method$fit$args$lambdaMinRatio, 0.2)
+  expect_true(result$method$fit$args$includeLambda0)
+  expect_null(result$method$fit$args$lambdaVals)
+})
+
+test_that("unsupported non-parallel models give engine guidance", {
+  spec <- ordinal_reg(parallel_reg = FALSE)
+
+  expect_error(
```
hfrick: and another snapshot test

R/ordinal_reg.R
```diff
@@ -16,6 +16,13 @@
 #'  model is `"polr"`.
 #' @param ordinal_link The ordinal link function.
 #' @param odds_link The odds or probability link function.
+#' @param threshold_structure The threshold structure for the cutpoints
+#'  (specific engines only).
+#' @param parallel_reg Logical; whether predictor effects are shared across
+#'   thresholds (`TRUE`) or category-specific effects (`FALSE`). The default,
+#'   `NULL` uses the engine default.
+#'
+#'   Available for specific engines only.
```
hfrick:
```suggestion
#'   `NULL`, uses the engine default. Available for specific engines only.
```

hfrick (commented):
I'll fix up the mechanical comments on this myself, leaving the questions like "what's this doing here?" to you if I don't come across the answer while doing me 🧹  ✨ sweep

R/ordinal_reg.R
hfrick: Ah, came across the comment on brms in the PR description. I would not make this bet here and remove it. We can always add it again but I'd like to prevent us from being surprised by our past selves for adding something but not everything for an engine.

corybrunson: That makes sense! To help me with the process, should i address the changes you comment on and push commits, or are you in the process of making changes yourself?

hfrick: I'm currently still working on some stuff so if you could hold off for the moment, that would be great, thanks! I'll comment on the PR when I'm handing it back to you and Max.

R/ordinal_reg.R
```diff
@@ -140,146 +160,250 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+
+  # translate `parallel_reg` to the `nominal` formula accepted by `clm()`
+  # NB: The formula must be explicitly constructed since `clm()` would expand
+  # `<resp> ~ .` against all variables (including the response).
+  # The formula is constructed at fit time, when the model formula is
+  # available, rather than at translation time.
+  parallel_arg <- eval_ordinal_arg(x$method$fit$args$parallel_reg)
+  if (isFALSE(parallel_arg)) {
+    x$method$fit$args$nominal <-
+      rlang::expr(ordinal_reg_nominal_formula(formula))
```
hfrick: This is now throwing a NOTE about no visible binding for global variable `formula`. Just adding it to global variables is likely just papering over it so what's a better way to solve it? Two options I got to before my brain melted: resolve it in a wrapper around the fit function or defer the call to an exported helper (like `min_cols()` for `rand_forest()`). I would probably tend towards the first option but I saw that Cory just took the opposite direction. Maybe we need to lay that out properly before we decide?

R/ordinal_reg.R
```diff
@@ -140,146 +160,250 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+  )
+  rlang::new_formula(NULL, rhs, env = rlang::f_env(formula))
+}
+
+eval_ordinal_arg <- function(x) {
+  if (rlang::is_quosure(x)) {
+    x <- rlang::eval_tidy(x)
+  }
+  x
+}
```
hfrick: This look like a coding agent taking a shortcut. This has nothing to do with ordinal regression per se. I would like to get rid of this. If it's necessary, there is probably already a version of this under a different (hopefully more sensible) name in parsnip.

NEWS.md
hfrick: And let's add the number of this PR to the two news bullets

R/ordinal_reg.R
```diff
@@ -140,146 +160,262 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
+  } else if (isTRUE(parallel_arg)) {
+    x$method$fit$args$nominal <- NULL
+  }
+  x$method$fit$args$parallel_reg <- NULL
+
+  x
+}
+
+ordinal_reg_nominal_formula <- function(formula) {
+  terms <- rlang::syms(all.vars(formula[[3L]]))
```
hfrick: Do we want the terms here? Because if so, `all.vars()` isn't getting us that.

hfrick:
@topepo @corybrunson handing back to you with a couple of questions/comments 🙌

R/ordinal_reg.R
corybrunson: I agree with the general principle and made the change in bf801a5aa17e114d39c707131746eb0d65ad2220. Note that the check now also applies to the {rms} engines.

I'll move them to `check_args()` in a forthcoming commit.

corybrunson: Done in cdf7539ad444a73865c254ece5518ec828d347e8.

R/ordinal_reg.R
corybrunson: Done in 614b7f8fd256894e1614519caeb0a6baa4d8c90c.

NEWS.md
corybrunson: Done in 2349be2b8cbe199ed6486df5ad9d0bf0b205f517.

R/ordinal_reg_vglm.R
corybrunson: @hfrick i'm sorry—what does it mean to inline them?

R/ordinal_reg.R
corybrunson: You're right; in d275d5f3d0d113fc31bb75cde4d489fe166cb6d1 i think i replaced all instances with `eval_tidy()`, which seems to work fine.

R/ordinal_reg.R
corybrunson: I felt there should be a way to "conceal" the name `formula` but still pass it directly to `nominal`. An attempt is in 6dea8dea75b82e3bc8500867bb9615b77ee8bdb5.

R/ordinal_reg.R
```diff
@@ -134,152 +154,271 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
...
-    if (rlang::is_quosure(link_arg)) {
-      link_val <- rlang::eval_tidy(link_arg)
-    } else {
-      link_val <- link_arg
+  invisible(NULL)
+}
+
+check_ordinal_reg_odds_link <- function(x, engine, call = rlang::caller_env()) {
+  if (engine == "polr" || engine == "clm" ||
+      engine == "lrm" || engine == "orm") {
```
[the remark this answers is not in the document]
topepo: or `engine %in% c("polr", "clm", "lrm", "orm")`?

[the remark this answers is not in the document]
corybrunson: Sure—i just got stuck on the pattern.

hfrick:
I've had another in-depth look at this. And while I still need to look at tuning `parallelPenaltyFactor`, here are the conclusions to two of my open questions. Shout if you disagree, especially @corybrunson  on the second one.

**What does "all or nothing" for the parallel regression assumption mean when the terms are not just the main effects?**

You can inline transformations in the formula, such as `y ~ I(x > 10)`, and hand that to `ordinal::clm()`, either in the main formula or the `nominal` formula. The previous version of  `ordinal_reg_nominal_formula()` turned that into `~ x`, `formula[[-2]]` preserves it. I also think it should mean "all terms" rather than "all untransformed covariates", especially since `ordinal::clm()` supports that. 

**Where should that translation live? In parsnip's `translate()` or a fit wrapper in ordinal?**

We typically use the pattern or writing a fit wrapper. So that would be the consistent pattern. 

A second reason for doing it in a fit wrapper is that the dot expansion of `y ~ .` needs the data and I think it's easier to read/understand if that happens when the data is actually available in the fit wrapper.

Third reason: `ordinal` treats predictors as aliased if they appear in both `formula` and `nominal`. It does not error on this over-specification but it reports the location coefficients as `NA`. We can fix that if we remove those terms from the RHS of the `formula`. That is something we can only do in a fit wrapper. If we were to do that in `translate()`, parsnip runs `make_form_call()` afterwards and overwrites the formula again.

My plan is to adapt this PR to these decisions and include a fit wrapper that can then move to ordered.

corybrunson:
@hfrick on the first question, i agree with what i think you're saying, that whatever formula is provided should have its RHS passed to `nominal`. Are you pointing out an instance where this is not the case?

I'm perfectly happy to put this in a wrapper. I've been trying to minimize the use of wrappers based on my sense from parsnip and its extensions that they are uncommon, that most translation happens (and is supposed to happen) in `translate()`. Your second reason is sound to me; i will take some time to better understand your third.

hfrick:
> Are you pointing out an instance where this is not the case?

Nono, I just wasn't sure yet and didn't quite know if you made the change because you knew the answer or because my review comment sounded like this was a change I was requesting. So I looked into it some more.

Your change to taking the RHS of the formula with `formula[[-2]]` is the right move at heart but I've changed it to bring back the ability to use the dot like in `y ~ .`. If we hand `~ .` to `ordinal::clm()` via the `nominal` argument, it expands it to _all_ columns in the data, including the response. Side note: the aliased coefficients set to `NA` is also visible in the example, in case that's helpful to see.

``` r
library(ordinal)

# formula[[-2]] would pass ~ . through to `nominal`
# but clm() then expands it so that it includes the response (here: rating)

clm(rating ~ ., data = wine, nominal = ~ .)
#> Warning: (-1) Model failed to converge with max|grad| = 24.2177 (tol = 1e-06) 
#> In addition: maximum number of consecutive Newton modifications reached
#> formula: rating ~ response + temp + contact + bottle + judge
#> nominal: ~response + rating + temp + contact + bottle + judge
#> data:    wine
#> 
#>  link  threshold nobs logLik AIC    niter max.grad cond.H  
#>  logit flexible  72   -6.65  181.29 4(0)  2.42e+01 -4.7e+15
#> 
#> Coefficients: (18 not defined because of singularities)
#>   response   tempwarm contactyes    bottle2    bottle3    bottle4    bottle5 
#>         NA         NA         NA         NA         NA         NA         NA 
#>    bottle6    bottle7    bottle8     judge2     judge3     judge4     judge5 
#>         NA         NA         NA         NA         NA         NA         NA 
#>     judge6     judge7     judge8     judge9 
#>         NA         NA         NA         NA 
#> 
#> Threshold coefficients: (8 not defined because of singularities)
#>             1|2        2|3        3|4        4|5       
#> (Intercept)  0.0231920  0.6283106  1.1059666  1.8584870
#> response    -0.0527304 -0.0296348 -0.0250473 -0.0307297
#> rating.L    -1.4054149 -1.0213060 -0.7003377 -0.5868302
#> rating.Q     2.0565729  0.3105848 -0.9663272 -1.9926813
#> rating.C    -1.9876511  2.0426120  1.4006753 -2.5070927
#> rating^4     1.1397919 -3.1188150  3.1493832 -1.6169452
#> tempwarm    -0.2394879  0.0077598 -0.0218046  0.1890912
#> contactyes  -0.0689732  0.0135410  0.0137423  0.3461624
#> bottle2     -0.0407691  0.2020207  0.2848289  0.0000000
#> bottle3      0.1704666  0.0198756  0.1776454  0.1043496
#> bottle4     -0.1606996  0.1398667  0.0563622  0.1787518
#> bottle5     -0.1639352  0.0766322  0.2068223  0.1479191
#> bottle6             NA         NA         NA         NA
#> bottle7     -0.0787402  0.0265101  0.0008362 -0.3449185
#> bottle8             NA         NA         NA         NA
#> judge2       0.5642132  0.1140684  0.1571642  0.0005755
#> judge3      -0.0049634  0.0625681 -0.0588725 -0.0888632
#> judge4       0.0596112  0.1573477  0.3352636 -0.2829015
#> judge5      -0.0375177 -0.0517408  0.2070207  0.1043496
#> judge6       0.0853001  0.1616827  0.0310744  0.1020033
#> judge7       0.5014423  0.2310852  0.2273612  0.0000000
#> judge8      -0.0352228  0.0755831  0.1620765  0.1234708
#> judge9       0.3256553  0.1735300  0.0101452  0.2920383
```

<sup>Created on 2026-08-25 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

hfrick:
A status update:

- I've added a helper for the whole formula business and a fit wrapper in `clm_train.R` which can be moved to `ordered`. I didn't add `ordinal` to parsnip's dependencies which is why R CMD check is currently failing.
- I removed the hard dependency on dials
- I've updated the tests to reflect that some checks now live in `check_args()` rather than `translate()`

I still have on my list to look at the changes related `gen_additive_mod()` and the topic of tuning the penalty parameter for ordinalNet. Reviews of this batch of changes is very welcome, from both of you!

R/ordinal_reg.R
```diff
@@ -157,6 +157,9 @@ check_args.ordinal_reg <- function(object, call = rlang::caller_env()) {
   # engine compatibility checks
   check_ordinal_reg_odds_link(object, object$engine, call = call)
   check_ordinal_reg_parallel(object, object$engine, call = call)
+  if (object$engine == "clm") {
+    check_ordinal_reg_nominal(object, call = call)
```
hfrick, line 161: rename to `check_clm_nominal()`?

tests/testthat/test-clm_train.R
```diff
@@ -0,0 +1,140 @@
...
+    rlang::f_rhs(res$nominal),
+    rlang::expr(`not valid` + log(x) + poly(x, 2))
+  )
+})
+
+test_that("clm_formulas() rejects a formula with no predictors", {
+  d <- data.frame(rating = ordered(rep(1:3, 4)))
+
+  expect_snapshot(error = TRUE, {
+    clm_formulas(rating ~ 1, d)
```
hfrick, line 65: Can we generally still fit an intercept-only model?

R/clm_train.R
```diff
@@ -0,0 +1,72 @@
...
+
+  # `do.call()` inlines the data into the recorded call. Rebuild it so that
+  # printing the fit shows the formulas and engine arguments that were used.
+  res$call <- rlang::call2(
+    "clm",
+    !!!formulas,
+    data = rlang::sym("data"),
+    !!!dot_exprs,
+    .ns = "ordinal"
+  )
```
hfrick, line 69: Comment from Max: Consider making the call with `rlang::call2()` and evaluating it with `rlang::eval_tidy()` instead so that we don't have to tweak the call afterwards. 

Also, that's the common pattern in parsnip already.

tests/testthat/test-clm_train.R
hfrick: That's something to check once the fit wrapper is in ordered, so that we can actually run it through the parsnip machinery.

NEWS.md
```diff
@@ -1,5 +1,9 @@
 # parsnip (development version)
 
+* `ordinal_reg()` gains arguments `threshold_structure` and `parallel_reg` to control threshold constraints and the parallel regression assumption. The `ordinalNet` engine can use `parallel_reg` while the `clm` and `vglm` engines can use both new arguments (#1393).
+
+* The `vgam` engine for `gen_additive_mod` has arguments corresponding to all four `ordinal_reg` model arguments; these are exposed as engine arguments but tuned by the same dials. (#1393)
```
hfrick, line 5: I think this second bullet is actually a NEWS item for ordered :) I would spell out which engine arguments are now tunable so that people don't have to cross-reference this with the documentation for `ordinal_reg()`. (And in the tidyverse style, function names get a `()` to make them easier to recognise as such.)

hfrick:
Okay, I've managed to pull my head out of the details. The big question here now is where should things live. And I think as much engine-specific code as possible should live in the extension that holds the engine.  I believe that's what we agree on anyway, so I'll work through what I think this means for this PR:

parsnip owns the modelling function `ordinal_reg()` and which engines accept which argument. The extension package owns what those arguments mean/which values they can hold. 

@corybrunson  Does this work for you?

**Translation of argument _values_**

In addition to translating the argument _names_, we also need to translate the argument _values_. That's a new situation for us because for main arguments (to parsnip's modelling function directly), the values had not been this different between engines. (And for pure engine parameters, we've made dials parameter objects specific to the engine and its values.) 

I think this translation should live in the extension package, i.e., it should go into the fit wrapper. All the engines that need such a value translation either already have a wrapper in ordered (`vglm`/`vgam`, `ordinalNet`) or are getting one now (`clm`). If anything changes in the engines, only ordered needs to be updated, not ordered and parsnip. 

From browsing the code with an LLM: This also matches what happens for a lot of other models for which the engines are in an extension package. There are no model-specific `translate()` methods for most of those models and the extension is doing the argument translation, e.g., in bonsai, discrim, baguette.

**Checks on which arguments are specified with the engine of choice**

Checking if an argument is usable by an engine is doable without writing a whole lot of engine-specific values into parsnip and it seems right to me that parsnip does the checking of which arguments it passes to which engine. Doing that kind of checking in ordered would require new wrappers in ordered for the `polr`, `lrm`, and `orm` engines (unlike the translation of values) so that's a bonus to keeping these checks in parsnip.

**What needs to stay in `translate.ordinal_reg()`**

For the `ordinalNet` and `glmnetcr` engines, we need to write something back into the spec to persist for prediction time:

```
# Since the `fit` information is gone for the penalty, we need to have an
# evaluated value for the parameter.
x$args$penalty <- rlang::eval_tidy(x$args$penalty)
```

We can't do that in a fit wrapper so we need to keep that here but we can slim the code for `translate()` down to those kind of essentials.

**`gen_additive_model()`**

The value translation for this should also move into ordered. We can then leave `translate.gen_adddive_mod()` as it is on `main`.
The LLM also noticed that there is a slight bug in this: the helpers for `translate_ordinal_vgam_args()`, like `match_ordinal_link_vglm()`, use the parsnip argument names like `ordinal_link`, which are not the same argument names as used for `gen_additive_mod()` (`link`). 

**Value sets for VAGM' (ordinal) link and threshold structure**

 `values_ordinal_link_vglm` and `values_threshold_structure_vglm` should live where the translation + validation lives, so in ordered. The changes I made to this are useful because they avoid the reference to dials' value going stale but really, if the creation of this vector gets moved into the function (as opposed to the package namespace) that uses it, e.g. `match_threshold_structure_vglm()`, that staleness problem also goes away.

topepo:
I agree with these comments.

corybrunson:
Sorry, i'm facing a deadline and will respond no later than Tuesday!

R/ordinal_reg.R
corybrunson: Yes, done in e03ab29e1963244a62dce5d3cb4b6eb1ffd0acd7.

tests/testthat/test-clm_train.R
corybrunson: Why is an error the desired behavior? My impulse is to allow but ignore `parallel_reg` when there are no predictors.

R/clm_train.R
corybrunson: Sorry again: Is Max's comment in response to the code above, or is the code above in response to Max's comment?

NEWS.md
corybrunson: Both bullets have counterparts in ordered. It seemed worth mentioning the `vgam` arguments here as well, since they are translated differently by `gen_additive_mod()` and `ordinal_reg()`. Should i make that explicit, or just drop the item entirely? I'll spell out the arguments ~in both new files~ in the parsnip file.

corybrunson: Done in 3afef5f6b071854b60eff70d9d86eb06099fd83f.

corybrunson:
@hfrick everything you've proposed sounds good to me. I'll work on it incrementally and should be done this week. I'll tag you both if something unexpected comes up that i don't feel prepared to resolve. Is that OK?

topepo:
That sounds good to me (Hannah's out of the office right now).

corybrunson:
Regarding translation of argument names and values in the wrappers, my plan is (1) for the wrappers to use the same argument names as the engine functions and convert the values internally, and (2) to rely on registration (`set_model_arg()`) to convert the names. Given that i want to adopt a common pattern for all wrappers, this makes sense to me as a way to separate steps (so a developer knows where to look) and to minimize the hassle of changing between using and not using a wrapper (though that'll be a rare change).

This is just to let know and invite advice.

corybrunson:
I believe the plan is to move the penalty path assembly to ordered as well; i just need some time to understand what memory aides will still be needed in `translate.ordinal_reg()`.

hfrick:
I think the penalty path parts need to stay in parsnip because we modify the spec and we can't really do that from a wrapper.

corybrunson:
@hfrick does the current division of labor make sense then? Is the shift complete?

Something i'd like to do is clarify and harmonize how the penalty path–related arguments are handled. Are there any changes in the works to these steps for the `glmnet` engine? Assuming not, i'll create a sub-branch to revise how the `ordinalNet` and `glmnetcr` engine arguments are handled, to better mimic `glmnet`. I can do that this weekend if time permits.

corybrunson:
@hfrick @topepo i've drafted a revision at [this parsnip PR](https://github.com/corybrunson/parsnip/pull/2) and [this ordered PR](https://github.com/corybrunson/ordered/pull/31).

hfrick:
I've not looked at ordered or the fork of parsnip but I think this PR here is fine to be merged as it is now.

corybrunson:
@topepo @hfrick thank you!

Would it be OK for me to shift the outstanding PR from targeting this branch to targeting `main`?

hfrick:
@corybrunson Which PR do you mean? Could you make an issue first so that we can more easily understand what you want to do before we dive into details?

corybrunson:
@hfrick sure, i'll open an issue and point to the new branch. I originally thought it would make sense here because it simplifies the translation of arguments and deviations from default behavior for constructing penalty paths.

PR #1404, opened 2026-08-13
quantile regression null model

### LINKED ISSUE #1348: Quantile regression mode for null model
It can compute the quantiles of the outcome data.

topepo:
Closes #1348

### DIFF
--- a/R/nullmodel.R
+++ b/R/nullmodel.R
@@ -1,36 +1,43 @@
 #' Fit a simple, non-informative model
 #'
-#' Fit a single mean or largest class model. `nullmodel()` is the underlying
-#' computational function for the `null_model()` specification.
+#' Fit a single mean, set of quantiles, or largest class model. `nullmodel()` is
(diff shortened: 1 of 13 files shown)


R/nullmodel.R
```diff
@@ -47,8 +54,16 @@ nullmodel <- function(x, ...) UseMethod("nullmodel")
 
 #' @export
 #' @rdname nullmodel
-nullmodel.default <- function(x = NULL, y, ...) {
-  if (is.factor(y)) {
+nullmodel.default <- function(x = NULL, y, quantile_levels = NULL, ...) {
+  if (!is.null(quantile_levels)) {
+    lvls <- NULL
+    pct <- NULL
```
edgararuiz: If you move lines 59 and 60 to before the `if()` statement, then you wouldn't need to have duplicates in lines 72 and 73, since the only time those vars are not NULL is when `y` is a factor

edgararuiz (approved):
Look great. Just one code change suggestion

Issue #1411, opened 2026-09-08
harmonize penalty path control for elastic net `ordinal_reg` engines


corybrunson:
As originally implemented for the `ordinalNet` engine, and as adapted to the `glmnetcr` engine, when the user does not explicitly provide `path_values`, `translate()` silently modifies certain engine arguments in order to produce a penalty path with greater range and resolution than would be created by default:

``` r
library(parsnip)

# `glmnet`: `family` specified, others placeheld
glmnet_spec <- linear_reg(penalty = .01) |> 
  set_engine("glmnet")
( glmnet_tran <- translate(glmnet_spec) )
#> Linear Regression Model Specification (regression)
#> 
#> Main Arguments:
#>   penalty = 0.01
#> 
#> Computational engine: glmnet 
#> 
#> Model fit template:
#> glmnet::glmnet(x = missing_arg(), y = missing_arg(), weights = missing_arg(), 
#>     family = "gaussian")
print(glmnet_tran$method$fit$args)
#> $x
#> missing_arg()
#> 
#> $y
#> missing_arg()
#> 
#> $weights
#> missing_arg()
#> 
#> $family
#> [1] "gaussian"

# `ordinalNet`: translation silently modifies arguments for penalty path
ordinalNet_spec <- ordinal_reg(penalty = .01) |> 
  set_engine("ordinalNet")
( ordinalNet_tran <- translate(ordinalNet_spec) )
#> ! parsnip could not locate an implementation for `ordinal_reg` model
#>   specifications using the `ordinalNet` engine.
#> ℹ The parsnip extension package ordered implements support for this
#>   specification.
#> ℹ Please install (if needed) and load to continue.
#> Ordinal Regression Model Specification (classification)
#> 
#> Main Arguments:
#>   penalty = 0.01
#> 
#> Computational engine: ordinalNet
print(ordinalNet_tran$method$fit$args)
#> $nLambda
#> [1] 120
#> 
#> $lambdaMinRatio
#> [1] 1e-08
#> 
#> $includeLambda0
#> [1] TRUE

# `glmnetcr`: similar to `ordinalNet`
glmnetcr_spec <- ordinal_reg(penalty = .01) |> 
  set_engine("glmnetcr")
( glmnetcr_tran <- translate(glmnetcr_spec) )
#> ! parsnip could not locate an implementation for `ordinal_reg` model
#>   specifications using the `glmnetcr` engine.
#> ℹ The parsnip extension package ordered implements support for this
[...]

PR #1412, opened 2026-09-10
generalize glmnet helpers to check & set penalty path arguments

corybrunson:
As prompted [here](https://github.com/tidymodels/parsnip/issues/1411#issuecomment-5618978993), this is a minimal revision to the glmnet helper functions that generalizes them to work for similar engines that may use different penalty parameters:
* `.check_glmnet_penalty_fit()` gets an `engine` argument with no default that is set to `"glmnet"` in all invocations. It is used only to specialize the `cli_abort()` message.
* `set_glmnet_penalty_path()` gets a `penalty_arg` argument with no default that is set to `"lambda"` in all invocations. It is used to access the penalty value stored at `x$method$fit$args[[penalty_arg]]`.

Additionally, a new line is added to the (already long) `.check_glmnet_penalty_fit()` abort message to explicitly reference the `path_values` argument. I think this is essential for users to be pointed to.

~Note that i have not updated the snapshot tests, in part since i don't have all engines installed.~ Snapshot tests are now updated.

The generalization is in anticipation of also invoking these functions for glmnetcr and ordinalNet penalty paths; if other engines have similar controls then they might invoke them as well.

The helper functions might be renamed, e.g. by replacing `_glmnet_` with `_elastic_net_` in each name; but, since the glmnet engine (i think) originated the design pattern, it might make sense to keep `glmnet` in the name.

### DIFF
--- a/R/glmnet-engines.R
+++ b/R/glmnet-engines.R
@@ -386,18 +386,23 @@ format_glmnet_multinom_class <- function(pred, penalty, lvl, n_obs) {
 #' @rdname glmnet_helpers
 #' @keywords internal
 #' @export
-.check_glmnet_penalty_fit <- function(x, call = rlang::caller_env()) {
+.check_glmnet_penalty_fit <- function(
+    x,
+    engine,
+    call = rlang::caller_env()
+) {
   pen <- rlang::eval_tidy(x$args$penalty)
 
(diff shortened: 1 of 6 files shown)


topepo:
Thanks!

PR #1416, opened 2026-09-17
Update glm_grouped() error for two-level outcomes

KJagz:
Updated the error glm_grouped() has when responses are not two level factor. Instead of a single line stating: "The response column {.val response} should be a two-level factor."

Now the error names the column, reports how many levels it has, and states what the levels are:

  ! The response column `Dept` should be a two-level factor.
  i It has 6 levels: "A", "B", "C", "D", "E", and "F".


topepo:
Thanks!

PR #1418, opened 2026-09-17
Fix predict() error for deprecated quantile arg

### LINKED ISSUE #1258: Deprecation of `quantile` arg to `predict_quantile()` not reached via `predict(type = "quantile")`
I would expect this to warn me to use the new argument name `quantile_levels` instead of `quantile` but `check_pred_type_dots()` doesn't let `quantile` through and the deprecation warning sits in `predict_quantile.model_fit()`

``` r
library(censored)

f_fit <- survival_reg() %>%
  fit(Surv(time, status) ~ age + ph.ecog, data = lung)

predict(f_fit,
  new_data = lung,
  type = "quantile",
  quantile = c(0.2, 0.8)
)
#> Error in `predict()`:
#> ! The ellipses are not used to pass args to the model function's predict
#>   function. These arguments cannot be used: "bad_args"
```

<sup>Created on 2025-02-13 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

bjornkallerud:
Fixes #1258

Added `quantile` to `other_args` so deprecation warning will surface. Also, converted `.val` to `.arg` because "bad_names" was being printed instead of its value.


topepo (approved):
Thanks!

PR #1424, opened 2026-09-21
use generalized glmnet helpers for ordinalNet and glmnetcr translation

corybrunson:
Previous PR #1412 generalized two glmnet `translate()` helper functions, as discussed in #1393. This PR uses these generalized helpers to de-duplicate and simplify translation code for two `ordinal_reg()` engines that follow the glmnet pattern: ordinalNet and glmnetcr.

The replaced code also silently modified several engine arguments that guide the assembly of the penalty path. Instead, the simplified code defers to engine defaults, which may result in narrower paths and result in prediction errors in code written since #1346 (version 1.5.0). The user must pass values to these engine arguments or specify a penalty path via `path_values` (as now mentioned in the error message) to control the path. (Predictions still interpolate as before.)


tests/testthat/test-translate.R
```diff
@@ -246,38 +246,19 @@ test_that("arguments (nearest_neighbor)", {
 test_that("arguments (ordinal_reg)", {
   suppressMessages({
     basic <- ordinal_reg(mode = "classification")
-    logit_link <- ordinal_reg(ordinal_link = "logistic") |> set_engine("clm")
-    ordinal_link <-
-      ordinal_reg(ordinal_link = "probit") |> set_engine("ordinalNet")
+    ordinal_link <- ordinal_reg(ordinal_link = "probit")
```
topepo, line 249: Just an FYI... sometimes when placing these in the middle of the file will require re-running all of the snapshots since they are indexed by order. That's the test failure that you saw

corybrunson: Thanks—i've been using `devtools::test()` and following the prompts to review snapshots; do you mean that i should do this twice, or should i take other steps as well? I thought i addressed them all, though i didn't update one because it was due to a package not being installed rather than to these changes.

topepo (approved):
Looks good. Just in case, I ran reverse dependency checks

We checked 103 reverse dependencies (102 from CRAN + 1 from Bioconductor), comparing R CMD check results across CRAN and dev versions of this package.

 * We saw 0 new problems

tests/testthat/test-translate.R
topepo: If there are test errors I often delete the snapshot md and just remake them all of that test file. Our IDEs will also have clickable messages to accept or review snapshot changes, which is helpful.

corybrunson: Thanks—i'll try to remember to do that when in doubt.

